PR #22, opened 2016-02-29
move code out of extraneous jacktrip/ folder

umlaeute:
since the double-indirection of the `jacktrip/` folder causes ugly hacks in the debian packaging, i suggest to get rid of it.

Closes https://github.com/jcacerec/jacktrip/issues/18


umlaeute:
fixed with e2d3e72430d773b988725bf0bd91855aa36e5154

PR #36, opened 2018-08-26
Select device ID when using rtaudio

spmp:
## Select device ID when using `rtaudio`
Allow the selection of `device_id` when using `rtaudio` instead of relying on the default.

Should be used in conjunction with an `rtaudio` version that supports outputting the device ID on `audioprobe`


jcacerec:
This looks great, thanks Jasper!

spmp:
No worries

PR #40, opened 2020-01-22
Feature/merge from chris gitlab

Foaly:
This is an attempt to merge the development [Chris GitLab](https://cm-gitlab.stanford.edu/cc/jacktrip/) instance back into the mainline jacktrip.
Right now it has not been verified to work, but I will try to do that this weekend.
Any code reviews are very welcome though!

Edit: ... for some reason Github displays the diff weird. If you want to verify the merge, please take a look at commit c21231855d81f23216b837ca658a0d23d338c65e

### DIFF
--- a/CHANGESLOG.txt
+++ b/CHANGESLOG.txt
@@ -1,5 +1,12 @@
 ---
 master
+
+---
+1.2
+- (fixed) Compilation under ubuntu
+- (removed) Rtaudio mode and dependencies
+- (fixed) IPv4-mapped IPv6 addressing bug
+...
 - (fixed) Fixed compilation for  MacOSX10.11.sdk.
 - (update) Updated to RtAudio 4.1.1, and using shared lib in linux.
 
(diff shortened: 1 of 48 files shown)


colinsullivan:
Indeed difficult to read the diff here because of all the whitespace changes.

One general comment: using the #ifdefs to modify the class constructor arguments, seems like an anti-pattern.  Consider a subclass, overloading the constructor, or breaking the WAIR-related stuff into a separate module (i.e. composition vs. specification).

That said, I have little context into the WAIR goals and a surface-level understanding of your codebase.  Is it to enable ugen plugins created in FAUST?

Foaly:
Yes I know. Unfortunatelly the two repos drifted apart quite a bit and this was the attempt to get them back together.
I'm also not a huge fan of the #ifdefs. There is some room for some general source code refactoring and better software design here. I'm not sure if I'll have time in the next weeks/month. But if you want to do some of it, feel more the welcome to contribute :)

colinsullivan:
Gotcha!

Thanks for the response and the invitation.

PR #44, opened 2020-04-04
fix to set clientname when in server mode

geraldmwangi:
This fix sets the jack clientname when mPeerAddress is not set (in server mode)


geraldmwangi:
Thanks

geraldmwangi:
Btw I saw that although jacktrip is mostly c++ code, strings are handled as c-strings. I think its better to use std::string due to proper bounds checking. Although c-strings are 0 terminated its easy to overflow the buffer if one is only supplied with the char* pointer to the string

PR #70, opened 2020-06-15
Add option to disable hub server autopatch

madwort:
* this is so that we can manage patches ourselves in our own simple scripts, without having to interact with the C++ server
* disadvantage is that when a client disconnects, its input is left dangling, leading to noise
* ideally the hub server would have a mechanism to notify external scripts that the client connection status has changed

### DIFF
--- a/src/JackTrip.h
+++ b/src/JackTrip.h
@@ -111,7 +111,8 @@ public:
         CLIENTECHO,  ///< Client Echo (client self-to-self)
         CLIENTFOFI,  ///< Client Fan Out to Clients and Fan In from Clients (but not self-to-self)
         RESERVEDMATRIX,  ///< Reserved for custom patch matrix (for TUB ensemble)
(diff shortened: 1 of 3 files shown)


psiborg112:
This won't work. It'll still connect the server to the clients in the default way. Instead you need to use --nojackportsconnect. That said, it would be good to have a hub patch mode that acts as a synonym for this, since it's the first option that most people will think to play with to get the desired result. (I know I sure did.) So here's a modification of the files that will do what you're seeking to do here. (I coded them independently before I saw this pull request for a separate project, so there are some minor, inconsequential differences.)
[Modified source.zip](https://github.com/jacktrip/jacktrip/files/4908067/Modified.source.zip)

madwort:
Ah, thanks for the feedback @psiborg112  - I've been running this on a server with `jackd -ddummy` so hadn't realised this. Sounds like we don't need this option, we need better documentation - how about this as a suggestion: https://github.com/jacktrip/jacktrip/pull/82

PR #75, opened 2020-07-01
first pass: Travis integration

spencersalazar:
Basic Travis Continuous Integration support. Adds a `.travis.yml` to the repo and fixes a few things that were preventing it from compiling on Travis' Linux VM.

Currently this simply compiles jacktrip on Linux + Mac to ensure that it compiles on both. A test suite + Windows + installer generation can be added on top of this foundation.  

To close the loop, Travis also needs to be added to the jacktrip repo, which needs to be done by someone with read/write permission to the repo (I can do this if you add me). 

Note: please use "squash and merge" when merging as there are a lot of pointless commits to test Travis functionality; squashing will reduce these down to a single commit.

### DIFF
--- a/src/RingBuffer.h
+++ b/src/RingBuffer.h
@@ -44,6 +44,8 @@
 
 #include "jacktrip_types.h"
 
+#include <atomic>
+
 //using namespace JackTripNamespace;
 
 
--- a/src/build
+++ b/src/build
@@ -17,9 +17,6 @@ if [[ $platform == 'linux' ]]; then
     if hash qmake-qt5 2>/dev/null; then
 	echo "Using qmake-qt5"
 	QCMD=qmake-qt5
-    elif hash qmake-qt4 2>/dev/null; then
-	echo "Using qmake-qt4"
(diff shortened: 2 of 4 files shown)


src/RingBuffer.h
```diff
@@ -44,6 +44,8 @@
 
 #include "jacktrip_types.h"
 
+#include <atomic>
```
spencersalazar, line 47: this header was causing compile errors on Travis / Linux

src/build
```diff
@@ -17,9 +17,6 @@ if [[ $platform == 'linux' ]]; then
     if hash qmake-qt5 2>/dev/null; then
 	echo "Using qmake-qt5"
 	QCMD=qmake-qt5
-    elif hash qmake-qt4 2>/dev/null; then
```
spencersalazar, line 20: this check for `qmake-qt4` was causing a compile failure if `qmake` defaults to Qt5 but `qmake-qt4` is also present. Since Qt4 is apparently no longer supported, this seemed fine to remove.

PR #76, opened 2020-07-06
build in a subdirectory

dyfer:
Currently running the `./build` script generates artifacts in the `src` directory, being intermingled with source files and showing up as uncommitted changes in the git repository.

In this PR I propose to create the `build` subdirectory, add its contents to `.gitignore` and change the build script to output files there. 

With my limited experience I'm not sure if this is the best approach, I'm looking forward to getting some feedback on this!

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -1,3 +1,6 @@
 # QtCreator user files
 *.pro.user
 *.pro.user.*
+
+# build directory
+builddir/
--- a/INSTALL.txt
+++ b/INSTALL.txt
@@ -103,7 +103,7 @@ $ qmake -config nojack jacktrip.pro
 $ make release
 
 
-If you want to install (using Terminal): on the /src directory type:
+If you want to install (using Terminal): on the /builddir directory type:
 
 $ sudo cp jacktrip /usr/local/bin/
   (enter your password when prompted)
--- a/src/build
+++ b/src/build
@@ -28,15 +28,17 @@ elif [[ $platform == 'macosx' ]]; then
 fi
 
 # Build
+mkdir -p ../builddir
+cd ../builddir
 if [[ $1 == 'nojack' ]]; then
     echo "Building without Jack"
-    $QCMD -spec $QSPEC -config nojack jacktrip.pro
+    $QCMD -spec $QSPEC -config nojack ../src/jacktrip.pro
     make clean
-    $QCMD -spec $QSPEC -config nojack jacktrip.pro
+    $QCMD -spec $QSPEC -config nojack ../src/jacktrip.pro
     make release
 else
-    $QCMD -spec $QSPEC jacktrip.pro
+    $QCMD -spec $QSPEC ../src/jacktrip.pro
     make clean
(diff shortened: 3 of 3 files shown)


ntonnaett:
Meson already does this. But we are not ready to switch yet. I'm not sure if I like the directory name because it's the same as the script itself. So maybe that will become confusing in the documentation afterwards. Maybe just rename it to builddir. That's the name we use in the meson installation description.

Let's touch the builddir in the build script. So we don't have an empty directory in the repository. And then put the builddir in the .gitignore of the root directory. What do you think?

We have PR #64 that also changes the documentation. But I think that shouldn't collide.

dyfer:
Thanks for comments @ntonnaett. How does this look now?
I'm not familiar with meson, is there any documentation for building jacktrip with it or should I just read up about meson itself?

ntonnaett:
Beware that the default in meson is a debug build. For compiling optimized binaries you can add [--buildtype=debugoptimized](https://mesonbuild.com/Quick-guide.html#compiling-a-meson-project) when running meson. That's convenient because you can have one build directory for debug builds and one for release builds.

> It seems I'm touching a part that would be removed by #64 so it would indeed collide. Should I remove my change to
> INSTALL.txt from this PR?

No, leave it. I don't know what the state of the other PR is.

I just saw that the debug directory in src/ gets tracked by git. I think that's also a build artefact. So we see again that it's a good practice to separate both. Could you remove that?
Everything else looks fine to me now.  Thanks for your work!

dyfer:
Thanks for the info on meson.

> I just saw that the debug directory in src/ gets tracked by git. I think that's also a build artefact. So we see again that it's a good practice to separate both. Could you remove that?

I'm not sure I understand. Currently the `./build` script will create `release` and `debug` directories in `builddir`. Does meson still create artifacts in `/src`?

ntonnaett:
Sorry, I wasn't clear enough. That artefact is not your fault. Someone falsely added it to the repository. It's also in master. I just thought that removing it would fit to this PR. So everything is fresh and clean afterwards.

dyfer:
> Sorry, I wasn't clear enough. That artefact is not your fault. Someone falsely added it to the repository. It's also in master. I just thought that removing it would fit to this PR. So everything is fresh and clean afterwards.

No problem! However, I'm sorry but I can't see any extra artifacts in `/src` in my branch nor in [master](https://github.com/jacktrip/jacktrip/tree/master/src)... what am I missing?

ntonnaett:
Okay, that's weird. You are missing nothing. Git seems to ignore directories called debug by default?!

dyfer:
> Okay, that's weird. You are missing nothing. Git seems to ignore directories called debug by default?!

Weird, I haven't seen that. Thanks for the review!

ntonnaett:
I just realized that this probably breaks travis. So we have to look into this first.

PS And GitHub could be clearer how their workflow is meant.

ntonnaett:
So it looks like that was a false alarm. As the build script stays at its place and travis doesn't mind where the result of the build process is going, there shouldn't be any problem.

PR #81, opened 2020-07-11
Travis update

spencersalazar:
Slight changes to Travis CI script to accommodate differences between travis-ci.com and travis-ci.org.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -10,8 +10,6 @@ dist: bionic
 
 branches:
   - master
-  - develop
-  - travis2
 
 env:
   global:
--- a/.travis.yml
+++ b/.travis.yml
@@ -28,7 +28,7 @@ addons:
 
 before_script:
   # manually install brew packages as addon: homebrew is broken in the default osx image
-  - if [ "$TRAVIS_OS_NAME" = "osx" ]; then brew install $PKGS_OSX; fi
+  - if [ "$TRAVIS_OS_NAME" = "osx" ]; then HOMEBREW_NO_AUTO_UPDATE=1 brew install $PKGS_OSX; fi
   # mac: workaround for homebrew qmake not being symlinked into $PATH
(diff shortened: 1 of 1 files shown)


ntonnaett:
Thank you for your work on CI. I think that's a great benefit. I'm new to Travis. So maybe you can clarify some things to me.
Why are there two checks — one on github, one on travis-ci.org? I see an empty commit named trigger build. How does a build usually get triggered?
It looks to me that this PR could be just one commit. Do we need to add all branches to .travis.yml that want to use travis? If so, we should add master (or main in the near future) and develop.

spencersalazar:
No problem! 

> Why are there two checks — one on github, one on travis-ci.org?

Im not totally sure, this seems to be an artifact of Github's app integration API. It appears that only one Travis build is actually performed but its registered as two checks through the API. It was a bit finicky to get it all connected and the documentation is not the best so I was content to leave it be since the basic functionality is working adequately. 

> How does a build usually get triggered?

Builds are normally triggered on any commit to a branch that Travis is watching- typically this would only happen once per PR on a `master`/`main`/`develop` branch after the PR is merged but if you committed straight to the branch it would trigger a build for each commit. Its set up right now to also build pull requests when they are created or updated as a pre-merge validation of the PR. 

> It looks to me that this PR could be just one commit. 

Yeah, this could be (should be) squashed into a single commit upon approve/merge using Github's "squash and merge" option. The multiple commits are an artifact of how the `.travis.yml` is developed and tested.  

> Do we need to add all branches to .travis.yml that want to use travis?

Yep, simply add these to `branches:` in the relevant branch. `master`/`main` and `develop` are standard. I didnt see a `develop` branch in the repo so Travis is not configured to build it right now, but could be added easily. 

Feature branches can optionally be added to Travis by adding to `branches:` in `.travis.yml` in that branch, if desired- for instance, to validate compilation on both Mac + Linux in a clean environment, or running an automated test suite if that is available.

PR #96, opened 2020-07-25
Fix all warnings

ntonnaett:
Fixes all compilation warnings.

Could somebody look at the fix for the sign compare warning. This was undefined behavior because nobody knows what the compiler casts. Maybe it wasn't a problem because of implicit boundaries. I'm sure there are some and we should make them explicit.

### DIFF
--- a/src/JMess.cpp
+++ b/src/JMess.cpp
@@ -177,7 +177,7 @@ void JMess::connectSpawnedPorts(int nChans, int hubPatch)
     QString IPS[gMAX_WAIRS];
     int ctr = 0;
 
-    const char **ports, **connections; //vector of ports and connections
+    const char **ports; //vector of ports
     QVector<QString> OutputInput(2); //helper variable
 
     //Get active output ports.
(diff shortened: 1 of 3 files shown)


ntonnaett:
I now think that n shouldn't be size_t.

cchafe:
Instead, should it be a qint64 ?

ntonnaett:
I think that's the best idea if we compare always to qint64. I only find users of receivePacket() in UdpDataProtocol.cpp. And I'm a little bit confused. Maybe because of the threading. 

In line 522 we get the pendingDatagramSize:
int first_packet_size = UdpSocket.pendingDatagramSize();

That we use as n for receivePacket in lin 528:
receivePacket( UdpSocket, reinterpret_cast<char*>(first_packet), first_packet_size);

And in receivePacket() we compare this n to pendingDatagramSize() again. What is the thinking behind this?

cchafe:
Switch it to merge to dev and I'll wrap it up

Issue #87, opened 2020-07-22
Hubserver deadlock with multiple concurrent connections


mikedickey:
Hubserver seems to deadlock when multiple connections are attempted in parallel. I tested both using master and chris branches (as of July 20th, 2020) and both have the same issue. JackTrip will stop accepting any connections when this happens, until it is restarted.

Please see the below script that I wrote called `jacktrip_load.py` to reproduce this. Note that you need to run it on a server that has enough cores (too few cores means everything happens in serial). I can reproduce it 100% of the time when using a --count of 5:

```python3 jacktrip_load.py --server <IP> --count 5```

Note that I added a `--sleep` parameter to sleep in between the spawning of each client. I was unable to reproduce this problem when using a sleep of only 1 second. In fact, I managed to scale JackTrip up to 100 concurrent clients using two 16-vCPU (8 core) servers (one for jacktrip -S and the other for jacktrip_load.py).

```#!/usr/bin/python3

import re
import time
import copy
import argparse
import subprocess
from multiprocessing import Pool

def parse_args():
    parser = argparse.ArgumentParser(description="Spawns JackTrip client processes to test a hubpatch server")
    parser.add_argument("--count", type=int, default=1, help="Number of JackTrip client processes to spawn")
    parser.add_argument("--server", type=str, default=None, help="Hostname or IP address of JackTrip server")
    parser.add_argument("--port", type=int, default=4464, help="Port number of JackTrip server")
    parser.add_argument("--exe", type=str, default="jacktrip", help="JackTrip client executable")
    parser.add_argument("--sleep", type=int, default=0, help="Seconds to sleep in between each client")
    args, extra_args = parser.parse_known_args()
    args.extra_args = extra_args
    if not args.server:
        print("No server specified, using localhost")
        args.server = "localhost"
    return args

def main():
    args = parse_args()
    spawn_clients(args)

def spawn_clients(args):
    jacktrip_args = [args.exe, "-C", args.server, "--peerport", str(args.port)]
    if args.extra_args and args.extra_args != "":
        jacktrip_args.extend(args.extra_args)
    pool = Pool(processes=args.count)
    for x in range(0, args.count):
        result = pool.apply_async(run_client, [x, jacktrip_args])
        if args.sleep > 0:
            time.sleep(args.sleep)
    pool.close()
    pool.join()

def run_client(id, jacktrip_args):
    my_args = copy.deepcopy(jacktrip_args)
[...]

PR #98, opened 2020-07-27
Fix for hubserver failure with concurrent connections

mikedickey:
TcpServer was not checking for pending connections before waiting to receive new ones. This was causing it to fail when more than 1 client connected to a hub server at the same time.

Also bumping pthreads limit -- this was tested at scale of 500 with no issues


mikedickey:
Fixes #87

PR #99, opened 2020-07-28
Too many hubs

psiborg112:
Was just merging some changes from the master back into my branch and noticed that we may have been a little overzealous with the renaming of hub to master. It's affected a few of the external libraries which I've reverted back here. It may also have affected the binary file libjack.dll. (So if anyone has an appropriate copy of that handy, it might not hurt to reupload it because we may have broken it.)

### DIFF
--- a/externals/includes/rtaudio-4.0.7/RtAudio.cpp
+++ b/externals/includes/rtaudio-4.0.7/RtAudio.cpp
@@ -435,7 +435,7 @@ unsigned int RtApiCore :: getDeviceCount( void )
 {
   // Find out how many audio devices there are, if any.
   UInt32 dataSize;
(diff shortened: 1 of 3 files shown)


cchafe:
The rename (using SED) replaced all occurences of "master" with "hub" -- I can't remember if I did that recursively down into /externals -- would that have been the cause?

psiborg112:
That would have done it if it was recursive.

cchafe:
Did it break a compile? The current version of master compiles ok.

psiborg112:
It would only break if someone was actually building using the externals that we're supplying. (And most people probably aren't.) And it's kind of moot if we decide to do away with the externals folder entirely.

ntonnaett:
I would support the decision to remove the externals folder. The library versions seem to be pretty outdated. AFAIK we don't encourage to use it on any platform. Please correct me if I'm wrong.

psiborg112:
This is resolved in the dev branch with the removal of the externals folder, so I'm closing this now.

Issue #102, opened 2020-07-29
issue a with long JACK client name


dyfer:
Hello,
When reporting #73 I noticed that there is an issue with long JACK client names. Originally this was associated with the fact that `--clientname` was broken and a (potentially long) hostname would be used as the client name. The limit for the client name seems to be 27 characters on my system.

This is happening on macOS 10.13 with the current build from `master` branch and JACK2 installation from the JackOSX package. As I [noted before](https://github.com/jacktrip/jacktrip/issues/73#issuecomment-647866990), my JACK reports maximum client name size as 65.

Here's a reproducer:
this fails (28 characters)
```
$ ./jacktrip -c localhost --clientname 123456789.123456789.12345678
Cannot open 123456789.123456789.12345678 client
JackShmReadWritePtr1::~JackShmReadWritePtr1 - Init not done for -1, skipping unlock
JackShmReadWritePtr::~JackShmReadWritePtr - Init not done for -1, skipping unlock
JackShmReadWritePtr::~JackShmReadWritePtr - Init not done for -1, skipping unlock
ERROR:
Maybe the JACK server is not running?
Exiting JackTrip...
```

this works (27 characters)
```
$ ./jacktrip -c localhost --clientname 123456789.123456789.1234567
Setting JACK Process Callback...
SUCCESS
---------------------------------------------------------
The Sampling Rate is: 48000
---------------------------------------------------------
The Audio Buffer Size is: 128 samples
                      or: 512 bytes
---------------------------------------------------------
The Number of Channels is: 2
---------------------------------------------------------
The RTAudio device ID is: 0
---------------------------------------------------------
Using UDP Protocol
---------------------------------------------------------
Peer Address set to: localhost
---------------------------------------------------------
UDP Socket Receiving in Port: 4464
---------------------------------------------------------
Waiting for Peer...
Received Connection from Peer!
```

ntonnaett:
We use char* for mJackClientName. I don't understand why that works at all. Maybe we allocate with the generic name and then we get more because of cache alignment? I think we should switch to QString or std::string.

dyfer:
I was trying to track this down. I started passing hardcoded strings right when initializing JACK client and indeed this seems to be an issue with JACK itself. Once I went over 63 characters (so 2 less than the value from `jack_client_name_size()`), there was an explicit message about the client name being too long. For anything between 28 and 63 characters I was getting the error as reported in this issue.
[...]

PR #106, opened 2020-07-30
rework handling of JACK client name

dyfer:
Per discussion in #102, here's a proposal to use `QString` for JACK client name.

### DIFF
--- a/src/JackTrip.cpp
+++ b/src/JackTrip.cpp
@@ -159,14 +159,14 @@ void JackTrip::setupAudio(
 #ifdef WAIRTOHUB // WAIR
         //Set our Jack client name if we're a hub server or a custom name hasn't been set
 	      if ( mPeerAddress.toStdString() != "" &&
-	          (mJackClientName == gJackDefaultClientName || mJackTripMode == SERVERPINGSERVER)) {
-            mJackClientName = QString(mPeerAddress).replace(":", ".").toLatin1().constData();
+	          (mJackClientName.constData() == gJackDefaultClientName.constData() || mJackTripMode == SERVERPINGSERVER)) {
+	          mJackClientName = QString(mPeerAddress).replace(":", ".");
         }
 //        std::cout  << "WAIR ID " << ID << " jacktrip client name set to=" <<
 //                      mJackClientName << std::endl;
 #endif // endwhere
 
-        mAudioInterface->setClientName(mJackClientName);
+        mAudioInterface->setClientName(mJackClientName.toLatin1().constData());
 
         if (gVerboseFlag) std::cout << "  JackTrip:setupAudio before mAudioInterface->setup" << std::endl;
         mAudioInterface->setup();
--- a/src/JackTrip.h
+++ b/src/JackTrip.h
@@ -227,7 +227,7 @@ public:
     }
     /// \brief Set Client Name to something different that the default (JackTrip)
     virtual void setClientName(const char* ClientName)
-    { mJackClientName = ClientName; }
+    { mJackClientName = QString(ClientName); }
     /// \brief Set the number of audio channels
     virtual void setNumChannels(int num_chans)
     { mNumChans = num_chans; }
@@ -500,7 +500,7 @@ private:
     int mTcpServerPort;
 
     unsigned int mRedundancy; ///< Redundancy factor in network data
-    const char* mJackClientName; ///< JackAudio Client Name
+    QString mJackClientName; ///< JackAudio Client Name
 
     JackTrip::connectionModeT mConnectionMode; ///< Connection Mode
     JackTrip::hubConnectionModeT mHubConnectionModeT; ///< Hub Server Jack Audio Patch Connection Mode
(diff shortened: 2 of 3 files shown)


src/JackTrip.cpp
```diff
@@ -159,14 +159,14 @@ void JackTrip::setupAudio(
 #ifdef WAIRTOHUB // WAIR
         //Set our Jack client name if we're a hub server or a custom name hasn't been set
 	      if ( mPeerAddress.toStdString() != "" &&
-	          (mJackClientName == gJackDefaultClientName || mJackTripMode == SERVERPINGSERVER)) {
-            mJackClientName = QString(mPeerAddress).replace(":", ".").toLatin1().constData();
+	          (mJackClientName.constData() == gJackDefaultClientName.constData() || mJackTripMode == SERVERPINGSERVER)) {
+	          mJackClientName = QString(mPeerAddress).replace(":", ".");
```
dyfer, line 163: I tried to match mixed whitespace here (tab + spaces). Should I fix this block to use spaces only?

psiborg112: Yeah, I think all spaces is the way to go. And I don't think a pointer comparison will work here to allow JackTrip as a name because I don't think they're likely to return a pointer to the same memory address.

ntonnaett:
Maybe we should dig a little bit deeper. What ClientName in JackAudioInterface is const char*. Is it because the jack library expects that? If this is the reason we should convert at this interface level. I don't like QString.toLatin1() because it converts "unknown" characters to a question mark. If jack really wants Latin1 then we should check if there are unknown characters an deny the setting of the name.
std::string has a method for converting to cstrings. Maybe that's more convenient when working with jack.h. But I don't know what the pros and cons of QStrings are.

src/JackTrip.cpp
dyfer: I tested and it seemed to work... when that comparison is true, the client name is replaced by the host name, and that seems to work. If we set `mJackClientName` [here](https://github.com/jacktrip/jacktrip/blob/7e7309cc15a41c54defd06803de2ee83e7cf8ecc/src/JackTrip.cpp#L109) wouldn't it point to the same data?

dyfer:
> Maybe we should dig a little bit deeper. What ClientName in JackAudioInterface is const char*. Is it because the jack library expects that? 

Yes, that's my understanding.

> If this is the reason we should convert at this interface level. 

Do you mean that `JackAudioInterface` should take a QString and convert before calling jack functions?

> I don't like QString.toLatin1() because it converts "unknown" characters to a question mark. If jack really wants Latin1 then we should check if there are unknown characters and deny the setting of the name.

I agree that `toLatin1()` is not great - I used it since it was used elsewhere in the code, albeit for host names so they're less likely to contain unknown characters. 

Now I've also tried `toUtf8()` and that seems to work, including using non-western characters for client name (which are then reported properly in jack). Would that be better?

> std::string has a method for converting to cstrings. Maybe that's more convenient when working with jack.h. But I don't know what the pros and cons of QStrings are.

I've used `QString` primarily for consistency, since it's used elsewhere in this code.

src/JackTrip.cpp
psiborg112: If it works, it works. I thought the QString constructor made a copy of the internal constData that's returned, but it seems like it doesn't.

src/JackTrip.cpp
```diff
@@ -158,15 +158,15 @@ void JackTrip::setupAudio(
...
+        if ( mPeerAddress.toStdString() != "" &&
+            (mJackClientName.constData() == gJackDefaultClientName.constData() || mJackTripMode == SERVERPINGSERVER)) {
+            mJackClientName = QString(mPeerAddress).replace(":", ".");
         }
 //        std::cout  << "WAIR ID " << ID << " jacktrip client name set to=" <<
 //                      mJackClientName << std::endl;
 #endif // endwhere
 
-        mAudioInterface->setClientName(mJackClientName);
+        mAudioInterface->setClientName(mJackClientName.toUtf8().constData());
```
psiborg112, line 169: So I found out the hard way that this is a bad idea. The pointer returned by constData() is only guaranteed to be valid while the QByteArray returned by mJackClientName.toUtf8() still exists. And since we don't keep a reference to it, that isn't long. So we either need to keep a reference, or the better solution (and what I've done in the qjackctl branch) is to have the mClientName variable in JackAudioInterface be a QString and only make the toUtf8().constData() call when passing it as an argument to jack_client_open.

(It should be safe to use the version of JackAudioInterface.cpp and .h from the qjackctl branch because I don't think I've made any other crazy changes to them, but I'd need to double check that. From memory you'd also need to pull in AudioInterface.h as well.)

dyfer: It crossed my mind this could be a problem.
[Few lines above](https://github.com/jacktrip/jacktrip/pull/106/files#diff-67ee6d5b8fb8b9a04f9fea4980ac7d56R163) we also do  `mJackClientName = QString(mPeerAddress).replace(":", ".");`. Would that also need to change?

psiborg112: No, that should be fine because we're keeping a reference to the result as a member variable.

PR #108, opened 2020-07-31
Set QT5 dependency in build script

driechers:
Updated build script to select QT5 build tools when 4 and 5 are installed. Resolves #105

PR #110, opened 2020-08-02
Revert "Set QT5 dependency in build script"

cchafe:
Reverts jacktrip/jacktrip#108
Creates a new dependency without which the build fails
Reports are "qtchooser: command not found"


ntonnaett:
I think the alternative would be to have hardcoded paths for Debian/Ubuntu.

Issue #56, opened 2020-05-16
target.path should be $${PREFIX}/bin/ in .pro file


yurivict:
https://github.com/jacktrip/jacktrip/blob/master/src/jacktrip.pro#L112

```
-target.path = /usr/bin
+target.path = $${PREFIX}/bin/
```

Foaly:
Could you please send a PR? Thanks :)

PR #111, opened 2020-08-02
Define PREFIX variable for installation path (Defaults to /usr/local)

ntonnaett:
- This forwards all parameters of the build script to qmake
- './build PREFIX=/usr; make install' installs jacktrip in /usr/bin
- './build; make install' installs jacktrip in /usr/local/bin, so it doesn't conflict with a packaged installation of jacktrip


ntonnaett:
Fixes #56

cchafe:
We're keeping "main" pristine and adding changes only to "dev", want to redirect this?

Issue #117, opened 2020-08-06
Windows: segfault in UdpDataProtocol::receivePacket (QUdpSocket)


tangmi:
Hi, I'm trying to run JackTrip v1.2.1 in hub mode on Windows 10 version 2004 (`jacktrip -S`) and have ports 4464 and 61000 open. When a client connects with `jacktrip -C <ip address>`, the server crashes almost immediately.

I've debugged the program in Qt Creator (qt 5.15.0, mingw 64-bit 8.1.0) and get a segfault somewhere on this line:: https://github.com/jacktrip/jacktrip/blob/v1.2.1/src/UdpDataProtocol.cpp#L323

The stack is as follows:

```
1 QNativeSocketEnginePrivate::nativeReceiveDatagram(char *, long long, QIpPacketHeader *, QFlags<QAbstractSocketEngine::PacketHeaderOption>)                         0x697aeb19     
2 QNativeSocketEngine::readDatagram(char *, long long, QIpPacketHeader *, QFlags<QAbstractSocketEngine::PacketHeaderOption>)                                         0x697ac5f1     
3 QUdpSocket::readDatagram(char *, long long, QHostAddress *, unsigned short *)                                                                                      0x697a0880     
4 UdpDataProtocol::receivePacket                                                                                                             UdpDataProtocol.cpp 323 0x40bfb1       
5 UdpDataProtocol::run                                                                                                                       UdpDataProtocol.cpp 521 0x40c599       
6 QThreadPrivate::start(void *)                                                                                                                                      0x68898eaa     
7 KERNEL32!BaseThreadInitThunk                                                                                                                                       0x7ff82cf76fd4 
8 ntdll!RtlUserThreadStart                                                                                                                                           0x7ff82d6dcec1 
9 ??                                                                                                                                                                                
```

The packet size `n` is 2064 and `buf` seems valid, so I'm unsure how to dig deeper since I am unable to step into Qt sources on Windows with MinGW (I think since it uses GDB instead of CDB). I'm not familiar with Qt Creator so please let me know if there's something I can do to step into the sources!

psiborg112:
The windows version has always been a little touchy after I made changes to the networking to enable IPv6. There were a few odd bugs that were hard to track down, and I suspect there's still something about the way we're using windows sockets and the way that Qt is using windows sockets that is conflicting. Are you able to try a build of the 'aaron' branch to see if the bug persists? (All of the QUdpSocket code has been replaced with winsock code.)

[...]

PR #119, opened 2020-08-06
Remove dependency on QUdpSocket in UdpDataProtocol

psiborg112:
Only use bsd socket and winsock functions instead of the current mix of these and QUdpSocket. (Some bad interaction between them seems to be causing issues on windows.)

### DIFF
--- a/src/UdpDataProtocol.cpp
+++ b/src/UdpDataProtocol.cpp
@@ -383,9 +383,9 @@ void UdpDataProtocol::getPeerAddressFromFirstPacket(QHostAddress& peerHostAddres
     ::recvfrom(mSocket, buf, 1, 0, (struct sockaddr*) &addr, &sa_len);
     peerHostAddress.setAddress((struct sockaddr*) &addr);
     if (mIPv6) {
-        port = ((struct sockaddr_in*) &addr)->sin_port;
-    } else {
         port = ((struct sockaddr_in6*) &addr)->sin6_port;
+    } else {
+        port = ((struct sockaddr_in*) &addr)->sin_port;
     }
 }
 
--- a/src/UdpDataProtocol.cpp
+++ b/src/UdpDataProtocol.cpp
@@ -53,6 +53,8 @@
 #endif
 #if defined (__LINUX__) || (__MAC_OSX__)
 #include <sys/socket.h> // for POSIX Sockets
+#include <unistd.h>
+#include <sys/fcntl.h>
 #endif
 
 using std::cout; using std::endl;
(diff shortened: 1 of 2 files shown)


psiborg112:
Meant to add that this seems to resolve #117.

src/UdpDataProtocol.cpp
```diff
@@ -358,15 +369,24 @@ int UdpDataProtocol::sendPacket(const char* buf, const size_t n)
...
+    struct sockaddr_storage addr;
+    std::memset(&addr, 0, sizeof(addr));
+    socklen_t sa_len = sizeof(addr);
+    ::recvfrom(mSocket, buf, 1, 0, (struct sockaddr*) &addr, &sa_len);
+    peerHostAddress.setAddress((struct sockaddr*) &addr);
+    if (mIPv6) {
+        port = ((struct sockaddr_in*) &addr)->sin_port;
+    } else {
+        port = ((struct sockaddr_in6*) &addr)->sin6_port;
+    }
```
ntonnaett, line 389: Is this the right way around?

psiborg112: It is not - good catch.

src/UdpDataProtocol.cpp
```diff
@@ -507,18 +525,17 @@ void UdpDataProtocol::run()
...
         }
-        int first_packet_size = UdpSocket.pendingDatagramSize();
         // The following line is the same as
-        int8_t* first_packet = new int8_t[first_packet_size];
+        int8_t* first_packet = new int8_t[full_redundant_packet_size];
         /// \todo fix this to avoid memory leaks
         // but avoids memory leaks
         //std::tr1::shared_ptr<int8_t> first_packet(new int8_t[first_packet_size]);
-        receivePacket( UdpSocket, reinterpret_cast<char*>(first_packet), first_packet_size);
+        receivePacket( reinterpret_cast<char*>(first_packet), full_redundant_packet_size);
```
ntonnaett, line 538: I thought the first packet is just the handshake. Why do we need full redundant packet size at this point? Actually those packet size names are not clear. But that's for another day.

psiborg112: We don't. Technically we could just use the header size and let the rest of the packet go, but reading the header size would have meant adding an additional getter in the JackTrip class. And this way we can just allocate a single buffer for both the initial packet and all of the subsequent packets.

psiborg112: (It's worth noting that the code that this replaces used to read the full first packet into the buffer as well, so this is no less efficient than the previous code. And the first packet isn't any smaller than the packets that follow it - there's never a header sent by itself without audio data.)

cchafe:
The new idea is to leave "main" pristine and equal to the stable release which has a tag (in this case, v1.2.1)
Want to merge it to "dev" instead?

psiborg112:
Happy to merge to dev, but I think we should try to test this as much as possible and shift it to main quickly because it's a pretty bad bug. I think it's a hangover from the previous issues we were having, and that our workaround only works for certain qt and mingw versions. (If it helps, I recently ran a hub server and one of the hub clients in a 24 hour Berlin/Perth/Melbourne livestream on this code without issue.)

psiborg112:
I think I need to close this to merge it to dev, so will do that.

PR #124, opened 2020-08-08
Merge changes from Aaron into dev branch

psiborg112:
So this is a big change, but it does address a few of the issues or feature requests that have been reported of late. Before merging the existing dev changes into it, I'd been using the GUI version based off of this branch almost exclusively for my own projects, including a 24 hour livestream. So it's received quite a bit of testing, but could always do with more. (And should definitely be tested to make sure it doesn't break any of the changes from dev that I just merged into it, like multiple hub servers.) There are also features here that have added to the networking protocol, so we should definitely have a discussion about how we want to proceed with those long term. (Backwards compatibility is maintained.)

- Added SIGINT and SIGTERM handler so that JackTrip can exit gracefully. (Equivalent CtrlHandler implemented for windows.)
- Add \_\_attribute\_\_((unused)) to get rid of unused parameter warnings for parameters that are unlikely to ever be used in the future.
- Use QString for mClientName in JackAudioInterface (and elsewhere). (Greater consistency with the rest of the code, and makes it harder to introduce memory bugs through additional conversions. Looking at me there.)
- Changed the way IPv6 addresses are converted to mClientName in JackTrip. (Using '_' instead of '.' as a replacement for ':' as it makes it easier to check for a mapped IPv4 client name in other software that may read this generated name later.)
- Removed extra JMess.cpp file in root folder.
- Use asynchronous network calls in JackTrip.cpp for the initial connection. This allows the startProcess function to be called from a GUI without explicitly creating an additional thread to prevent blocking.
- Pass IOStatLogStream down the line to JackTrip using QSharedPointers. (Removes UdpHubListener’s dependency on Settings.h)
- Run JackTrip (or UdpHubListener as appropriate) from main rather than Settings.
- Introduce timeout (-t) flag to allow JackTrip to automatically quit after 10 seconds of no network activity (in server, client or hubclient mode).
- Introduce --remotename switch to allow a hub client to set the JackTrip client name on the hub server. (Currently done using a maximum 64 byte UTF8 string, inclusive of null terminator, sent immediately following the port in the same buffer. We should make sure this is a protocol we want to use for this.)
- Derive JackTrip from QObject instead of QThread since it doesn’t actually extend QThread. (The same has been done for Settings.) Commented out lines in the unused JackTripThread.cpp to allow it to build. (Alternatively could remove this file from our build settings.)
- Added NOAUTO patch mode as synonym for --nojackportsconnect in Hub Listener mode (since this is where users are more likely to look for this setting.)
- Change UdpDataProtocol to use only bsd socket or winsock calls. There still seems to be an issue with the current mix of these and QUdpSocket calls introduced by me for IPv6 support. (Segfault in windows with Qt 5.15.0 and mingw 64-bit 8.1.0.)
- Use a 64 byte control packet in UdpDataProtocol to signal that a client is exiting. (Currently just setting every byte to 0xff and checking for that, but we could establish an actual protocol for messages that we want to send over the udp port in this way.)
- Changed NULL to 0 for mServerUdpPort default value to avoid compiler warnings.


psiborg112:
Damn it! This was meant to be a merge into dev not main. Will redo.

PR #130, opened 2020-08-08
Update install txt

wholenotes:
updated build, installation and verification instructions for OSX, Linux and Windows


cchafe:
Is it possible to restructure just a bit? I'm not sure we should show "nojack" since that's deprecated along with rtaudio

6) dependencies (for all operating systems)
11) Qt 5.14.2
13) note: rtaudio is no longer a dependency

        ...then... all linux 
-- fedora (show the dnf commands) ./build or QtCreator
-- debian / raspbian (show apt) ./build or QtCreator
i- ubuntu (show apt)
-- any others (show yum)

        ...then... macOS ./build or QtCreator (correct me?)

        ...then... windows 10 QtCreator (only)

wholenotes:
If nojack is deprecated, should we explicitly list Jack as a dependency after Qt 5.14.2?

cchafe:
It's there (line 12)

wholenotes:
Pushed some updates.

‐‐‐‐‐‐‐ Original Message ‐‐‐‐‐‐‐

cchafe (approved):

PR #132, opened 2020-08-09
Merge dev into jos

mikedickey:
Only conflict that required manual resolution was in Settings.cpp -- please review especially that one


josmithiii:
Since I cannot commit to dev, I'm creating jos2 and deleting jos.  Gotta be able to push somewhere.

PR #137, opened 2020-08-13
jos branch Aaron/Anton/JOS bugfix PR

josmithiii:
Bugfix PR for commit Jos (#135): 

Aaron: 
  - allocate Faust audio effects during mJackTrip client creation only (not server).
  - do not deallocate effects in Effects.cpp because AudioInterface.cpp is doing it.
  - return 0 upon successful parsing of --assumednumclients
  - reduce compiler warnings

Anton:
  - load allocated audio effects during mJackTrip client creation only (not server).

jos: 
 - add explanatory comments
 - reduce compiler warnings

### DIFF
--- a/src/Effects.h
+++ b/src/Effects.h
@@ -69,7 +69,7 @@ class Effects
   Reverb* inFreeverbP = nullptr;
   Reverb* outFreeverbP = nullptr;
   int parenLevel = 0;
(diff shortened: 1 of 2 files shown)


psiborg112:
Looks good

josmithiii:
Chris approved via email.

PR #144, opened 2020-08-21
Fix nojack build

kltdwrds:
The introduction of `JMess` broke `build nojack` builds - This PR fixes the issue by using the existing `__NO_JACK__` definition to avoid referencing `JMess` when building with `nojack`.


kltdwrds:
Ah I hadn't checked dev recently, and after a closer look it does seem to be broken when trying to run the nojack build.  Thanks for the heads up.  

Closing for now.

kltdwrds:
I'd be curious to know if there are plans to re-implement something like this in the future.

Cheers!

PR #151, opened 2020-08-28
studying new -x

omarcostahamido:
small commit to help enforce consistency on the printed messages.


omarcostahamido:
I'll suggest adding a note Settings.cpp explaining that this only works in client-server mode and not hub mode.
And also perhaps mention (in there and in AudioInterface.cpp, i.e. when jacktrip is started) that it will send sound to client automatically as soon as the client connects.

josmithiii:
Thanks for the polish

josmithiii:
Please feel free to add as much clarity as you feel the need for.

I used to assume that the JACK audio callback would get called in the hub server, because I heard people were adding local JACK processing at the hub.  However, I have no experience with the hub side of JackTrip.

PR #152, opened 2020-08-29
Add -x (--examine-audio-delay) feature

josmithiii:
Adds a new command line option "-x printInterval" which runs a jacktrip client in a special test mode for measuring audio latency down to sample accuracy, printing the mean and standard deviation in ms at the specified time interval in seconds.

### DIFF
--- a/src/AudioInterface.cpp
+++ b/src/AudioInterface.cpp
@@ -237,6 +237,14 @@ void AudioInterface::callback(QVarLengthArray<sample_t*>& in_buffer,
     computeProcessFromNetwork(out_buffer, n_frames);
     // =============================================
 
+    // out_buffer is from the network and goes "out" to local audio
+    // hardware via JACK:
+
+    // mAudioTesterP will be nullptr for hub server's JackTripWorker instances
+    if (mAudioTesterP && mAudioTesterP->getEnabled()) {
+      mAudioTesterP->lookForReturnPulse(out_buffer, n_frames);
+    }
+
 #ifdef WAIR // WAIR
     // nib16 result now in mNetInBuffer
 #endif // endwhere
@@ -276,7 +284,11 @@ void AudioInterface::callback(QVarLengthArray<sample_t*>& in_buffer,
     // compute cob16
 #endif // endwhere
 
-    // Run Faust plugins for the outgoing stream:
+    // mAudioTesterP will be nullptr for hub server's JackTripWorker instances
+    if (mAudioTesterP && mAudioTesterP->getEnabled()) { // in_buffer is "in" from local audio hardware via JACK
+      mAudioTesterP->writeImpulse(mInBufCopy, in_buffer, n_frames);
+      computeProcessToNetwork(mInBufCopy, n_frames);
+    } else { // Run Faust plugins for the outgoing stream:
     int nop = mProcessPluginsToNetwork.size(); // number of OUTGOING processing modules
     if (nop>0) { // cannot modify IN_BUFFER so make a copy
       //#ifdef DEBUG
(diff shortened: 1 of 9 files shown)


josmithiii:
This needs to go in before I can make a clean feature-request PR containing the new effects parameter-specification and help system.  I have tested -c, -C, and -S, but still do not know how to test -s (or whether that's even necessary after testing the other three cases).  I don't think it can be an issue because jacktrip will "exit 1" if one tries to use -x with either -s or -S.

josmithiii:
Since there does not seem to be any interest in this new feature, I am closing the pull request.  It remains available in the jos branch.

The expanded help system for effects is also only in the jos branch.  If there is any desire for that, please ask someone who can to `git checkout dev && git merge --squash jos`.

I decided not to make a new Faust gate module, because I don't like audio gates myself, and using it to suppress silent channels (effectively providing mono outgoing and stereo incoming) was not of interest to the group, according to the IETF email thread on that.

src/JackTrip.cpp
```diff
@@ -211,6 +212,10 @@ void JackTrip::setupAudio(
     }
 
     mAudioInterface->setLoopBack(mLoopBack);
+    if (mAudioTesterP) { // if we're a hub server, this will be a nullptr - MAJOR REFACTOR NEEDED, in my opinion
+      mAudioTesterP->setSampleRate(mSampleRate);
+    }
+    mAudioInterface->setAudioTesterP(mAudioTesterP);
```
mikedickey, line 218: should this be inside the if{} block? what is the impact of setting this to a nullptr when in hub server mode?

src/Settings.cpp
```diff
@@ -632,12 +661,15 @@ JackTrip *Settings::getConfiguredJackTrip()
...
+
     if (mIOStatTimeout > 0) {
         jackTrip->setIOStatTimeout(mIOStatTimeout);
         jackTrip->setIOStatStream(mIOStatStream);
     }
 
+    jackTrip->setAudioTesterP(&mAudioTester);
+
+    if (not mAudioTester.getEnabled()) { // No effects plugins allowed while testing:
     // Allocate audio effects in client, if any:
```
mikedickey, line 673: fix indenting?

src/AudioInterface.cpp
```diff
@@ -276,7 +284,11 @@ void AudioInterface::callback(QVarLengthArray<sample_t*>& in_buffer,
     // compute cob16
 #endif // endwhere
 
-    // Run Faust plugins for the outgoing stream:
+    // mAudioTesterP will be nullptr for hub server's JackTripWorker instances
+    if (mAudioTesterP && mAudioTesterP->getEnabled()) { // in_buffer is "in" from local audio hardware via JACK
+      mAudioTesterP->writeImpulse(mInBufCopy, in_buffer, n_frames);
+      computeProcessToNetwork(mInBufCopy, n_frames);
+    } else { // Run Faust plugins for the outgoing stream:
     int nop = mProcessPluginsToNetwork.size(); // number of OUTGOING processing modules
```
mikedickey, line 292: fix indenting?

mikedickey (commented):
This looks like a pretty clean merge / impact if protected via feature flags. Happy to approve if you'd still like to merge it.

src/JackTrip.cpp
josmithiii: AudioInterface needs the null pointer so it can check it before using it.  This has been tested in hub server mode.

src/Settings.cpp
josmithiii: Sure, I was just trying to minimize the diffs.  What I normally do is a minimized change like this followed by a separate commit flagged as "whitespace/formatting only", but given how hard it is to contribute commits in this system, I suppose it is better to go directly for one commit, even though it's more of a mess.

src/AudioInterface.cpp
josmithiii: OK

src/Settings.cpp
josmithiii: On second thought, a pull request can technically contain any number of commits, so maybe the right process is to submit a multi-commit pull-request equivalent to the current result of a single-commit PR followed by a few update commits, designed for maximum clarity.

src/JackTrip.cpp
josmithiii: Furthermore, -x is not allowed in server modes (jacktrip exits when -S or -s is used with -x).

PR #153, opened 2020-08-31
Add more network diagnostic output to verbose mode

madwort:
* dump lots of info about the contents of the UDP packets
* I'm interested in the behaviour of the RingBuffer
* Nb. you should use larger jack frame sizes (128+) to avoid catastrophically huge output files

### DIFF
--- a/src/JackTrip.h
+++ b/src/JackTrip.h
@@ -319,7 +319,8 @@ public:
     virtual void sendNetworkPacket(const int8_t* ptrToSlot)
     { mSendRingBuffer->insertSlotNonBlocking(ptrToSlot); }
     virtual void receiveNetworkPacket(int8_t* ptrToReadSlot)
-    { mReceiveRingBuffer->readSlotNonBlocking(ptrToReadSlot); }
+    { mReceiveRingBuffer->readSlotNonBlocking(ptrToReadSlot);
+      if (gVerboseFlag) mReceiveRingBuffer->debugDump(); }
     virtual void readAudioBuffer(int8_t* ptrToReadSlot)
     { mSendRingBuffer->readSlotBlocking(ptrToReadSlot); }
     virtual void writeAudioBuffer(const int8_t* ptrToSlot)
--- a/src/RingBuffer.cpp
+++ b/src/RingBuffer.cpp
@@ -247,10 +247,12 @@ void RingBuffer::overflowReset()
 //*******************************************************************************
 void RingBuffer::debugDump() const
 {
-    cout << "mTotalSize = " << mTotalSize << endl;
-    cout << "mReadPosition = " << mReadPosition << endl;
+    cout << "mTotalSize = " << mTotalSize << " ; ";
+    cout <<  "mFullSlots = " << mFullSlots << " ; ";
+    cout <<  "mUnderruns = " << mUnderruns << " ; ";
+    cout <<  "mOverflows = " << mOverflows << " ; ";
+    cout << "mReadPosition = " << mReadPosition << " ; ";
     cout << "mWritePosition = " << mWritePosition << endl;
-    cout <<  "mFullSlots = " << mFullSlots << endl;
 }
 
 //*******************************************************************************
--- a/src/RingBuffer.h
+++ b/src/RingBuffer.h
@@ -107,6 +107,8 @@ public:
         uint32_t overflows;
     };
     virtual bool getStats(IOStat* stat, bool reset);
+    /// \brief Helper method to debug, prints member variables to terminal
+    void debugDump() const;
 
 protected:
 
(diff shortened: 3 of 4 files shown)


antonrunov:
My 2 cents here. First, the RingBuffer is going to be replaced with JitterBuffer (WIP in my branch https://github.com/antonrunov/jacktrip/tree/jitter-buffer, hopefully it will be finalized and moved to the main repo soon). Second, most of the time the direct output from the real-time thread ticking each few milliseconds might not be very helpful. It's very difficult to process (as it's huge), and more importantly it will likely affect the real-time behavior. That's why I added `--iostat` option to dump statistic metrics instead of instant values.

madwort:
Hey @antonrunov , thanks for the feedback. Just been checking out your PR, and I notice that your 'strategy 1' is what I was going to propose. When do you think your PR is going to be ready to merged? Is it worth doing any work in the meantime, or we should just wait?

So, as I guess you've figured out this out, but just to explain the thinking. The issue is that when the buffer underruns it skips one frame, but when it overflows is skips (buf/2)+1 frames. I'm experiencing brief delays, but with no packet loss and with no packet re-ordering. So in (example) scenario `jacktrip -q10`:
* the buffer is typically ~7 units (6-8) full
* 16 packets are delayed
* underruns by 8
* the packets arrive
* overflow by two increments (12)
* now the buffer is typically ~3 (2-4) full
* this is what your iostats would be reporting as -4.
We then need to wait for more underruns to refill the queue - but also the skew is going to be the cause for the underruns - if we had not decreased the buffer so quickly we wouldn't have seen these secondary underruns at all. 

I realise that this kind of logging can affect behaviour, but in my tests you can directly observe this scenario in the output, and I was interested to see whether others experienced the same. With iostats you can see the skew, but I'm interested in the practical distribution of runs of delayed packets, which I can observe from these files, and also looking at quantifying how many underruns are due to excess skew due to the overflow skip.

antonrunov:
Hi @madwort, you are absolutely right about this issue in RingBuffer. Moreover, it skips the current frame on overruns instead of dropping only the oldest ones. It's all addressed in my JitterBuffer. I plan to push updates to my branch this week (hopefully today/tomorrow but it depends). I also need to merge the dev branch into it to prepare a PR, I don't think there should be significant conflicts though.
I get your point about obtaining the distribution of events. I was thinking about it as well, I agree that this would be very useful information for better understanding of JB work and any further optimization. But it's kind of complicated. Typically, you don't want doing such logging directly from the real-time thread. Instead, you save the log data in some extra buffer and periodically write it from a normal thread. I'd also prefer logging into a separate file rather then stout/stderr (to keep the normal output readable), and use binary or easily parsable format (because you will definitely need some postprocessing to analyze/visualize such amount of information).

madwort:
> Hi @madwort, you are absolutely right about this issue in RingBuffer. Moreover, it skips the current frame on overruns instead of dropping only the oldest ones. It's all addressed in my JitterBuffer. I plan to push updates to my branch this week (hopefully today/tomorrow but it depends). I also need to merge the dev branch into it to prepare a PR, I don't think there should be significant conflicts though.

Ok, thanks - if it was a month or two I was thinking I'd address bits of that in the current buffer, but if it's a week or two then I'll wait!

> I get your point about obtaining the distribution of events. I was thinking about it as well, I agree that this would be very useful information for better understanding of JB work and any further optimization. But it's kind of complicated. Typically, you don't want doing such logging directly from the real-time thread. Instead, you save the log data in some extra buffer and periodically write it from a normal thread. 

Yes of course, but in this case I think we're looking for pretty large scale things (at the frame sizes I'm running, I'm looking at issue at the multiple msecs kind of level, not the usecs thread-conflict kind of level) - and we're also only actually interested in the network traffic at the moment - I don't think it matters for the purposes of this test if the audio thread gets delayed. But you're right, that is the correct way to approach it.

> I'd also prefer logging into a separate file rather then stout/stderr (to keep the normal output readable), and use binary or easily parsable format (because you will definitely need some postprocessing to analyze/visualize such amount of information).

Yeah, all good thoughts, but more than I have time to implement - was mostly curious to have a scout out to see how common that particular scenario is rather than a full data analysis. 

One first step visualisation that would be really useful to me, in terms of organising jacktrip sessions, would be to keep tabs of the network runs (including those ones solely compensated for by the buffer) and output something like a histogram at the end of a session. At the moment, we just try running JT, see how it sounds, change the queue length, etc, until you've reached the best balance of latency and glitches. But if you could actually see at the end of a session that (e.g.) your buffer was size 10, should be 7 full, nearly all of your network runs were less than 4, or maybe they were mostly 8 or less, or maybe nearly all of them are less than 2 but when they do happen they're 20, or whatever, that would really help make more honed setups more quick to do - you could just set it arbitrarily, play for 5 mins while you do a sound-check & then get a read-out with a decent estimation of what size queue you should be using.

josmithiii:
Do you mean to merge into dev instead of main?

ntonnaett:
You don't need to close this to merge into dev. Just edit the PR beside the title and change the base.

madwort:
Not planning to merge this anywhere, this was an experimental branch for testing, not for production use.

Issue #134, opened 2020-08-11
Cannot connect to jack server even though server running


afterlifenirvana:
I have installed jacktrip on a ubuntu server. My jack server is running with server name jacktrip-mylocation. If the server name is set to default jacktrip is able to connect to jack server if i set it to anything else I cannot connect to it. I wanted to change name of the server for identification.

ntonnaett:
I just stumbled upon something that might fix this. Could you test #156 ?

ntonnaett:
Fixed in #156

PR #156, opened 2020-09-08
Don't nullptr jack server name when creating jack client

ntonnaett:
Setting the jack server name null may be unintended here.


ntonnaett:
This seems to fix #134 .
Serverside:
``` 
jackd -r -n jtserver -d dummy -p 128 &
export JACK_DEFAULT_SERVER=jtserver
./jacktrip -S -p 1
``` 
One client connected without any problems.

wholenotes (approved):

PR #157, opened 2020-09-09
Split library and client

ntonnaett:
This is the first part of splitting jacktrip into a client and a library. The meta issues is #159 . There is the place for discussing the broader endeavour.

### What does this PR
- split in client and library
- library is subproject
- both build with meson
- library can be used from system or locally
- meson generates a pkgconfig file
- Travis builds for Linux and MacOS
- include statements changed, e.g. #include <jacktrip/JackTrip.h> (maybe we don't need a subdirectory in the future)

### What is missing?
- audio parts are still in library

### How to build
I've added a **libjacktrip subproject**. You can (but you don't have to) build and install it with:
```
cd subproject/libjacktrip/
meson builddir
cd builddir
meson install
```
This should compile, generate a pkgconfig file and install the library, headers and the pkgconfig file in /usr/local/… .

To build jacktrip you can type in the jacktrip root directory:
```
meson builddir
ninja -C builddir
```
Most certainly it couldn't find the installed library and just uses the fallback that is the subproject. If it could find the pkgconfig file it uses the installed library. pkgconfig only looks in /usr/lib/pkgconfig/. To add /usr/local/… you must reconfigure your builddir:
```
cd builddir
meson setup --reconfigure -Dpkg_config_path=/usr/local/lib64/pkgconfig/ # lib64 for Fedora and lib for Ubuntu
meson compile
```

The include statements changed because the include headers moved to a subdirectory.


mikedickey:
The description is a little scary but the changes look quite tame. Seems fine to me, but others would probably be better to review.

ntonnaett:
I already talked with Chris about this. We should definitely merge all the great stuff that's already in the pipe, have a new release and do the splitting afterwards.

josmithiii:
Sorry I missed the email where I was requested to review this (need to find a way to route it better in gmail), and thanks to Bonnie for pinging me this morning.  Looking now, the first thing I notice is that
```
cd src
build
```
fails with errors. I would prefer to make it work, but if that's too hard, then I think we should delete the build script or replace it with a trivial script following the new build instructions.

Also,
```
qmake jacktrip.pro
```
fails with similar errors (cannot find expected source files).
Finally, the same sort of failure occurs with
```
sh makeXcodeproj.sh
```
This one is super important for me because I use Xcode for all of my JackTrip development.  I _have_ to get that working (which means getting qmake to work again).  What IDE do you use?  I expect QtCreator is broken also, though I didn't test.

All that said, I agree with the goal of having separate library and client!

josmithiii:
Perhaps the problems I listed will disappear after the conflicts are resolved?

ntonnaett:
You can use meson's xcode backend: `meson setup builddir --backend xcode`

On the other side this PR is outdated now. I will write a jacktrip library from scratch without qt dependency. So I will close this for now.

josmithiii:
Great to learn that meson supports Xcode as a back end!

PR #164, opened 2020-09-27
Hard-clip outgoing conversion from floats to fixed-point

josmithiii:
This is not needed when there is a limiter, but since limiters are off by default (I think an outgoing limiter should be ON by default in every audio channel unless -b 32 is in use), it makes sense to at least hard-clip instead of wrap around in two's complement.  Wrap-around is much noisier than clipping.

There was another pull request about this, which I approved subject to comments, but it never got completed, so I wrote this based on that discussion.

### DIFF
--- a/src/AudioInterface.cpp
+++ b/src/AudioInterface.cpp
@@ -507,11 +507,8 @@ void AudioInterface::fromSampleToBitConversion
  const AudioInterface::audioBitResolutionT targetBitResolution)
 {
     int8_t tmp_8;
-    uint8_t tmp_u8; // unsigned to quantize the remainder in 24bits
     int16_t tmp_16;
     double tmp_sample;
-    sample_t tmp_sample16;
-    sample_t tmp_sample8;
     switch (targetBitResolution)
     {
     case BIT8 :
--- a/src/AudioInterface.cpp
+++ b/src/AudioInterface.cpp
@@ -488,16 +488,6 @@ void AudioInterface::computeProcessToNetwork(QVarLengthArray<sample_t*>& in_buff
     mJackTrip->sendNetworkPacket( mInputPacket );
 } // /computeProcessToNetwork
 
-bool is_big_endian(void)
-{
-    union {
-        uint32_t i;
-        char c[4];
-    } bint = {0x01020304};
-
-    return bint.c[0] == 1;  // 1 if bigEndine, 4 if littleEndian
-}
-
 //*******************************************************************************
 // This function quantize from 32 bit to a lower bit resolution
 // 24 bit is not working yet
@@ -507,8 +497,11 @@ void AudioInterface::fromSampleToBitConversion
  const AudioInterface::audioBitResolutionT targetBitResolution)
 {
     int8_t tmp_8;
+    uint8_t tmp_u8; // unsigned to quantize the remainder in 24bits
     int16_t tmp_16;
     double tmp_sample;
+    sample_t tmp_sample16;
+    sample_t tmp_sample8;
     switch (targetBitResolution)
     {
     case BIT8 :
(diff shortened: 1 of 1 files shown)


src/AudioInterface.cpp
```diff
@@ -497,44 +507,39 @@ void AudioInterface::fromSampleToBitConversion
...
+    case BIT24 : {
+        // tmp_sample = (*input) * 32768.0; // 2^15 = 32768.0
+        tmp_sample = std::max(-8388607.0, std::min(8388607.0, std::round( (*input) * 8388607.0 ))); // 2^23 = 8388608
+        union {
+          uint32_t tmp32;
+          char tmp32c[4];
+        } tmp32u;
+        tmp32u.tmp32 = tmp_sample;
+        int offset = ( is_big_endian() ? 1 : 0 );
+        std::memcpy(output, &(tmp32u.tmp32c[offset]), 3); // 24bits = 3 bytes
```
madwort, line 536: The 24-bit implementation looks pretty weird to me - your new 24-bit implementation looks like it's incompatible with the previous version - is this breaking change intentional? it looks like it would mean that older clients are incompatible with newer clients when using 24-bit mode.

madwort: I susppect 24-bit mode is un-used in practice, but I think it's better to explicit about these things.

src/AudioInterface.cpp
```diff
@@ -488,6 +488,16 @@ void AudioInterface::computeProcessToNetwork(QVarLengthArray<sample_t*>& in_buff
...
 } // /computeProcessToNetwork
 
+bool is_big_endian(void)
+{
+    union {
+        uint32_t i;
+        char c[4];
+    } bint = {0x01020304};
+
+    return bint.c[0] == 1;  // 1 if bigEndine, 4 if littleEndian
```
madwort, line 498: typo `Endine`

src/AudioInterface.cpp
madwort: and presumably you should also be updating this block https://github.com/jacktrip/jacktrip/blob/main/src/AudioInterface.cpp#L516-L526 in the same PR?

josmithiii: The previous 24-bit implementation is [documented as not working](https://github.com/jacktrip/jacktrip/blob/9f596d3778b6614650d9253e201ceeb67abed938/src/AudioInterface.cpp#L442).  I therefore did not test in detail how my implementation would behave differently in all 24-bit cases, but rather wrote what I thought was a good approach for handling the conversion in an endian-independent way, intending to put out the same stream of byes in the little-endian non-clipped case.   (I did single-step it in my testing to see that it was doing what I wanted.)  As a result, I expected my 24-bit case to behave differently on big-endian machines, but the same across big/little endian platforms, which I felt was more important.   I didn't worry about breaking cases involving a new or old big-endian or little-endian clients or servers in 24-bit mode, e.g., because the -b 24 feature is documented as not working, and I've never encountered a server in that mode.  While I consider what I wrote to be a move in the right direction, I agree it is not a complete solution for the 24-bit case, since I didn't rewrite the other direction as well.

Since I "touched" the 24-bit case, it is reasonable to ask me to fix it in both directions in this PR, but I think that should happen in a separate PR under current recommendations.  I therefore conclude that I should back out of my recommended change to the 24-bit case in this PR so that both directions can be addressed in a PR aimed at fixing 24-bit support (which, as you note, nobody seems to use).

src/AudioInterface.cpp
josmithiii: Fixed in jos branch

madwort: 👍 thanks

src/AudioInterface.cpp
madwort: Yes, I agree the 24-bit support should be addressed separately - my PR got delayed because I didn't have time to get to the bottom of the 24-bit handling & it would be really good to get a 16-bit fix merged and in the hands of users. 

To be honest, I would suggest as a next step removing / disabling 24-bit support in JackTrip, pending a re-write.

src/AudioInterface.cpp
```diff
@@ -499,22 +499,23 @@ void AudioInterface::fromSampleToBitConversion
...
         // 8bit integer between -128 to 127
-        tmp_sample = floor( (*input) * 128.0 ); // 2^7 = 128.0
+        tmp_sample = std::max(-127.0, std::min(127.0, std::round( (*input) * 127.0 ))); // 2^7 = 128
         tmp_8 = static_cast<int8_t>(tmp_sample);
         std::memcpy(output, &tmp_8, 1); // 8bits = 1 bytes
         break;
     case BIT16 :
         // 16bit integer between -32768 to 32767
-        tmp_sample = floor( (*input) * 32768.0 ); // 2^15 = 32768.0
+        // tmp_sample = floor( (*input) * 32768.0 ); // 2^15 = 32768.0
```
madwort, line 515: I think probably better to remove this line, not sure it adds anything

src/AudioInterface.cpp
```diff
@@ -534,7 +535,10 @@ void AudioInterface::fromSampleToBitConversion
         std::memcpy(output+2, &tmp_u8, 1); // 8bits = 1 bytes
         break;
     case BIT32 :
-        std::memcpy(output, input, 4); // 32bit = 4 bytes
+        tmp_sample = *input;
+        // not necessary yet:
+        // tmp_sample = std::max(-1.0f, std::min(1.0f, tmp_sample));
```
madwort, line 540: if uncommented, this line would throw a compiler error, because it's comparing `float`s with a `double` - could you either fix or remove? the problems with committing code that is commented out! I think this might have worked if `tmp_sample` was a `sample_t`, but you've changed it to a `double`.

src/AudioInterface.cpp
josmithiii: Agreed - I often leave prior versions in a comment for reference in case corner cases might arise

src/AudioInterface.cpp
josmithiii: changed 1.0f to 1.0

wholenotes: Am I reading this thread correctly?  Looks like this line of commented code is the only point of discussion at this stage.

ntonnaett: I would prefer to revert to the onliner for 32bit but as godbolt told me that it's actually the same with -O2 and this already took way too much time, we should push this now.
[Godbolt Comparison](https://godbolt.org/z/nEhhv3)

ntonnaett (approved):

PR #172, opened 2020-10-17
add audio packet latency histogram to --examine-audio-delay (-x) when print interval is 1 sec or more

josmithiii:
In addition to measuring the mean audio packet latency and standard deviation, the --examine-audio-delay (-x) option (for clients only) prints out an ASCII histogram of packet latencies when the printout interval is 1 second or more.  Ideally only one cell will be seen in the histogram, corresponding to a fixed network round-trip time plus audio buffering latency (to within 5 ms).  merge --squash recommended as usual for these little feature PRs.


mikedickey (approved):

josmithiii:
Feel free to suggest any improvements and I'll put them in a new PR.

PR #174, opened 2020-10-20
more general argument settings for Faust effects

josmithiii:
Support more general parameter-setting options for Faust effects modules, particularly the Compressor (requested by Russ of the Stanford Marching Band, since the vocal setting was not working well for brasses).  After this, the jos branch is almost merged into dev.  This PR should only affect Faust effects modules.


faust-src/Makefile
```diff
@@ -2,8 +2,9 @@ h headers :
 	install -d headersdir
 	$(MAKE) DEST='headersdir/' ARCH='faust2header.cpp' -f Makefile.headers
 
+# install sets the "executable" bit, so we use cp instead:
 ih hi installheaders : headers
-	install headersdir/*.h ../src/
+	/bin/cp -p headersdir/*.h ../src/
```
mikedickey, line 7: FYI install has a "-m" parameter to set the specific mode

src/CompressorPresets.h
```diff
@@ -0,0 +1,81 @@
...
+  ~CompressorPreset() = default;
+};
+
+namespace CompressorPresets
+{
+  //                     name   ratio  thresh  attack  rel  mugain
+  const CompressorPreset voice { 2.0f, -24.0f, 15.0f, 40.0f, 2.0f };
+  const CompressorPreset horns { 3.0f, -10.0f, 100.0f, 250.0f, 2.0f };
+  const CompressorPreset snare { 5.0f, -4.0f, 5.0f, 150.0f, 3.0f };
+  const uint numPresets { 3 };
```
mikedickey, line 27: what about all the others? ;-)

mikedickey (approved):

faust-src/Makefile
josmithiii: Thanks -  I switched to that and realized the headers should be installed as readonly since they are derived source.

src/CompressorPresets.h
josmithiii: I was thinking to be conservative until I have someone who can test and vouch for each particular preset choice, since it will not be good form to change a standard preset later after it is in a release.  These presets were obtained from a quick Internet search, and I don't know how good they are.  The two new ones span a large range of compression, and hopefully cover Russ's needs for the Band.

PR #176, opened 2020-10-22
improve clarity of options-related printout

josmithiii:
Print options-related error message after usage so that it can be seen on the same screen.
Minor additional clarity improvements.


josmithiii:
This may be the last PR from the jos branch.  There are more diffs, but they can await further development and testing.

mikedickey (approved):

PR #181, opened 2020-11-01
optionally print error message in optional outgoing limiter when approaching clipping and advise on level reduction, with both options OFF by default

josmithiii:
This small PR was suggested by Sarah Weaver's experience with musicians having multichannel audio interfaces that were being mixed down to mono in JACK by connecting multiple "system capture" channels to one JackTrip input channel. In this situation, the audio interface output does not clip, and local monitoring looks good and sounds fine, but the JackTrip channel often overflows because the input channels were summed without dividing by the number of channels. In dev we now hard-clip instead of wrapping around as in main, which helps, and offer a limiter, but this PR adds sparse notification of the clipping in stderr and says by how much it was in each block so it is then known how far input levels need to be reduced.

Minor additional fixes: -O default in command-line help, and filename typo in jacktrip.pro

### DIFF
--- a/src/Limiter.h
+++ b/src/Limiter.h
@@ -32,9 +32,9 @@
 /**
  * \file Limiter.h
  * \author Julius Smith, based on LoopBack.h
- * \date May 2020
+ * \date May-Nov 2020
+ * \license MIT
  */
-
 
 /** \brief Applies limiter_lad_mono from the faustlibraries distribution, compressors.lib
  *
@@ -51,6 +51,7 @@
 #include "ProcessPlugin.h"
 #include "limiterdsp.h"
 #include <vector>
+#include "assert.h"
 
 /** \brief The Limiter class confines the output dynamic range to a
  *  "dynamic range lane" determined by the assumed number of clients.
@@ -61,6 +62,7 @@ public:
   /// \brief The class constructor sets the number of channels to limit
   Limiter(int numchans, int numclients, bool verboseFlag = false) // xtor
     : mNumChannels(numchans), mNumClients(numclients)
+    , warningAmp(0.0), warnCount(0), peakMagnitude(0.0), nextWarning(1)
   {
     setVerbose(verboseFlag);
     for ( int i = 0; i < mNumChannels; i++ ) {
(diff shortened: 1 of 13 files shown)


josmithiii:
Please hold off on this. Today at jam time I heard some behavior I don't understand (loud clipping with no limiter that should have been much less abrasive due to the new hard-clipping).  I don't think the issue has to do with clip-counting, but resolving it will likely require revising this as well.

josmithiii:
I plan to revise this PR to summarize exceeding the limiter threshold, when the limiter is on, and hard-clipping when it is off, but I no longer plan to investigate the surprisingly loud hard-clipping. It does not happen when the limiter is on, and users of the jacktrip command line are now expected to know what they are doing and either manage their levels below clipping or use the limiter.

josmithiii:
Ok, I changed it to print warnings after every 10,000 samples at level -6dB or greater.
When the outgoing limiter happens to be on, this is when limiting kicks in.
Otherwise it's just good practice to leave a bit of headroom.
If anyone doesn't like this, I can make a new parameter that turns on the warnings, and explicitly set the warning level.

cchafe:
Thanks. Super great to have this as a selectable feature (because of overhead and also for headless it should be off by default). Is it for all modes (-s, -c, -S, -C)?

josmithiii:
This implementation is in AudioInterface::fromSampleToBitConversion which, according to my understanding, catches all conversions from floating-point to 16-bit fixed-point for network transmission.   The printout is always far less than the familiar "UDP waiting too long", which is up to every 30 ms, but maybe you plan to be suppressing that as well?

I don't know if Sarah would benefit from overflow warnings on the server, as many seem to be headless these days, so that leaves my own use case:  For me it would suffice to move this warning into the limiter, since I always use one.  There it could be an optional sparse warning (as implemented here) about having crossed the limiter's compression threshold (typically inaudible peak-bending, but distortion nevertheless).  I think that's the zone where studio level meters go from green to yellow, i.e., "entering soft-limiting compression".  It would be easy to extend "-Oo" to "-Oow" to turn on the warnings for outgoing limiters only.  No server mode would then have this warning feature, but that's fine when we assume the server manager runs through a nice JACK rack, or SuperCollider, etc., which takes care of all such things.

josmithiii:
Ok done!  The feature has been moved into the outgoing limiter from the client only, which is where it is most needed, and it is turned off by default, as is the limiter itself.  The implementation in AudioInterface.cpp is ifdef'd out with ENABLE_DISTORTION_WARNINGS, and if you don't think anyone will ever want that (e.g., to monitor clipping on a server in the console), it is easy to delete.

src/Limiter.h
```diff
@@ -113,6 +115,51 @@ class Limiter : public ProcessPlugin
...
+            fprintf(stderr, "\tReduce audio input level(s) by %0.1f dB to avoid limiter compression distortion.\n",
+                    peakMagnitudeDB-warningAmpDB);
+          }
+          peakMagnitude = 0.0; // reset for next group measurement
+          if (nextWarning < maxWarningInterval) { // don't let it stop reporting for too long
+            nextWarning *= 10;
+          } else {
+            warnCount=0;
+          }
+        } // warnCount==nextWarning
```
mikedickey, line 158: This code looks familiar.. is there a way we could reuse it versus having two copies?

mikedickey (changes requested):

src/Limiter.h
josmithiii: There is actually only one copy, because the copy in AudioInterface.cpp is #ifdef'd out.  I am expecting Chris to choose to delete the more general warning in AudioInterface.cpp, in which case the apparent 2nd copy will disappear.  On the other hand, if we do want such a feature, then we can make a shared utility.  However, I see no demand for that.

josmithiii: To simplify looking at this, I removed the disabled code blocks, but kept them in the jos branch, both to facilitate reviving them if anyone wants them, and to give me a double-check that the warnings in AudioInterface.cpp never fire regardless of which limiter I'm using.

josmithiii:
> I heard some behavior I don't understand (loud clipping with no limiter that should have been much less abrasive due to the new hard-clipping).

I investigated this a bit further (unrelated to the present PR, as it currently stands, but important nevertheless, and discovered in this context), and my findings are summarized by the following comments in `AudioInterface.cpp` in the jos branch:
```
case BIT16 : {
        // 16bit integer between -32768 to 32767
        // original scaling: tmp_sample = floor( (*input) * 32768.0 ); // 2^15 = 32768.0
        const double maxAmp16 = 32767.0; // 2^15 = 32768
        double scaledSample = std::round( (*input) * maxAmp16 );
        tmp_sample = std::max(-maxAmp16, std::min(maxAmp16, scaledSample)); // FIXME: THIS SOUNDS WORSE THAN IT SHOULD - WHY?
        // TEST HACK: tmp_sample = std::max(-maxAmp16+1, std::min(maxAmp16-1, scaledSample)); // check for fencepost bug - JUST AS BAD
        // TEST HACK: tmp_sample = std::max(-maxAmp16*0.75, std::min(maxAmp16*0.75, scaledSample)); // MUCH BETTER! WHY???
        // TEST HACK: tmp_sample = std::max(-maxAmp16/2, std::min(maxAmp16/2, scaledSample)); // same quality as 0.75 I think
        // TEST HACK: tmp_sample = scaledSample; // jacktrip original (bad, of course, and I think about the same as maxAmp16[-1])
        // None of this is an issue when an outgoing limiter is in use, so let's just do that. No idea where the downstream gain is,
        // or MSB lossage, etc.
        tmp_16 = static_cast<int16_t>(tmp_sample);
        std::memcpy(output, &tmp_16, 2); // 2 bytes output in Little Endian order (LSB -> smallest address)
        ...
```

josmithiii:
I originally added Chris as a reviewer because the original version of this PR introduced new overflow warnings by default for all users.  Now the proposed changes (aside from the typo fix in jacktrip.pro), only provide a new Limiter "yellow and red light" warning, which is off by default, and of course the Limiter itself is off by default.  For this reason, I no longer think it is necessary to bother Chris with a mandatory review, as he has enough on his plate already.  Of course everyone is invited as a volunteer reviewer.

josmithiii:
I don't know why this one is stuck, but hopefully it's just everyone is busy. Apologies if it's because I am not reacting to the GitHub website properly.

cchafe:
would be great to have this in the about to be "beta" -- is there anything needs testing before merging?

josmithiii:
I don't think so.  It's been ready for a long time, and I've been using it without any problems, and really depend on the level feedback in the console (which must be turned on).  My last two commits are merely cosmetic tweaks.  Also, not only is it tested and working, but it only touches users of the outgoing limiter.

josmithiii:
Also, I addressed Mike's requested changes to avoid duplication of similar code quite a while ago.

josmithiii:
Trying to remove all required reviewers so that any collaborator can approve.

josmithiii:
Closing this one and resubmitting.

PR #183, opened 2020-11-02
Use outgoing limiter at client by default

josmithiii:
I feel very strongly that we should have an outgoing limiter on by default (from the client).  While we recently added hard-clipping (which is better than wrap-around), it didn't help as much as I expected when there is no limiter (I plan to investigate this further - it sounds to me like the hard-clipping might not be working as expected).  I have another PR (presently suspended) that warns about and summarizes hard-clipping in the console, but if this PR is accepted, I would like to change that PR to propose warning instead about exceeding -6 dB, which is where the limiter begins compressing toward clipping.  The limiter-compression threshold can of course be set to anything; I think -3dB is the most aggressive I've seen, but I prefer the more graceful transition to clipping starting at -6dB - we can of course make this a user-settable parameter if there is any disagreement about the value.  So, this PR needs to be decided before the suspended one can be resumed.

### DIFF
--- a/src/Effects.h
+++ b/src/Effects.h
@@ -90,12 +90,10 @@ private:
 
 public:
 
-  Effects() :
+  Effects(bool outGoingLimiterOn=true) :
     mNumIncomingChans(2),
     mNumOutgoingChans(2),
-    // JOS recommends: mLimit(LIMITER_OUTGOING),
-    // Paranoid choice:
-    mLimit(LIMITER_NONE),
+    mLimit(outGoingLimiterOn ? LIMITER_OUTGOING : LIMITER_NONE),
     mNumClientsAssumed(2)
   {}
 
--- a/src/Settings.cpp
+++ b/src/Settings.cpp
@@ -97,6 +97,7 @@ Settings::Settings() :
     mHubConnectionMode(JackTrip::SERVERTOCLIENT),
     mConnectDefaultAudioPorts(true),
     mIOStatTimeout(0),
+    mEffects(true), // outgoing limiter ON by default
     mSimulatedLossRate(0.0),
     mSimulatedJitterRate(0.0),
     mSimulatedDelayRel(0.0),


mikedickey (approved):

mikedickey:
I agree with enabling the limiter by default. Clipping is a huge problem otherwise. With VS, we've been telling everyone to always enable this, and I recently enabled it by default for all new devices being registered.

cchafe:
Here's the problem with enabling by default -- it's at cross-purposes with many use cases which are geared to transmitting untouched audio. What goes in is what comes out.

josmithiii:
Thanks for explaining the rationale.  Thoughts/comments/ideas:

(1) Below amplitude 1/2 (or wherever the limiter is set to start compressing), the audio remains "untouched".
     Anyone with the chops to avoid clipping manually could also leave 1 bit of headroom.

(2) Could the verbatim use cases be detected in the options somehow?  If so, the limiter could be off in those cases automatically by default.

(3) It could be said that the default is already NOT audio preserving:
- The client audio interface can send whatever it wants to JACK (always fixed-point as far as I've seen)
- JACK converts it to floating point and sends it to JackTrip (this should be bit-preserving for 16 bit ADCs, but 24?)
- JackTrip converts the floats to ints, and sends the lower 16 bits over the net (by default).
   Any floats equal to 1.0 or greater get wrapped around to negative numbers and information is lost.
- The Hub Server converts 16-bit ints back to float and sends them to JACK at the server.
- A similar series of conversions occurs from JACK at the server back to JACK at the client.
- Two different computers are involved running separately compiled versions of JACK and JackTrip, 
   and potentially using different floating-point number formats, FPU hardware, etc. 
   I don't know of any way to guarantee that the returned fixed-point equals the sent fixed-point.

(4) In Mike's case with VS, it's easy enough to have the limiter checkbox on by default.

(5) For "minimizing global trouble", I would propose that the small number of experts needing verbatim transfer are better equipped to use the "-O n" option ("no limiter"), as opposed to figuring out to add "-O o" ("outgoing limiter") to the command line for the many casual users in need of ear, equipment, and "musical headspace" protection. :-)

(6) It's your call!

josmithiii:
We discussed further and decided to leave the limiter off by default.

PR #185, opened 2020-11-02
revise quantization of -x audio latency histogram bins

josmithiii:
The histogram bins did not quite match the range that was being printed out for them.


mikedickey:
Is this supposed to target dev instead of main? If so, note you can just edit the PR to change target branch.

josmithiii:
Fixed - had to happen eventually!

mikedickey (approved):

PR #187, opened 2020-11-04
fix limiter default claimed in usage printout

josmithiii:
Only affects one line of USAGE printout.


josmithiii:
This PR is obsoleted by PR #181

PR #193, opened 2020-12-04
Hublistener loop fix

psiborg112:
Hopefully fixes an issue where the UdpHubListener gets stuck in an infinite loop while waiting for a previous JackTripWorker instance to finish.

### DIFF
--- a/src/JackTripWorker.cpp
+++ b/src/JackTripWorker.cpp
@@ -65,6 +65,7 @@ JackTripWorker::JackTripWorker(UdpHubListener* udphublistener, int BufferQueueLe
     mUnderRunMode(UnderRunMode),
     mClientName(clientName),
     mSpawning(false),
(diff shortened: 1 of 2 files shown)


josmithiii:
Shouldn't we change gVersion in src/jacktrip_globals.h to 1.3.1 ?

psiborg112:
Probably. But in the meantime, Chris has created a script that reliably triggers the bug and it seems that even with this patch there's still a code path that sees the thread not sending the signal that it's been terminated. So we should probably hold of on this. (I have a plan to simplify the way we're handling the spawning of jacktrip instances that should make it easier to debug these concurrency issues, but it'll be a few days off yet.)

ntonnaett:
There wasn't a 1.3 release yet. So it should be 1.3rc2 I assume.

PR #199, opened 2020-12-27
Merging hotfix-1.2.2 branch into dev by way of merge from main

mikedickey:
Had to manually resolve a bunch of conflicts. Needs more review on impact wrt new async listener.

@psiborg112 please review logic of return statements when client udp port fails check

### DIFF
--- a/src/UdpHubListener.cpp
+++ b/src/UdpHubListener.cpp
@@ -182,26 +182,13 @@ void UdpHubListener::receivedClientInfo(QTcpSocket *clientConnection)
     }
     uint16_t peer_udp_port= readClientUdpPort(clientConnection, clientName);
 
-    //TODO: Check if this is the best way to handle this now that we've refactored this code.
-    if ( peer_udp_port == 0 ) {
-        cout << "JackTrip HUB SERVER: Client UDP Port is = " << peer_udp_port << endl;
-        cout << "JackTrip HUB SERVER: Exiting " << endl;
-        return;
-    }
-
-    if ( peer_udp_port < gBindPortLow ) {
-        cout << "JackTrip HUB SERVER: Client UDP Port is = " << peer_udp_port << endl;
-        cout << "JackTrip HUB SERVER: Exiting " << endl;
-        return;
-    }
-
-    if ( peer_udp_port > gBindPortHigh ) {
-        cout << "JackTrip HUB SERVER: Client UDP Port is = " << peer_udp_port << endl;
-        cout << "JackTrip HUB SERVER: Exiting " << endl;
-        return;
-    }
-
     cout << "JackTrip HUB SERVER: Client UDP Port is = " << peer_udp_port << endl;
+    if ( peer_udp_port == 0 || peer_udp_port < gBindPortLow || peer_udp_port > gBindPortHigh ) {
+        cout << "JackTrip HUB SERVER: Exiting " << endl;
+        clientConnection->close();
+        clientConnection->deleteLater();
+        return;
+    }
     
     // Check is client is new or not
     // -----------------------------
--- a/src/UdpHubListener.cpp
+++ b/src/UdpHubListener.cpp
@@ -176,13 +176,31 @@ void UdpHubListener::receivedClientInfo(QTcpSocket *clientConnection)
     // ------------------------
     QString clientName = QString();
     cout << "JackTrip HUB SERVER: Reading UDP port from Client..." << endl;
-    if (clientConnection->bytesAvailable() < (int)sizeof(uint16_t)) {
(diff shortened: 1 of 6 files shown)


src/UdpHubListener.cpp
```diff
@@ -176,13 +176,31 @@ void UdpHubListener::receivedClientInfo(QTcpSocket *clientConnection)
...
+        cout << "JackTrip HUB SERVER: Exiting " << endl;
+        return;
+    }
+
+    if ( peer_udp_port > gBindPortHigh ) {
+        cout << "JackTrip HUB SERVER: Client UDP Port is = " << peer_udp_port << endl;
+        cout << "JackTrip HUB SERVER: Exiting " << endl;
+        return;
+    }
+
```
psiborg112, line 203: This will leave the TCP connection open and allow for further packets to be sent from the client. I'm assuming that what we really want to do when receiving an invalid port number is to close the connection. You'll need something like this before each return statement:
```
clientConnection->close();
clientConnection->deleteLater();
```

src/jacktrip_globals.h
```diff
@@ -99,6 +99,8 @@ const int gWaitCounter = 60;
 /// \name Network related ports
 //@{
 const int gDefaultPort = 4464; ///< Default JackTrip Port
+const int gBindPortLow = 3464; ///< lowest Bindport
+const int gBindPortHigh = 5464; ///< highest Bindport
```
psiborg112, line 103: Do we need to restrict the port range this heavily? Wouldn't it work to just check that we have a valid, non-zero, unsigned 16 bit integer? (We could also prevent the use of well known ports below 1024 if we need to.)

src/UdpHubListener.cpp
psiborg112: You could simplify this considerably as:
```
cout << "JackTrip HUB SERVER: Client UDP Port is = " << peer_udp_port << endl;
if ( peer_udp_port < gBindPortLow || peer_udp_port > gBindPortHigh ) {
    cout << "JackTrip HUB SERVER: Exiting " << endl;
    clientConnection->close();
    clientConnection->deleteLater();
    return;
}
```

src/jacktrip_globals.h
mikedickey: I recall signed wrap-around being a problem, not sure if there are others @cchafe ?

BTW, I'm curious why the client sends the port it's bound to in the protocol, versus just using the port from the UDP header? Seems likely this is what causes NAT issues with JackTrip. Namely, every device behind a NAT needs to be configured to listen on a different port. Also it seems like we are just getting lucky that NAT's aren't remapping this to another random port when they forward packets?

cchafe: There were some uninitialized variables in the same neighborhood. Fixed those that I saw and made the types consistent. The port range was chosen arbitrarily. Essentially, it's doing the job a firewall would do. Historically, TCP / UDP were the same but became independent to support a wider range of ephemeral ports + the ability to have multiple clients behind the same NAT. We just had a situation today where excluding all but a single TCP port was advantageous for "securing" a session (definitely a hack). A future feature would be setting that from cmd line,

psiborg112: The only problem I see with arbitrarily limiting the port range is that we're not mirroring that restriction in Settings.cpp, which could cause a lot of confusion when people try using ports outside of that narrowly defined range on the client side.

On the NAT front, yes, we probably should be overriding the port that's sent by the client with the port that's taken from the initial UDP packet, and it's a trivial, three line, fix. (At which point, the restriction on the sent port becomes kind of moot except as a very basic security device.) I've just implemented it in QJackTrip, but the UDP hub code in that has diverged quite a bit from the current dev branch.

mikedickey:
What happens if the client connects to hub server via TCP, the server thinks it's valid because it gets a packet that "just happens" to be within the expected client bind port range, and then the client disappears?

Just wondering more generally if restricting the port range may just happen to be making a bigger problem less likely to trigger, but still possible when it happens to hit to port lottery?

src/jacktrip_globals.h
psiborg112: Scratch that. It's not a trivial fix. On a networking level we're already using the port in the UDP header rather than the client supplied one. We're using the port sent by the client to identify whether a client is unique or not, and to replace the existing connection if we already have a record of it. There's a less trivial fix for it (having JackTripWorker notify UdpHubListener of the actual port being used and updating the record), but it's going to be a bit of a concurrency headache when multiple machines from behind the same NAT initiate a connection on the same port at the same time.

cchafe:
In that case, it'll reach line 217 and clean up correctly.

cchafe (approved):
Because UdpHubListener in main uses a threadloop structure and in dev it doesn't a bit of testing on this before merging to beta.

PR #209, opened 2021-01-20
exclude gitconfigs from archive and formatting fixes

umlaeute:
## exclude repository configuration
this PR excludes repository configuration (like `.gitignore` and `.travis.yml` from the archive files (generated via `git archive`, or when downloading a zip/tar release (resp. snapshot) from github).

reasoning: those file really only make sense in the context of this gitrepository.

speaking with my Debian maintainer hat on:
however, if some downstream consumer of the source packages (e.g. the a distribution package) uses `git` to manage the packaging, they usually have different requirements on how the repository should be configured.

## formatting fixes
- this PR also fixes the line-endings of the INSTALL.txt file to only use LF, like the rest of the repository (git usually does a nice job of automatically converting text-files to the correct line-ending, depending on where the repository is cloned to)
- break overlong lines in INSTALL.txt to be pleasant to look on even on a traditional 80chars terminal
- use <code>`</code> to mark code-examples, to make them more easily distinguishable from the running text.

you might consider converting `INSTALL.txt` to markdown (but this is often a controversial field, so I didn't attempt to do it)


ntonnaett:
What was wrong with the line-endings?

I like the markdown change. Could you add a 'git mv INSTALL.txt INSTALL.md'?

umlaeute:
> What was wrong with the line-endings?

all files have native line-ending (`LF` on my linux system), but for a single exception.
INSTALL.txt had `CRLF` ending.

> I like the markdown change. Could you add a 'git mv INSTALL.txt INSTALL.md'?

done.

ntonnaett:
Thank you! Could you rebase to dev?

I'm not sure about the git archive exclude. Is it part of the Debian packaging rules? If I download the packaged version off Github I usually don't want those files. But on the other side it's not an archive of the repository anymore then.

umlaeute:
> I'm not sure about the git archive exclude. Is it part of the Debian packaging rules? If I download the packaged version off Github I usually don't want those files. 

yes, that's the point.

> But on the other side it's not an archive of the repository anymore then.

i don't get the problem here.
why do you want it in the archive?
the *archive* (as generated by `git archive`) is a zipfile/tarball that **does not** contain the repository (the `.git/` directory) - it's a snapshot.

so:
- why would you want to have a `.gitignore` file in this directory that is not under version control?
- if you put that directory under version-control *again*, why do you want to impose the configuration of the original repository onto that brand-new (and independent) repository?
- why do you want the configuration for `github-actions` in the tarball? github-actions *only* make sense if they are in a github repository. if you are working with the `git archive`, then you have obviously left github, so the actions are meaningless

basically it boils down to:
- if you want to work close to the upstream repository (*this* repository), you should just do a `git clone` (and all the files will be there)
- if you want to work agnostic of the upstream repository, get the `git archive` (and it shouldn't interfere with your own archive).

> Is it part of the Debian packaging rules?

Debian packaging rules expose the problem, but I don't think they are the root of it.

umlaeute:
ach ja: i rebased and force-pushed :-)

ntonnaett:
I've never used git archive. So thank you for the clarification.
After the one thing I've found in INSTALL.md, I'll merge this. The Github Actions Linux build failed because of the missing apt update. That's fixed now. So next build will pass.

ntonnaett:
It seems there were some merge hiccups. Would you remove the last merge commit? I don't understand where this conflict comes from. I'm happy to merge then.

umlaeute:
hrmpf. force pushed again.

ntonnaett (approved):

PR #212, opened 2021-01-21
GitHub actions: update apt before installing packages

dyfer:
It seems I forgot to update Linux repositories, thus Linux build failed: https://github.com/jacktrip/jacktrip/runs/1745102411?check_suite_focus=true#step:3:194

EDIT: the Linux build in this PR passed


ntonnaett (approved):
That was fast!

PR #213, opened 2021-01-23
Add clang-format coding style

ntonnaett:
Everybody loves beautiful arranged code. But nobody wants to lose time indenting and remembering what the right style was in this project. So here is a .clang-format file. I tried to get a similar style that we have right now, but enhances readability.

A lot of IDE and texteditor are able to use .clang-format files. So most of us just have to tell our tools to use it. I used meson. If you setup the builddir with a .clang-format file in the root directory of the project, you get a clang-format target that you can start with `ninja clang-format`.

Although clang-format is able to reflow comments and most of the time it's beneficial, sometimes it gets a mess. So I turned it off for now. For two places in faust-src I turned clang-format off because I thought those are a kind of macro.

It broke the build first, because it reordered the #includes (so you find headers that are not self-contained). I had to change an uint into an unsigned int because of that. We might even avoid bugs with this in the future.

Are there some style decisions you don't like?

### DIFF
--- a/.clang-format
+++ b/.clang-format
@@ -1,0 +1,182 @@
+---
+Language:        Cpp
+# BasedOnStyle:  Google
+AccessModifierOffset: -1
+AlignAfterOpenBracket: Align
+AlignConsecutiveMacros: true
+AlignConsecutiveAssignments: true
+AlignConsecutiveBitFields: false
+AlignConsecutiveDeclarations: false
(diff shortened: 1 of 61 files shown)


josmithiii:
Looks better to me

josmithiii (approved):

ntonnaett:
If we trigger clang-format for every PR, there will be no tabs in the future.

PR #216, opened 2021-01-26
Clang tidy

ntonnaett:
Clang-tidy is a static analysis tool. There's a .clang-tidy file that lists and configures all checks that are turned on. Most of those checks are just about rephrasing. But even this can improve performance and readability.

There could be some breaking changes. Especially those changes to the signedness of some variables (e.g. Jitterbuffer) could be a problem. But those changes were made because the signedness clashed with each other before.


src/jacktrip_globals.cpp
```diff
@@ -146,7 +146,6 @@ void setRealtimeProcessPriority()
 #ifdef __UBUNTU__
     priority = 95;  // anything higher is silently ignored by Ubuntu 18.04
 #endif
-    priority = 3;
```
ntonnaett, line 149: We should investigate this.

ntonnaett: For every Linux priority was set to 3. I don't think that's what you expect if you use this flag. So I just set it to 50 now. It's below jack and the audio threads. So that shouldn't interfere.

antonrunov: I agree that the code was not clear at all (it was a mix of layers from different times). But setting the priority to 3 is the expected behavior. This priority should be lower than the priority of jack callbacks. Ideally I think it should be calculated directly from the jack callback priority which can be obtaind with jack API.

One reason why all this priority handling code is quite messy is that there were discussions and different attempts to improve jacktrip performance but this work hasn't been finished. There are still open questions and probably some disagreement.

ntonnaett: I pushed a commit that sets the udp rt priority 10 lower than the jack audio clients' priority. It's only tested with Linux.

antonrunov: Well, on my Debian at least, jack sets the callback RT priority to 5 by default. That will result in a negative value for the priority although the valid range is 1..99 (if I remember correctly). That's where 3 came from btw. I think `jack_callback_priority - 1` should be fine. We don't whant to block jack callbacks and even one step lower is ok. Even more accurate way would be to check `sched_get_priority_min()` and use it as a low limit.

ntonnaett: Interesting. I was already surprised that the default for the jack server is 20 on Fedora.

ntonnaett: So it's -1 now.

ntonnaett: Is this okay or should I add a lower boundary too?

cchafe: If the question is about that UBUNTU line, please take it out. I put that there in 2018 when we still used a scary maximum.

ntonnaett: Yes, I removed that and modified the code so it sets the udprt priority 1 lower than the audio threads. But maybe we also need to set a lower boundary for the case that the audio thread priority was set to 0… A possible but rare case.
I think I will do it, because safety first.

ntonnaett: Done

ntonnaett:
I have to run clazy over this before merging.

ntonnaett:
This will be a place of cherry picking. Some changes were already done a second time in other branches. Some were automatic.

cchafe:
I've been building with qmake -- please remind me the steps to test this

ntonnaett:
This branch/PR is outdated. I split two PR out of it already #316 and #273. If you look in the upcoming documentation #298 you will find a page for static analysis now.

PR #227, opened 2021-02-14
Fix regression in remote client name handling

ntonnaett:
Now that the port is only being read as a 16 bit integer by the hub server, the requested remote client name isn't being read properly. We need to discard the two bytes that come before it.


ntonnaett:
This is already in main and got already approved twice. I couldn't merge main directly to dev because of conflicts, so I created a separate branch for this.

cchafe (approved):

PR #230, opened 2021-02-17
Don't trigger Github Action build on every push

ntonnaett:
but only on PRs and manually. Right now it often gets triggered twice.

And remove Travis.


cchafe (approved):

dyfer:
Are we sure we don't want GHA to run on every push?

ntonnaett:
I think if you really need a CI build before a PR you can trigger it manually. And if you have an open PR and push it gets triggered automatically.

dyfer:
I'd argue that build on every push is useful, so you can see if anything is broken, especially for a relatively small project like JT.
I actually couldn't manually trigger the workflow in my fork, not sure why that is... I temporarily changed it to run on push, so that I can test things.

ntonnaett:
Did you add workflow_dispatch? Don't you build on your local machine?

If we add it again, I think we should enable caching first and don't trigger twice on one push.

I'm really interested in your workflow now.

dyfer:
Re: workflow - I'm working on my [fork](https://github.com/dyfer/jacktrip). I updated dev from `upstream` (i.e. `jacktrip/jacktrip` repo), which had the `workflow_dispatch` entry, as added in this PR. Then I created another [branch](https://github.com/dyfer/jacktrip/tree/topic/fix-windows-gha-build) based off of `dev` and pushed it (to my fork), and that did not trigger the build (obviously, since push trigger is disabled) but I also **don't see an option to trigger builds** in my [Actions](https://github.com/dyfer/jacktrip/actions) tab.
I wonder if this has to do with changing/renaming default branch, my fork didn't seem to have that update and I tried to change this (i.e. set the default branch in my for to `main` manually), though maybe something got messed up in the process...

ntonnaett:
Mmh, I see. Maybe we have to push activate workflow dispatch to main?

dyfer:
> Mmh, I see. Maybe we have to push activate workflow dispatch to main?

I was wondering about that.

ntonnaett:
Maybe you were right. It looks like we need trigger on push even if we push to PRs.

dyfer:
Not sure how PR trigger works - maybe it's only for PRs from forks? I don't think it would make sense...

PR #232, opened 2021-02-17
[WIP] Windows GitHub actions

ntonnaett:


cchafe (commented):
should I approve this so it can be tested, or wait?

ntonnaett:
I thought I need a normal PR for testing but actually a draft is all I need. And if the check doesn't fail anymore you can merge. So for now you can just relax :) Right now there's a problem with the path to the build script. I hope we can just use this build script.
Qt and libjack installed with no error.

PR #233, opened 2021-02-17
RtAudio Support, Meson Wraps, MSVC Support, Skip Duplicate Builds

ntonnaett:
- Meson gets RtAudio now if it's missing and builds it.
- RtAudio builds again (Jack disabled)
  - CoreAudio for MacOS
  - Alsa and Pulseaudio for Linux
  - WASAPI for Windows
- Fixed MSVC build and Github Actions builds with MSVC
  - uses meson wrap file for getopt (is POSIX API)
- Meson has options now: rtaudio, jack and wair
- Meson builds with C++17 as default
- There shouldn't be anything missing for meson in comparison to the jacktrip.pro file
- Skips duplicate Github Action builds
  - Before: 12 builds ((3 old, 3 meson) *2)
  - Now: 6 builds + 4 skip-duplicates (identifier jobs) + 2 skipped jobs

### DIFF
--- a/.github/workflows/jacktrip.yml
+++ b/.github/workflows/jacktrip.yml
@@ -1,7 +1,22 @@
 name: build
 on: [push, pull_request]
 jobs:
+  skip-duplicates:
+    runs-on: ubuntu-latest
+    outputs:
+      should_skip: ${{ steps.skip_check.outputs.should_skip }}
+    steps:
+      - id: skip_check
+        uses: fkirc/skip-duplicate-actions@master
+        with:
+          # All of these options are optional, so you can remove them if you are happy with the defaults
+          concurrent_skipping: 'same_content'
+          skip_after_successful_duplicate: 'true'
+          paths_ignore: '["**/README.md", "**/INSTALL*"]'
+          do_not_skip: '["pull_request", "workflow_dispatch", "schedule"]'
   build:
+    needs: skip-duplicates
+    if: ${{ needs.skip-duplicates.outputs.should_skip != 'true' }}
     runs-on: ${{ matrix.runs-on }}
     name: ${{ matrix.name }}
     strategy:
--- a/.github/workflows/meson.yml
+++ b/.github/workflows/meson.yml
@@ -84,14 +84,16 @@ jobs:
         env:
           DEVELOPER_DIR: '/Applications/Xcode_11.7.app/Contents/Developer' # specify XCode version on macOS
         run: |
-          meson $BUILD_PATH
+          meson --buildtype release $BUILD_PATH
           cd $BUILD_PATH
+          meson configure -Ddefault_library=static
           meson compile
       - name: build on Windows
         if: runner.os == 'Windows'
         run: |
-          meson setup builddir
+          meson --buildtype release builddir
           cd builddir
+          meson configure -Ddefault_library=static
           meson compile
       - name: upload artifacts
         uses: actions/upload-artifact@v2
--- a/.github/workflows/meson.yml
+++ b/.github/workflows/meson.yml
@@ -1,0 +1,101 @@
+name: meson-build
+on: [push, pull_request]
+jobs:
+  skip-duplicates:
+    runs-on: ubuntu-latest
+    outputs:
+      should_skip: ${{ steps.skip_check.outputs.should_skip }}
+    steps:
+      - id: skip_check
+        uses: fkirc/skip-duplicate-actions@master
+        with:
+          # All of these options are optional, so you can remove them if you are happy with the defaults
+          concurrent_skipping: 'same_content'
+          skip_after_successful_duplicate: 'true'
+          paths_ignore: '["**/README.md", "**/INSTALL*"]'
+          do_not_skip: '["pull_request", "workflow_dispatch", "schedule"]'
+  build:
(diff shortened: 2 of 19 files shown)


ntonnaett:
With `meson configure` enabling and disabling of Jack, RtAudio and WAIR is now possible.
When turning off Jack on my Linux machine I get a linking Error:

> /usr/bin/ld: jacktrip.p/src_UdpHubListener.cpp.o: in function `UdpHubListener::receivedClientInfo(QTcpSocket*)':
/home/nils/Development/jacktrip/builddir/../src/UdpHubListener.cpp:280: undefined reference to `UdpHubListener::connectPatch(bool)'
/usr/bin/ld: jacktrip.p/src_UdpHubListener.cpp.o: in function `UdpHubListener::releaseThread(int)':
/home/nils/Development/jacktrip/builddir/../src/UdpHubListener.cpp:476: undefined reference to `UdpHubListener::connectPatch(bool)'

ntonnaett:
I've just made the meson Github Action build for Ubuntu and MacOS. So it builds a Version for Ubuntu with Jack and RtAudio (Alsa, Jack and RtAudio Backends) and for MacOS with Jack and RtAudio (Jack and CoreAudio Backends). RtAudio gets pulled in via a meson wrap file.

ntonnaett:
I've got jacktrip to build with Meson and MSVC on another branch. JackTrip didn't build on MSVC before. I see if I can add it to this PR.

dyfer:
Thanks Nils!
I've tested this on macOS with builds from GHA. RtAudio works! tested in the client (`-c`) mode only. Here are my comments/observations:
- rtaudio is only available in the meson build - not sure if this was intentional
- the only functional issue I see is with the default device selection: RtAudio should select system device if no `deviceid` is provided.
  - if jack is running, then jack server is selected (which I think is good). However, if jack is not running, the _first_ audio device is selected, instead of the default system device. This fails for me, since the first device is microphone (input only). 
  - also, JackTrip name in jack (through rtaudio) is `RtApiJack`. It should probably match JT's native Jack backend and be set to the remote hostname
- finally, this might be for another PR: we need two additions for RtAudio IMO:
  - listing devices with their IDs
  - setting input and output independently (did you mention that his has been implemented?)

ntonnaett:
Thank you for testing!

>     * rtaudio is only available in the meson build - not sure if this was intentional

I won't touch the qmake build system anymore. But I actually assumed that this should just work with qmake because RtAudio support isn't new. It was just broken.

>     * the only functional issue I see is with the default device selection: RtAudio should select system device if no `deviceid` is provided.
>       
>       * if jack is running, then jack server is selected (which I think is good). However, if jack is not running, the _first_ audio device is selected, instead of the default system device. This fails for me, since the first device is microphone (input only).
>       * also, JackTrip name in jack (through rtaudio) is `RtApiJack`. It should probably match JT's native Jack backend and be set to the remote hostname

Selecting the default device means iterating over all devices and check if it is default. And then you can't distinguish between input and out default devices. But having one default device is doable.
The part where we set the client name is a mess right now and needs a lot of rearranging. At least I changed it to set it to the default client name.
 
>     * finally, this might be for another PR: we need two additions for RtAudio IMO:
>       
>       * listing devices with their IDs (another command line switch)
>       * setting input and output independently (did you mention that this has been implemented?)

I think choosing the default device should be part of this PR. Everything else is too much for this. I don't remember if somebody else worked on independent input and output.

dyfer:
> Selecting the default device means iterating over all devices and check if it is default. And then you can't distinguish between input and out default devices. But having one default device is doable.

That seems clunky. In PortAudio you can a) request system default device and b) check for number of inputs/outputs when iterating over devices to distinguish inputs from outputs. Is this not feasible with RtAudio?

> The part where we set the client name is a mess right now and needs a lot of rearranging. At least I changed it to set it to the default client name.

Ah okay. 
If the native Jack backend is still in, I'd just disable Jack support in RtAudio for now so that there's less confusion.

> I think choosing the default device should be part of this PR. Everything else is too much for this. 

Agreed.

ntonnaett:
> That seems clunky. In PortAudio you can a) request system default device and b) check for number of inputs/outputs when iterating over devices to distinguish inputs from outputs. Is this not feasible with RtAudio?
> 
> EDIT: FWIW JackTrip 1.1 did choose default system device properly with RtAudio.

It's the first time I'm really looking into the RtAudio API and it seems to be okay if you do basic things. But maybe directly using CoreAudio would also be an option.
If the device ID is set to 0 in RtAudio, you get the default or first device. You can't tell. Maybe the behavior changed or there were some changes in JackTrip. Thanks for the hint.

> If the native Jack backend is still in, I'd just disable Jack support in RtAudio for now so that there's less confusion.
Yes, that's the best solution.

ntonnaett:
It's not that easy right now to only statically link RtAudio. So I defaulted to statically link everything in GHA. And it is a release build now.

ntonnaett:
It's hard to disable Jack at buildtime for RtAudio. But I specified CoreAudio as backend on MacOS now.

ntonnaett:
As we disabled Jack RtAudio now and are using default input and output devices, do we need to list the device IDs? That would be the only thing missing now.

cchafe (approved):

PR #236, opened 2021-02-20
changed script to record broadcast channels too

wholenotes:


ntonnaett (approved):
I didn't test it but it looks fine to me.

PR #238, opened 2021-02-25
Merge main to dev

dyfer:
This is my attempt to merge main to dev... not sure if this is correct?

### DIFF
--- a/INSTALL.txt
+++ b/INSTALL.txt
@@ -1,0 +1,169 @@
+Jacktrip : Build Instructions
+
+JackTrip: A System for High-Quality Audio Network Performance over the Internet.
+
+The following are instructions for compiling Jacktrip from source.  Compiling from source is the best way to keep up with the latest changes, both stable and experimental.  For quicker ways to install Jacktrip, go to README.md in the root directory of the project.
+
+Steps:
+Dependencies
+Build
+Installation (Linux and OS X only)
+Installation or Build Verification
+---
+Dependencies(for all operating systems):
+g++
+Qt 5.14.2
+Jack
+Note: rtaudio is no longer a dependency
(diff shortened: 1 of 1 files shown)


ntonnaett:
What's the advantage to the PR that has just the hotfix commit?

dyfer:
what I did was:
- update `dev` from `upstream` (this repo)
- create new branch `mp/merge-main-to-dev` based on `dev`
- merge `main` into that branch
- create PR from that branch
I was expecting merge conflicts that's why I created a separate branch for the merge, it if there are none, we should be able to just merge `main` into `dev` directly... maybe?

ntonnaett:
Do we need this if nothing changes?

mikedickey:
Just tried same and got the same result: no changes as a result of the merge. Agree that it may therefore not be necessary? We could just wait on this until after there are changes in main that we want to also migrate to dev.

ntonnaett:
For example merge #237 to main and then main to dev?

dyfer:
Yeah, I think #237 would be a good candidate to test this workflow - a small change, but one that we want to have in both branches (I presume sooner than later).

mikedickey:
If you do another merge from main into this PR branch, it should pick up the changes in the PR

dyfer:
Hm, maybe I messed this up. I removed previous "merge" commit (using rebase) and merged from main again. But it doesn't look right now does it?

dyfer:
Okay, I redid it. I reset this branch to dev, merged main, and force pushed. I think it's correct. What do you think @mikedickey ?

mikedickey:
@dyfer Looks right to me!

mikedickey (approved):

PR #239, opened 2021-02-26
Use Jack 1.9.17 in GHA, update qmake file for Jack 1.9.17 (Windows)

dyfer:
I tried switching jack version on windows to 1.9.17 in GitHub Actions. This also required updating include and lib paths in the `jacktrip.pro` file.

Right now I'm checking for existence of both "new" and "old" Jack locations on windows and using the one that's found. If there is a better way to do this, I'd be happy to learn.

Previously we discussed pkg-config with Nils. I now think we shouldn't rely on it, at least for qmake build, as it's an additional dependency that needs to be installed on windows. Again, I wonder if there's a better way to search multiple library locations in plain qmake config. 

I've also tried [building](https://github.com/dyfer/jacktrip/runs/1989751578) with this change using the [old version](https://github.com/dyfer/jacktrip/runs/1989751578?check_suite_focus=true#step:5:14009) and it seems [fine](https://github.com/dyfer/jacktrip/runs/1989751578?check_suite_focus=true#step:9:22).

I also think this should go into main before the next release.

### DIFF
--- a/.github/workflows/jacktrip.yml
+++ b/.github/workflows/jacktrip.yml
@@ -38,7 +38,7 @@ jobs:
       - name: install dependencies for Windows
         if: runner.os == 'Windows'
         run: |
-          choco install jack
+          choco install jack --version=1.9.17
       - name: Cache Qt
         id: cache-qt
         if: runner.os == 'Windows'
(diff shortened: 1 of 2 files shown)


ntonnaett (approved):

ntonnaett:
This will be in the next release.

PR #240, opened 2021-03-04
now works with JACK 1.9.11 and 1.9.17 on Win 10

wholenotes:


dyfer:
Thanks @wholenotes 
As discussed last week, this has been addressed in #239 and the change is already in `dev`. I propose to close this, is that okay?

ntonnaett:
Do we want to wait for the next release or backport the fix?

cchafe:
Good question. We could also advise to use the latest dev build. Does it now get generated automatically? If so, we'd just need to put it in a good place so folks can get at it.

dyfer:
CI builds are not statically linked so probably not feasible for Windows and macOS users...

ntonnaett:
The meson builds are.

dyfer:
> The meson builds are now.

Including Qt?

ntonnaett:
I'm confused right now. I sent you the wrong link in the mail. It was actually the qmake build. Why is the qmake build statically linked?
The meson build seems to be mostly statically linked. The RtAudio subproject build makes problems.

dyfer:
> I'm confused right now. I sent you the wrong link in the mail. It was actually the qmake build. Why is the qmake build statically linked?

Ah, I see.
I looked at it with `otool` and haven't seen reference to rt-audio so I thought it's statically linked. But instead if just doesn't include rtaudio (which is to be expected for the qmake link right now).

> The meson build seems to be mostly statically linked. The RtAudio subproject build makes problems.

Here are the libraries linked from this build https://github.com/jacktrip/jacktrip/actions/runs/646861131 
```shell
$ otool -L jacktrip
jacktrip:
	@rpath/libcm_rtaudio.dylib (compatibility version 0.0.0, current version 0.0.0)
	/usr/local/opt/qt@5/lib/QtCore.framework/Versions/5/QtCore (compatibility version 5.15.0, current version 5.15.2)
	/usr/local/opt/qt@5/lib/QtNetwork.framework/Versions/5/QtNetwork (compatibility version 5.15.0, current version 5.15.2)
	/usr/local/opt/jack/lib/libjack.0.1.0.dylib (compatibility version 0.1.0, current version 1.9.17)
	/usr/lib/libc++.1.dylib (compatibility version 1.0.0, current version 902.1.0)
	/usr/lib/libSystem.B.dylib (compatibility version 1.0.0, current version 1281.100.1)
```
For Qt, AFAIU, you need to _build_ it as a static lib first, that's why it's so cumbersome.

ntonnaett:
Although I have also the static version of Qt installed it doesn't want to find it. And there's no static version of Jack at all. So, is it feasible at all to link jacktrip statically?

ntonnaett:
And I totally forgot. If we link statically with qt jacktrip becomes at least LGPL.

dyfer:
> So, is it feasible at all to link jacktrip statically?

YES. 
On macOS and Windows we want a single binary that can be just dowloaded and run. Although we could consider installing the required libraries (Qt and maybe RtAudio) instead, though it didn't seem what JT devs were doing so far.

> And there's no static version of Jack at all. 

Jack should not be static - we need Jack to run anyway. And with static jack we would have actual compatibility problems with different jack versions. The way around this would be to use WeakJack, as discussed elsewhere.

> Although I have also the static version of Qt installed it doesn't want to find it.
I don't know how to do this properly - I experienced this with RtAudio from homebrew on macOS. It provides both static (`.a`) and dynamic (`.dylib`) RtAudio library. However, I have no idea how to make `qmake` choose the static one. `-static` doesn't work since there are other libraries that need to be dynamic (namely Qt, which I only have from homebrew, i.e. dynamic). In the end I tried just removing the dynamic rtaudio library files directly from homebrew installation, and that resulted in `qmake` using the static lib... Is there really no way to make static/dynamic distinction in the linker parameters?

Good point on the licensing (I think... my understanding of that is murky at best).

ntonnaett:
I mean we only need to bundle Qt5Core, Qt5Network and RtAudio. That's not much.

dyfer:
> I mean we only need to bundle Qt5Core, Qt5Network and RtAudio. That's not much.

It's not much, but they need to be installed.
Even though we provide installers for macOS and Windows, there's no extra files that need to be installed, the installer is only responsible for putting a single binary in `/bin` or `Program Files` and creating shortcuts if needed. There's more to discuss about handling libraries on those systems, and I only have partial information on that. Let's discuss that separately to keep this discussion on track (i.e. - do we backport the windows library change to main or not?)

dyfer:
We've discussed this at today's meeting with @cchafe and @ntonnaett 
Since the equivalent change is already in `dev`, we'll close this. The `1.4.0` release should come soon enough to address building JT on Windows in `main`.  In the meantime, Windows users who want to build with the current Jack2 should use the `dev` branch.

dyfer (commented):
Thanks @wholenotes ! The equivalent change is already in `dev`.

PR #241, opened 2021-03-12
Meson rtaudio static

ntonnaett:
This should link everything in the meson build statically. It should be configurable in the future.


ntonnaett:
Bad Idea for Windows

PR #255, opened 2021-03-26
Change helpscreen wording for --broadcast argument

michaeldessen:
Suggesting this new wording for --broadcast argument on help screen (changed in Settings.cpp): "Creates duplicate receive ports for each connected client that use the specified broadcast_queue length, to enable creating a broadcast output mix with added latency but less packet loss"

### DIFF
--- a/src/Settings.cpp
+++ b/src/Settings.cpp
@@ -710,7 +710,8 @@ void Settings::printUsage()
          << endl;
     cout << " --bufstrategy # (0, 1, 2)                Use alternative jitter buffer"
          << endl;
-    cout << " --broadcast <broadcast_queue>            Duplicates receive ports for each connected client with the specified broadcast_queue length. Broadcast outputs have higher latency but less packet loss." << endl;
+    cout << " --broadcast <broadcast_queue>            Duplicate receive ports with the specified broadcast_queue length. "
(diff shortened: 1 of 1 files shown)


dyfer:
Thanks!
We've discussed this at the dev meeting today and it seems that we'd prefer to have this PR (or most PRs) directed at `dev`. In that case you'd need to resolve conflict with the changes in `dev` - the conflict shows up if I just change the base branch to `dev` in this PR...
CC @cchafe

dyfer:
Thanks! 
The new code takes up only one line, while the surrounding code uses multiple lines per parameter. Is this to be consistent with `clang-format`? Or should we change this to be consistent with the surrounding code?

ntonnaett:
@michaeldessen Are you okay with my changes?

ntonnaett:
I'm not totally satisfied with the wording. I think the impact of broadcast_length is not clear enough.

michaeldessen:
Sorry I am so late seeing this! Yes, looks good, thank you @ntonnaett!

dyfer (approved):
I'm not necessarily thrilled with the whitespace changes, but otherwise I'm good with this PR. Feel free to merge as is or undo the whitespace changes and then merge.

PR #257, opened 2021-03-27
Build static Qt in GHA on macOS and Windows

dyfer:
I've added static Qt builds to our GHA - qmake only for now.

Thanks @wholenotes for all the pointers. I've also found [this page](https://retifrav.github.io/blog/2018/02/17/build-qt-statically/) to be helpful, especially for the Windows build.

Here are some notes:
- Qt is built once and then cached
- initial build takes a long time (almost an hour), but subsequent builds for a given branch are fast, see [this job on my fork](https://github.com/dyfer/jacktrip/runs/2210027958?check_suite_focus=true)
- I used the latest Qt5 - 5.15.2; on Windows that probably doesn't matter (it supports Windows 7 and later). On macOS this probably means that we support only macOS 10.13 and later; I think previous JT releases used older Qt and thus might've supported older OSs (?). We could change the Qt version that we use, or even build two versions (the latter is probably an overkill?)
- on Windows Qt is built using `mingw32`, like JT; in the process of working on this I also got the [MSVC build working, linked here for reference](https://github.com/dyfer/jacktrip/blob/mp/static-qt-msvc/.github/workflows/jacktrip.yml#L115)
- ~~[this change](https://github.com/jacktrip/jacktrip/compare/dev...dyfer:mp/static-qt-gha?expand=1#diff-12d8f0d8ae7fbe32ec242aac7bb0b0402495ff5553c089cfcef9576b2ad836ceR116) seemed like a good idea, but I'm not sure if it's correct, could someone confirm?~~ I've tested [Windows build without this change](https://github.com/dyfer/jacktrip/actions/runs/730676824) - it still worked fine and had no external dependencies aside from libjack and system libraries. In that case I removed this change from my PR.
- macOS Qt building step is labeled `Linux / macOS`, since according to the website linked earlier the same commands should work on Linux; I have not tried building static Qt on Linux though
- I've set the qt build to [skip numerous modules](https://github.com/jacktrip/jacktrip/pull/257/files#diff-400b21146a0aa3667e1c8f1d95c364ac50306b03359d195bff0e90725ecaabf8R62), which I don't think we're using anyway. This of course can be further changed/tweaked if we need in the future. The [cache key](https://github.com/jacktrip/jacktrip/pull/257/files#diff-400b21146a0aa3667e1c8f1d95c364ac50306b03359d195bff0e90725ecaabf8R61) needs to be updated to trigger rebuilding Qt.

### DIFF
--- a/.github/workflows/jacktrip.yml
+++ b/.github/workflows/jacktrip.yml
@@ -25,16 +25,30 @@ jobs:
         include:
           - name: Linux
             runs-on: ubuntu-18.04
+            qmake-spec: linux-g++
+            make-cmd: make
+
           - name: macOS
             runs-on: macos-10.15
+            qmake-spec: macx-clang
+            make-cmd: make
+
           - name: macOS-static
             runs-on: macos-10.15
+            qmake-spec: macx-clang
+            make-cmd: make
             static: true
+
           - name: Windows
             runs-on: windows-2019
+            qmake-spec: win32-g++
+            make-cmd: mingw32-make
+
           - name: Windows-static
-            static: true
             runs-on: windows-2019
+            qmake-spec: win32-g++
+            make-cmd: mingw32-make
+            static: true
 
     env:
       SRC_PATH: ${{ github.workspace }}/src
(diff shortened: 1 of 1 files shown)


cchafe (approved):

cchafe:
works great on win10, tested against jackd 1.9.11 on a VM and a laptop

PR #263, opened 2021-04-29
Build with rtaudio

dyfer:
I've added rtaudio static to our build system for macOS and Windows. RtAudio support is added to `qmake` with `-config rtaudio` flag.

Have not tested extensively - it just builds and runs for now. There seem to be no external dependencies (i.e. static library seems to be properly used).

I've removed some dead code in `jacktrip.pro` file from previously included rtaudio in the source. I'm using rt-audio from vcpkg and pkg-config to get the include paths and linker flags, so this means that windows builds with rtaudio will require installing pkg-config and rtaudio separately. I'd be happy to add that to the docs in another PR. (Building without rtaudio is still possible and doesn't require either library).

On Windows, I had a problem adding proper linker flags - it seems that we need to link the main app with Windows libraries when using rtaudio (not sure if it's supposed to be like that - shouldn't the static library be linked to these already?) Anyway, the problem was that the additional linker flags need to come _after_ `-lrtaudio`, but when using `pkg-config` in `qmake`, linker flags from pkg-config are added _after_ linker flags specified manually. Thus on Windows I added linker flags including `-lrtaudio` (so `-lrtaudio` ends up being used [twice](https://github.com/dyfer/jacktrip/runs/2471276875?check_suite_focus=true#step:12:1012)) which was needed for successful linking.

### DIFF
--- a/.github/workflows/jacktrip.yml
+++ b/.github/workflows/jacktrip.yml
@@ -32,22 +32,26 @@ jobs:
             runs-on: macos-10.15
             qmake-spec: macx-clang
             make-cmd: make
+            rtaudio: true
 
           - name: macOS-static
             runs-on: macos-10.15
             qmake-spec: macx-clang
             make-cmd: make
+            rtaudio: true
             static: true
 
           - name: Windows
             runs-on: windows-2019
             qmake-spec: win32-g++
             make-cmd: mingw32-make
+            rtaudio: true
 
           - name: Windows-static
             runs-on: windows-2019
             qmake-spec: win32-g++
             make-cmd: mingw32-make
+            rtaudio: true
             static: true
 
     env:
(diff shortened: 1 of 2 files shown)


cchafe (approved):

josmithiii (commented):
Instead of 
```
  brew install rt-audio
  rm -f `brew --prefix rt-audio`/lib/*.dylib # remove the dynamic library, we want static only
```
how about
```
  brew install rt-audio &&  rm -f \`brew --prefix rt-audio\`/lib/librtaudio*.dylib # remove ...
```

dyfer:
@josmithiii sure, should I make another PR as this has been already merged?

josmithiii:
That sounds right to me - I'll watch for it

PR #264, opened 2021-04-30
GitHub Actions: use jack 1.9.16 on macOS

dyfer:
Currently there is jack 1.9.17 available in homebrew. Linking against this version prevents from running with jackosx, due to difference in library name (no API conflicts), see https://github.com/jackaudio/jack2/issues/738
Linking against Jack 1.9.16 should make JackTrip compatible with both JackOSX and the newer Jack2.

EDIT: 1.3.0 release of JT is linked against Jack 1.9.16 since this was the version that was in homebrew at the time of releasing.


dyfer:
Based on feedback in previous discussions, I'm going to push this directly to `dev`

josmithiii (approved):

PR #265, opened 2021-04-30
GHA small cleanup

dyfer:
Small cleanup in GitHub Actions

### DIFF
--- a/.github/workflows/jacktrip.yml
+++ b/.github/workflows/jacktrip.yml
@@ -84,8 +84,7 @@ jobs:
             brew link qt5 --force
           fi
           if [[ "${{ matrix.rtaudio }}" == true ]]; then 
-            brew install rt-audio
-            rm -f `brew --prefix rt-audio`/lib/*.dylib # remove the dynamic library, we want static only
+            brew install rt-audio && rm -f `brew --prefix rt-audio`/lib/*.dylib # remove the dynamic library, we want static only
           fi
       - name: install dependencies for Windows
         if: runner.os == 'Windows'


josmithiii (commented):
I tested on Mac OS and discovered a losing case: If rtaudio is already installed, then "brew install" will fail and the deletion of the dynamic library will not happen.  That would argue for leaving this alone.  However, in that case, I am worried about what happens when there is no rtaudio installed:  Will the "rm -f" be safe in all cases?  On Mac OS, it appears to work ok whether rt-audio is installed or not, depending only on the existence of a formula for it.

dyfer:
This code is only used in GH actions, i.e. in a somewhat controlled environment, so IMO we don't need to be concerned whether it works for people when run on their machines. Currently we are not instructing people to install rtaudio anyway. And if we did, I don't think we'd recommend removing the dynamic library - if the build is meant to be done locally, then it will run just fine with dynamically linked rtaudio. 

> However, in that case, I am worried about what happens when there is no rtaudio installed: Will the "rm -f" be safe in all cases?

In what case rtaudio would not be installed? When there's a network error etc?

As a side note, I made this change on your request, but I personally don't think there's much difference in GHA whether we do this install and removal in one line or in two... what do you think?

josmithiii:
Ok, thanks for the clarifications.  I am new to GH actions and did not understand the execution context.  I thought it was an installation script!

Yes, I was worried about rtaudio failing to install for any reason, the brew prefix being blank as a result, and every dylib in /lib/* getting deleted.

As you know, I prefer the practice of letting trusted developers commit and push freely to the dev branch (which is how I function on the Faust and moForte teams - super efficient!), but if you want a second pair of eyes on things, we should assemble a team that spans your contribution areas.  My role on this team, beyond the usual "signal processing consultant" role, should be Mac OS user-side code review and tests, since that's what I do and know best.  I can also do Linux on a laptop or VM, but I have essentially zero experience on anything server-side like GH actions.

dyfer:
> Yes, I was worried about rtaudio failing to install for any reason, the brew prefix being blank as a result, and every dylib in /lib/* getting deleted.

I see, good point about what would happen if homebrew was not installed.

However, since this is in GHA, then I don't think we need to worry about it.

> As you know, I prefer the practice of letting trusted developers commit and push freely to the dev branch 

Based on this and on the conversations at the last meeting maybe I'll commit directly changes that I feel confident about and ask for second pair of eyes on things that I am unsure about, does that sound good to you?

josmithiii (approved):

josmithiii:
Sounds good!

On this particular one, I approved it and you can merge it or not as you prefer.

dyfer:
Thanks! I'll leave this one unmerged.

Issue #69, opened 2020-06-08
Add manpage


ntonnaett:
Manpage is missing. Especially packagers like to have manpages. There should be a way to have one source that can generate help and man.

dyfer:
I think this is fixed in #268

PR #268, opened 2021-05-12
Add manpage

ntonnaett:
Meson generates a manpage with help2man which parses --help and --version of jacktrip. The modifications to --help are for the output of the manpage.

```
meson builddir
cd builddir
meson compile
man -l jacktrip.1
```


ntonnaett:
Solves #69

dyfer:
This is great, thanks for the addition. 
Just in terms of workflow, how will the packagers know to generate the man page using meson? Does this need to be documented/set somewhere?

ntonnaett:
If help2man is present on your system, meson automatically generates the manpage and installs it to the right location. I think packagers know help2man and meson tells them if it's missing. But I think the whole build documentation could be improved. For example, I just forgot about installing jacktrip with `sudo meson install`.

It might be good to talk about what the default build system should be. I just tested the qmake build and it fails on my side because it can't link to the pipewire's jack.

ntonnaett:
I should add help2man to the dependency list.

dyfer:
> If help2man is present on your system, meson automatically generates the manpage and installs it to the right location. I think packagers know help2man and meson tells them if it's missing. But I think the whole build documentation could be improved. For example, I just forgot about installing jacktrip with `sudo meson install`.

So should the manpage be generated now in the CI? I looked at the meson linux artifacts and couldn't see it (though I'm not sure if I'm looking for the right thing)

> I just tested the qmake build and it fails on my side because it can't link to the pipewire's jack.

How does one suppose to link with piperiwire? What's the same, what's different than linking with jack?

> It might be good to talk about what the default build system should be.

I agree, I'd prefer to maintain one build scenario. But it seems like there was no desire to make a switch from qmake last time we talked about this.

ntonnaett:
> So should the manpage be generated now in the CI? I looked at the meson linux artifacts and couldn't see it (though I'm not sure if I'm looking for the right thing)
> 
Good point. I will add this.

> How does one suppose to link with piperiwire? What's the same, what's different than linking with jack?
It's actually the same. So I guess qmake just doesn't find it. CMake works fine.

> > It might be good to talk about what the default build system should be.
> 
> I agree, I'd prefer to maintain one build scenario. But it seems like there was no desire to make a switch from qmake last time we talked about this.

Yes, I think that needs good instructions that show the benefits too. I try to install Windows again to test that side too. The virtual machine is just annoying.

ntonnaett:
Just to elaborate a bit. Pipewire installs their jack stuff at another place than JACK. That's good if you accidentally install both.
```
$ pkgconf --libs jack
-L/usr/lib64/pipewire-0.3/jack -ljack -ljacknet
```
Pkgconfig gives you linker flags where -L gives you the place to search. Maybe qmake doesn't use pkgconfig?

dyfer:
> > How does one suppose to link with piperiwire? What's the same, what's different than linking with jack?
> 
> It's actually the same. So I guess qmake just doesn't find it. CMake works fine.

hm, looks like `jacktrip.pro` just hardcodes `-ljack` instead of using pkg-config. We should probably change that (for Linux and macOS)

dyfer (approved):
Thanks!

PR #270, opened 2021-05-13
Build rtaudio subproject statically (except for Windows)

ntonnaett:
This should always compile rtaudio statically when it is a subproject.


ntonnaett:
The installation doesn't work yet...

ntonnaett:
Installation worked but it installed the library that we don't need or want.

ntonnaett:
MSVC seems to have problems with mixed linking. So let's not build rtaudio statically on windows.

cchafe (approved):

PR #273, opened 2021-05-19
Set udp realtime prio below jack's

ntonnaett:


ntonnaett:
Needs some refactoring for platform and audio backend dependent parts.

PR #274, opened 2021-05-20
pipewire versions of scripts/hubMode with args for FPP

cchafe:
allows testing of a hub server to a hub client, both on localhost
for example, from 
jacktrip/scripts/hubMode
run
./test_hub_mode_server_and_clientPipeWire.sh  ../../builddir/jacktrip 32
the scripts set the environment variable identically for each process
PIPEWIRE_LATENCY="$FPP/48000"
the same scripts run ok on a non-PipeWire machine, so they can be used to compare Jack natively and
Jack under PipeWire (for the former, jackd needs to be started separately)


cchafe:
Should be ready to use in CI (does require jack_iodelay be installed)
run it from the scripts/hubMode directory with
./test_hub_mode_server_and_clientPipeWire.sh  ../../builddir/jacktrip 32 1

ntonnaett:
I'm just wondering if the pipewire in the script names are confusing because they run on native jack too. What do you think?

ntonnaett (approved):

cchafe:
Good idea, thanks -- renamed, tested and pushed

PR #283, opened 2021-05-22
Qmake: use pkg-config for Jack

dyfer:
This should allow for linking to jack in various locations, particularly for building with pipewire.


ntonnaett (approved):
This builds with pipewire's jack library. Thank you!

PR #290, opened 2021-05-24
Fix jacktrip object creation in UI

psiborg112:
Need to update to reflect new signature created by asynchronous channels.


psiborg112:
Made redundant by #291

PR #293, opened 2021-05-25
Install jacktrip.desktop file and icon for Linux

ntonnaett:
This PR adds a desktop file and a scalable icon that will be installed with meson on linux. Now you will find Jacktrip in most desktop environments and it will have an icon. There might be some bitmap icons missing for older platforms.


psiborg112:
Thanks for taking care of that. (It is slightly ridiculous that linux is my main platform and yet it's the one that I've done the least amount of work on as far as any sort of packaging goes.)

psiborg112 (approved):

PR #294, opened 2021-05-27
Update build script, create win/mac installers and mac application bundle

dyfer:
I've updated the qmake build script and added creating macOS bundle to GitHub Actions (qmake build).

~~I've made following changes that I'd like to get feedback on (particularly @psiborg112 ):~~
- ~~In the bundle creation script, I changed `QJackTrip` to `JackTrip`. IIUC this is consistent with what we decided on the app name. Is that OK @psiborg112 ?~~
- ~~I've also renamed folder `os x` to `macos`; I'm happy to revert that change if you don't like it, please let me know.~~

The above changes were done outside of this PR.

I've also removed dynamic builds for macOS and Windows, and added a new Linux build with `nogui` option (also I plan to use that for running tests later). I think I'd also like to switch the main Linux build to static, just so it can be easily used for testing etc.

I've updated build artifacts so just the binary is uploaded, instead of the whole directory.

Any and all feed back welcome!

EDIT: 
Current list of changes:
- [x] Updated qmake build script
  - [x] takes command line options `./build [noclean nojack rtaudio nogui static [-config static]]`
  - [x] works also on Windows with MinGW
- [x] removed (commented out) dynamic Windows/macOS builds
- [x] added static Linux build
- [x] ~~renamed `os x` to `macos` folder~~ done in #303
- [x] ~~renamed `qjacktrip` to `jacktrip` in `win` and `macos` folders (**looking for confirmation from everyone whether this is what we want**)~~ done in #303 and #308
- [x] ~~added building Qt tools in static Qt build to get `macdeployqt`/`windeployqt` (BUT maybe we don't need it? see below)~~
- [x] added app bundle creation on macOS
- [x] added installer creation on macOS (no signing!)
- [x] added installer creation on Windows
- [x] ~~did not use provided scripts for creating bundle/installer~~ I used updated scripts for creating bundle and installer
- [x] ~~I did not use `macdeployqt`/`windeployqt` - I don't think we need that for static builds, do we?~~
  - ~~now that we are building app bundle / installer, we could consider dynamic builds as well... but a single static binary is nice~~ now handled by the scripts
- [x] Windows: I modified the installer's list (`files_static.wxs`)
  - [x] tested windows installer
- [x] macOS: I modified installer file removing Aaron's certificate
- [x] Added version string to the artifacts: it's commit SHA (long) for non-tagged build, and a tag name for tagged builds (releases)
  - [ ] Question: would we want to shorten the SHA to first 7 digits, like e.g. on the [commit list](https://github.com/jacktrip/jacktrip/commits/dev) in github?
- [ ] Question (@psiborg112): **can macOS binaries be signed after being packaged into an installer?** 

### Here's the current [artifact list](https://github.com/jacktrip/jacktrip/actions/runs/887636201#artifacts) from this PR.

### DIFF
--- a/.github/workflows/jacktrip.yml
+++ b/.github/workflows/jacktrip.yml
@@ -45,7 +45,7 @@ jobs:
             static: true
             binary-path: jacktrip # relative to builddir
             app-bundle: true
-            bundle-installer-path: macos/JackTrip.app # relative to base (repo) directory
+            bundle-installer-path: bundle # relative to builddir
 
           # - name: Windows-dynamic
           #   runs-on: windows-2019
@@ -189,6 +189,8 @@ jobs:
           fi
           cd macos
           ./assemble_app.sh
+          mkdir -p $BUILD_PATH/${{ matrix.bundle-installer-path }}
+          cp -R JackTrip.app $BUILD_PATH/${{ matrix.bundle-installer-path }}
       - name: upload artifacts
         uses: actions/upload-artifact@v2
         if: matrix.binary-path
@@ -200,4 +202,4 @@ jobs:
         if: matrix.app-bundle == true || matrix.installer == true
         with:
           name: JackTrip-${{ matrix.name }}-${{ matrix.suffix }} # version number could be added here
-          path: ${{ github.workspace }}/${{ matrix.bundle-installer-path }}
+          path: ${{ env.BUILD_PATH }}/${{ matrix.bundle-installer-path }}
--- a/.github/workflows/jacktrip.yml
+++ b/.github/workflows/jacktrip.yml
@@ -200,14 +200,30 @@ jobs:
             CONFIG_STRING="nogui $CONFIG_STRING"
           fi
           ./build $CONFIG_STRING
-      - name: create app bundle
-        if: runner.os == 'macOS' && matrix.bundle-path
+      - name: create app bundle and installer on macOS
+        if: runner.os == 'macOS' && (matrix.bundle-path || matrix.installer-path)
         run: |
-          if [[ "${{ matrix.static }}" == true ]]; then 
-            export PATH=$QT_STATIC_BUILD_PATH/bin:$PATH
-          else
-            export PATH=`brew --prefix qt5`/bin:$PATH
+          CONFIG=
+          if [[ -n "${{ matrix.installer-path }}" ]]; then 
+            CONFIG="-i $CONFIG"
           fi
+          cd macos
+          ./assemble_app.sh $CONFIG
+          if [[ -n ${{ matrix.bundle-path }} ]]; then
+            mkdir -p $BUILD_PATH/${{ matrix.bundle-path }}
+            cp -R *.app $BUILD_PATH/${{ matrix.bundle-path }}
+          fi
+          if [[ -n ${{ matrix.installer-path }} ]]; then
+            mkdir -p $BUILD_PATH/${{ matrix.installer-path }}
+            cp -R package/build/* $BUILD_PATH/${{ matrix.installer-path }}
+          fi
+        run: |
+          CONFIG=
+          if [[ -n "${{ matrix.installer-path }}" ]]; then 
+            CONFIG="-i $CONFIG"
+          fi
+          cd macos
+          ./assemble_app.sh $CONFIG
       - name: upload jacktrip binary
         uses: actions/upload-artifact@v2
         if: matrix.jacktrip-path
--- a/.github/workflows/jacktrip.yml
+++ b/.github/workflows/jacktrip.yml
@@ -27,6 +27,11 @@ jobs:
             runs-on: ubuntu-18.04
             binary-path: jacktrip
 
+          - name: Linux-nogui
+            runs-on: ubuntu-18.04
+            binary-path: jacktrip
+            nogui: true
+
           - name: macOS
             runs-on: macos-10.15
             rtaudio: true
@@ -191,6 +196,7 @@ jobs:
           ./assemble_app.sh
       - name: upload artifacts
         uses: actions/upload-artifact@v2
+        if: matrix.binary-path
         with:
           name: JackTrip-${{ matrix.name }} # version number could be added here
           path: ${{ env.BUILD_PATH }}/${{ matrix.binary-path }}
--- a/.github/workflows/jacktrip.yml
+++ b/.github/workflows/jacktrip.yml
@@ -23,19 +23,15 @@ jobs:
       fail-fast: false # don't abort if one of the build failse
       matrix:
         include:
-          - name: Linux
+          - name: Linux-x64
             runs-on: ubuntu-18.04
+            static: true
             binary-path: jacktrip
 
-          - name: Linux-nogui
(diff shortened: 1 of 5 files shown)


psiborg112:
Change to macos is fine, but I thought we'd only decided on the change to the app binary name and not the bundle name? There's the very real possibility here of less technical users who have already been using QJackTrip installers and bundles ending up with parallel installs of different versions of the app. (I'd like to hear @wholenotes opinion on this one.)

wholenotes:
Hi Aaron,

Yea, we would definitely want to avoid any possibility of duplicate installations. Could you step through how this might happen?

Bonnie

‐‐‐‐‐‐‐ Original Message ‐‐‐‐‐‐‐

psiborg112:
Anyone who's previously installed QJackTrip.app and doesn't explicitly delete it before installing JackTrip.app will end up with both installed. It would be very easy for them to keep running the old version then out of habit.

dyfer:
> Change to macos is fine, but I thought we'd only decided on the change to the app binary name and not the bundle name?

At the meeting today I've asked about the change from QJackTrip and others did express the same sentiment as previously - it's confusing to have a different name between the software itself and the application name

> Yea, we would definitely want to avoid any possibility of duplicate installations. 

- we won't be offering an installer for macOS - since it's an app bundle people will just drag it to `/Applications`. IMO it would be enough to note the name change (and instruction to delete the old version) on the release/downloads page
  - for technical users  running from the command line, we'll note that they should link jacktrip from `/Applications/JackTrip.app...` to `/usr/local/bin`

I haven't used QJackTrip much previously, is there an uninstaller for it? If not, we should definitely create one...

psiborg112:
But we need an installer for a symlink to be created that allows access to jacktrip from the command line.

dyfer:
> But we need an installer for a symlink to be created that allows access to jacktrip from the command line.

In the conversations the consensus was that since this is mostly needed for advanced users, we can just instruct them to create a symlink themselves. This was expressed by others during the meeting.

I don't feel strongly about it, but I also lean towards not needing the installer and just instructing (advanced) users to make that symlink. If they want to use JT from the command line, they're likely familiar enough with it, I think.

psiborg112:
But even a lot of advanced users don't know where in an app bundle an executable is. Why not simplify that process?

dyfer:
> But even a lot of advanced users don't know where in an app bundle an executable is. 

IMO simple instruction on the download page/readme would be sufficient to solve that.

> Why not simplify that process?

I'm personally not a fan of installers unless absolutely necessary - I'd rather drag the app myself. But again - this should be a group decision, and when we discussed it didn't seem like people see the need of an installer. I'd be happy to add one if this is what we decide.

psiborg112:
If it comes down to it I'm happy to maintain my own installers.

dyfer:
> If it comes down to it I'm happy to maintain my own installers.

If you feel that strongly about the installer, I can make one as part of this PR :)

psiborg112:
It's not that important.

dyfer:
What I meant was I think it's better to have a centralized place with one version of JT, instead of "this is the official release, but if you'd like an installer go to http://...". So if you plan to publish an installer of JT, I think it's better to include it as a part of the release.

wholenotes:
An installer would be ideal UX. Thanks for putting in the extra effort, Aaron. There are folks who aren’t even getting past the installation process, let alone operating JackTrip. Good to know we’re taking care of them.

Bonnie

dyfer:
I've added an installer for macOS, as well as Windows (Windows untested ATM). I've summarized all the changes in the first comment.

psiborg112:
>     * [ ]  Question (@psiborg112): **can macOS binaries be signed after being packaged into an installer?**

Don't know off hand. It might be possible, but I've never tried it myself.

wholenotes:
I haven’t looked into the signing process in such detail either.

macos/JackTrip.app_template/Contents/Info.plist
```diff
@@ -11,11 +11,11 @@
 	<key>CFBundleIconFile</key>
 	<string>jacktrip</string>
 	<key>CFBundleIdentifier</key>
-	<string>org.psi-borg.qjacktrip</string>
+	<string>org.psi-borg.JackTrip</string>
```
psiborg112, line 14: It might be worth leaving the bundle identifier as is. It's not user facing, and on the installer/upgrade front, we may be able to leverage it to guard against parallel installation issues. Either way, an all lowercase bundle ID is generally considered best practice. (There are some weird case issues that can occur otherwise.)

psiborg112: Also, have just created an updated bundle script in the macos-assemble-script branch. It takes a version number as an argument so that that gets consistently set in both the bundle and the package, and you can optionally specify the app name and bundle identifier. (If you don't it reverts to defaults, which are currently my defaults, but could easily be changed.)

psiborg112: (So, whatever gets decided on the naming front, it's a simple variable change. At least on the macos front.)

wholenotes:
Perhaps the install script could check for an existing bundle named qjacktrip and ask the user for permission to uninstall it first, to guard against parallel installs? It would be nice to have an app bundle with the same name as the application.

macos/JackTrip.app_template/Contents/Info.plist
dyfer: I'll take a look, thanks!

dyfer:
> Perhaps the install script could check for an existing bundle named qjacktrip and ask the user for permission to uninstall it first, to guard against parallel installs? It would be nice to have an app bundle with the same name as the application.

That might be a good idea. I'm very new to creating an installer so maybe with this PR I'd like to focus on getting it into github actions and then we can tweak it as needed? If not, I'd be glad to accept any advice how to best perform such check and add it to the installer.

Also, this PR is now waiting on #303, after which I'll rebase and update it.

dyfer:
I think this is ready for review now. Windows installer seems to work.

> Perhaps the install script could check for an existing bundle named qjacktrip and ask the user for permission to uninstall it first, to guard against parallel installs? It would be nice to have an app bundle with the same name as the application.

If we want this, I would like it be be a separate issue or PR, and keep this one to get the base GHA infrastructure for the bundle/installer creation.

dyfer:
Question: do we want to shorten the SHA in filenames to 7 digits? I'm leaning toward shortening it.

psiborg112:
> I think this is ready for review now. Windows and macOS installers seem to work.
> 
> > Perhaps the install script could check for an existing bundle named qjacktrip and ask the user for permission to uninstall it first, to guard against parallel installs? It would be nice to have an app bundle with the same name as the application.
> 
> If we want this, I would like it to be a separate issue or PR, and keep this one to get the base GHA infrastructure for the bundle/installer creation.

Agreed. I think this is for a later PR. All looks good to me. (Did the windows installer for static builds end up working?)

psiborg112 (approved):

dyfer:
>  (Did the windows installer for static builds end up working?)

Yes! I installed it and it works. It even warned me first that i had an older version installed!

PR #296, opened 2021-05-27
GitHub Actions: automatically deploy mkdocs pages

dyfer:
I'm not sure if I can test this before it's merged... IMO we could merge this and see if there are any issues.

### DIFF
--- a/.github/workflows/mkdocs.yml
+++ b/.github/workflows/mkdocs.yml
@@ -1,0 +1,23 @@
+name: Publish docs to GitHub Pages
+on:
+  push:
+    branches:
+      - main
+      - dev
+    paths:
+      - mkdocs.yml
+      - docs/**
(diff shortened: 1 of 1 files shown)


wholenotes (approved):

ntonnaett:
```
    paths:
         - mkdocs.yml
         - docs/**
```
That means only deploy docs if changes were made to the docs?

ntonnaett:
Maybe the GITHUB_TOKEN needs some love from @cchafe afterwards:
  

> Note that for this approach, Github Pages will be enabled in Settings but you will not have a URL displayed or environment tab > yet. So change the Github Pages settings to another target and then back again to gh-pages (if that is your branch to serve) - then you will see a URL. This step is only needed on the first deploy and no action is needed later on.

Issue #284, opened 2021-05-22
Version string setting


dyfer:
Right now version string is set in multiple places:
Primarily in https://github.com/jacktrip/jacktrip/blob/dev/src/jacktrip_globals.h#L47
but also:
https://github.com/jacktrip/jacktrip/blob/dev/os%20x/package/QJackTrip.pkgproj#L840
https://github.com/jacktrip/jacktrip/blob/dev/win/qjacktrip.wxs#L6
https://github.com/jacktrip/jacktrip/blob/dev/meson.build#L1

We at least need to remember to change them all for the release... but it would be even better to set it in one place and automatically populate it the other places.

ntonnaett:
I thought about adding this to meson but I only found the place in jacktrip_globals.h. The good thing is that it looks like cmake uses the same syntax for this.

psiborg112:
#303 takes care of most of this. (The exception being the version in the meson.build file.) The scripts to build the installers on mac and windows now take the version from the binary itself.

dyfer:
Thanks @psiborg112 
@ntonnaett do we need that version identifier in the meson file?

ntonnaett:
The version identifier in meson is not used for anything right now. We can remove it.

dyfer:
Great, let's do that. I think after that's done we can close this issue

PR #297, opened 2021-05-28
Win installer UI

psiborg112:
Use WixUIExtension to start to develop a user interface for the windows installer.


src/gui/about.cpp
```diff
@@ -27,7 +27,7 @@
 #include "ui_about.h"
 #include "../jacktrip_globals.h"
 
-const QString About::sBuildID = "2021052500";
+const QString About::sBuildID = "2021052600";
```
dyfer, line 30: I think it would be good to talk about version identifiers in general - see https://github.com/jacktrip/jacktrip/issues/284

Why do we need a build ID?

psiborg112: I was using the build ID (and still do) to track my binary releases because QJackTrip dev was happening at a different pace to JackTrip dev. By default this is probably not needed any more, and should probably be set at build time, not in the code. Will update.

wholenotes:
I was wondering about the build ID. Thanks for the update.

Bonnie

dyfer (approved):
The changes look good to me. I have not tested this though - I'm thinking that the easiest will be to merge this first and then I'll rebase #294 on top of if (which will give us installer in the artifacts)

PR #299, opened 2021-05-28
gha: Add static nojack jobs

mattahorton:
Given that JACK is a huge barrier to entry for use of JackTrip, I am often most interested in static, nojack builds of jacktrip. I can't be the only one, and instead of building these manually myself, I thought I'd add them to the Github Actions. 

Now, I do know that @dyfer is currently doing some work on the build scripts and has some ideas on how the scripts might be better structured. If you decide not to accept this PR and instead work it into that work, that makes sense!

### DIFF
--- a/.github/workflows/jacktrip.yml
+++ b/.github/workflows/jacktrip.yml
@@ -41,6 +41,14 @@ jobs:
             rtaudio: true
             static: true
 
+          - name: macOS-static-nojack
+            runs-on: macos-10.15
+            qmake-spec: macx-clang
+            make-cmd: make
+            rtaudio: true
+            static: true
+            nojack: true
+
           - name: Windows
             runs-on: windows-2019
             qmake-spec: win32-g++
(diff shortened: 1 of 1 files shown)


dyfer:
We've discussed this with @mattahorton and I think we will try to use weakjack first (it's on my todo list). Once we are sure weakjack works, we can close this PR.

dyfer:
I think this is now superseded by #313

@mattahorton could you test the latest [dev build](https://github.com/jacktrip/jacktrip/actions/runs/904483918) on Windows to make sure it does indeed work both with and without Jack installed? (any issues with rtaudio notwithstanding)

mattahorton:
I'm sorry i didn't follow up on this @dyfer. I saw that this was closed and not that you wanted me to test.

I don't have access to my PC right now, but if you'd still like me to, I will do so in about a week.

dyfer:
@mattahorton if you could test that would be great, thanks!

PR #300, opened 2021-05-29
Cross compile JackTrip for windows

ntonnaett:
This PR depends on #301 . Please merge afterwards.

That's the first step to cross compile for Windows. The cross compilation instructions are tested on Fedora 34.

- RtAudio subproject is linked statically. I disabled static linking on Windows before.
- Qt is not linked statically. I need static qt. Can you help @dyfer?

Edit: I think static linking is not working before this is resolved: https://github.com/mesonbuild/meson/pull/7733


dyfer:
> Qt is not linked statically. I need static qt. Can you help @dyfer?

Do you mean help with instructions?
For building static Qt with MinGW64, it is currently done in [qmake GHA build](https://github.com/jacktrip/jacktrip/blob/dev/.github/workflows/jacktrip.yml#L152). I think it needs to run in `cmd` because if the `.bat` file so not sure how that would work for cross-compiling. 

So what's with the `mingw64-qt5-qtbase-static` package, doesn't that give us the right library?

ntonnaett:
> Do you mean help with instructions?
> For building static Qt with MinGW64, it is currently done in [qmake GHA build](https://github.com/jacktrip/jacktrip/blob/dev/.github/workflows/jacktrip.yml#L152). I think it needs to run in `cmd` because of the `.bat` file so not sure how that would work for cross-compiling.

I wasn't clear at all. There are two different static builds that I need to get done. The GHA static build on Windows. There I need what you pointed me to. Can I use the same qt static artifact although it's a different GH Action?

> So what's with the `mingw64-qt5-qtbase-static` package, doesn't that give us the right library?

The other thing is that it looks like qt really likes to use shared objects. Meson's qt module seems to not comply to the default_library option. I think that's what the meson issue I posted seems to be about. But I have to dig deeper.

ntonnaett:
Just another reason for the rtaudio subproject. There's no mingw rtaudio package. So when cross compiling we need to compile it from source.

dyfer:
> The GHA static build on Windows. There I need what you pointed me to. Can I use the same qt static artifact although it's a different GH Action?

I think you should copy both the build and cache steps into meson GHA and use a different cache path (e.g. add meson) just in case.

BUT I was thinking about bringing meson builds into the the `jacktrip.yml` soon, after #294 is merged (so maybe sometime next week?), there you'd have Qt set up for static builds - would you like to wait for this or continue with meson separately?

ntonnaett:
Let's merge #301 and then this one. Static GHAs shouldn't be part of this.

I've added a last change to have a separate directory for cross files. There's a extra static file that sets some extra flags.

ntonnaett:
This fixes nogui with meson too.

dyfer:
> Let's merge #301 and then this one. Static GHAs shouldn't be part of this.

Is there a chance we could also be set up to cross-compile on Ubuntu? I'm particularly thinking about cross-compiling for armv8 (RPi 4)

ntonnaett:
There are mingw-w64 packages on Ubuntu. I didn't find qt. The cross files don't have a path. So they might just work on Ubuntu if the right packages are installed.

Cross compiling for ARM for Raspberry Pi is slightly more difficult, because we might not have precompiled qt packages. But compiling at all shouldn't be a big problem. There's gcc-c++-aarch64-linux-gnu.x86_64 on Fedora. I assume all other necessities too.

dyfer:
Yeah for Qt I'd think we'd compile a static one anyway. 
However, we'd need to compile jack, right? since it needs to be linked with the library build for the target architecture...?

As for this PR - I can't comment on cross-compiling, as I can't test that. Do you want to get anything else here or are you looking on feedback or anything in particular?

ntonnaett:
I tested this with wine. As this is new and shouldn't break anything, I think we don't need more testing for the cross compilation.
It will become more interesting when qt static is working with meson.

nogui seems to be fixed. And windows doesn't try to build a manpage anymore. I think this can be merged.

ntonnaett:
> Yeah for Qt I'd think we'd compile a static one anyway.
> However, we'd need to compile jack, right? since it needs to be linked with the library build for the target architecture...?
Right. But I don't want to build jack myself. It looks like a really complicated thing. And they use waf. Maybe we just use weakjack?

dyfer:
> Right. But I don't want to build jack myself. It looks like a really complicated thing. And they use waf. 

I think we should just add waf for building JT. Simple.

Just kidding ;)

> Maybe we just use weakjack?

That's a great idea for this! Haven't thought of it. We'll cross that bridge when we get there though.

dyfer (approved):

PR #301, opened 2021-05-29
[Meson] Use rtaudio meson subproject instead of cmake

ntonnaett:
This uses a RtAudio subproject build with meson instead of cmake. With this I get rid of a lot of linking errors on Windows. Qt is linked dynamically.

### DIFF
--- a/.github/workflows/meson.yml
+++ b/.github/workflows/meson.yml
@@ -86,7 +86,7 @@ jobs:
         run: |
           meson --buildtype release $BUILD_PATH
           cd $BUILD_PATH
-          meson configure -Ddefault_library=static
+          meson configure -Ddefault_library=static -Drtaudio=enabled
           meson compile
       - name: build on Windows
         if: runner.os == 'Windows'
--- a/.github/workflows/meson.yml
+++ b/.github/workflows/meson.yml
@@ -93,7 +93,7 @@ jobs:
         run: |
           meson --buildtype release builddir
           cd builddir
-          meson configure -Ddefault_library=static
+          meson configure -Ddefault_library=static -Drtaudio=enabled
           meson compile
       - name: upload artifacts
         uses: actions/upload-artifact@v2
--- a/meson.build
+++ b/meson.build
@@ -86,9 +86,12 @@ if rtaudio_dep.found() == true
 endif
 
 if host_machine.system() == 'windows'
+	deps += compiler.find_library('ws2_32', required: true)
+endif
+
+if compiler.get_id() == 'msvc'
 	wingetopt = cmake.subproject('wingetopt')
 	deps += wingetopt.dependency('wingetopt')
-	deps += compiler.find_library('ws2_32', required: true)
 endif
 
 if host_machine.system() == 'darwin'
--- a/meson.build
+++ b/meson.build
@@ -78,22 +78,11 @@ src = [	'src/DataProtocol.cpp',
 	'src/SslServer.cpp',
 	'src/Auth.cpp']
 
-if not get_option('rtaudio').disabled()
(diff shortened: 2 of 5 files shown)


dyfer:
Thanks Nils.

Regarding using your fork of RtAudio: Is this meant as a temporary solution or a permanent one? For a temporary one, it's fine I guess, but I really don't think linking to a fork (esp that it's out of sync with upstream repo) is a good approach as a permanent solution.

Is there really no way to set proper cmake flags using the regular RtAudio repo? Here's what [vcpkg does](https://github.com/microsoft/vcpkg/blob/master/ports/rtaudio/portfile.cmake#L12) for the static library with msvc (not 100% sure this is relevant, just pointing it out).

If we absolutely need to keep fetching and building RtAudio, I'd prefer if 1) we fetch the regular repo (and a particular commit instead of HEAD) and 2) apply a patch with any changes.

Why do we want to build it ourselves instead of using one provided with package managers? I know vcpkg is a bit of a pain to install, but I'm really not thrilled by having such a different approach to rtaudio between different build systems we use.

ntonnaett:
> Thanks Nils.
> 
> Regarding using your fork of RtAudio: Is this meant as a temporary solution or a permanent one? For a temporary one, it's fine I guess, but I really don't think linking to a fork (esp that it's out of sync with upstream repo) is a good approach as a permanent solution.

That was meant to be temporarily. I opened a PR for RtAudio: https://github.com/thestk/rtaudio/pull/297
But this might not get accepted. So I will create a patch file for the RtAudio Wrap and set it to mainline again.

> Is there really no way to set proper cmake flags using the regular RtAudio repo? Here's what [vcpkg does](https://github.com/microsoft/vcpkg/blob/master/ports/rtaudio/portfile.cmake#L12) for the static library with msvc (not 100% sure this is relevant, just pointing it out).
> 

When using cmake I have to do things in meson that should actually be handled by cmake. It is even easier to set linking behavior from a user's point of view.

> If we absolutely need to keep fetching and building RtAudio, I'd prefer if 1) we fetch the regular repo (and a particular commit instead of HEAD) and 2) apply a patch with any changes.

Will do.

> Why do we want to build it ourselves instead of using one provided with package managers? I know vcpkg is a bit of a pain to install, but I'm really not thrilled by having such a different approach to rtaudio between different build systems we use.

I assume this is more of a question for GHAs. If you don't enable the rtaudio subproject explicitly it won't use the subproject. I decided to use the meson subproject for RtAudio because it's a pain to handle dependencies on Windows. And then test as many different things with the meson GHA as possible compared to the qmake GHA.

dyfer:
I _think_ this looks fine. 
I can't run it on my system since I don't have Qt libraries in the specific place, but I've also saw in dependencies (using DependenciesGui.exe) that WS2_32.dll is missing some imports: `ext-ms-win-shell32-shellcom-l1-1-0.dll`, `api-ms-win-core-stringansi-l1-1-0.dll` and maybe others. I'm not sure if this is a valid issue, I'm out of my depth on this. My point being: are you sure you've included all required [linker flags](https://www.music.mcgill.ca/~gary/rtaudio/compiling.html) in your rtaudio meson file?

ntonnaett:
The link was my blue print. WS2_32 is a dependency of jacktrip, not rtaudio. If it's linking and not complaining about missing symbols, there shouldn't be any missing linker flags or dependencies.

dyfer:
>  If it's linking and not complaining about missing symbols, there shouldn't be any missing linker flags or dependencies.

Okay, if you say so, I defer to your experience here. We can test once we get a static build I guess, or run windeployqt on it.

dyfer (approved):
Thanks!

PR #302, opened 2021-05-29
Add a French comment in the desktop file

trebmuh:
Hope that helps


ntonnaett (approved):
That's nice. We don't have anything of JackTrip translated yet. Would you be interested in translating more?

trebmuh:
In principle, yes. However, I'm not sure how much time I can spare at this time. How many strings need to be translated (roughly)?

PR #303, opened 2021-05-31
UI and build script changes for rename

psiborg112:
If we are going to be changing everything in the UI to JackTrip rather than QJackTrip, then this PR addresses that. It also changes where settings are stored, and migrates old settings across if it detects them (for those who have been using QJackTrip previously.)

The application bundle script in macos now updates the versioning information in both the bundle and the installer package from the binary itself. It can also take command line options to customize the app name (default is JackTrip) and the bundle identifier (default is org.jacktrip.jacktrip) I'm hoping to do something similar with the windows batch file equivalent eventually. 

Apologies to @dyfer: there's some overlap in the scope of this and your work on #294, but I haven't changed anything with the build scripts or with git hub actions, so it will definitely still be needed.


psiborg112:
@dyfer: your static_files.wxs file should still slot right in unedited. To use the JackTrip_noCert.pkgproj with my altered script you'll just have to name it JackTrip_noCert.pkgproj_template and replace the references to the version, appname, and bundle id with the placeholders (%VERSION%, %BUNDLENAME%, and %BUNDLEID%) that I'm using.

ntonnaett:
Isn't the bundle identifier supposed to be a reversed URL kind of thing?

psiborg112:
> Isn't the bundle identifier supposed to be a reversed URL kind of thing?

Yes it is. I spaced out and wrote it the wrong way around in the comment. (Have corrected it in an edit now.)

dyfer:
Thanks for the heads up @psiborg112 

I think it would be nice to get this into `dev` first so that I can rebase #294 and resolve conflicts.

It would be good to discuss installer, signing etc in the context of GitHub actions. E.g. for the app bundle and installers I ended up not using your scripts but writing out script contents in the yml file directly (based on your scripts) - I'd like for everyone to be on the same page why this is done in a certain way, so that we can choose the best solution moving forward.

macos/JackTrip.app_template/Contents/Info.plist
```diff
@@ -11,15 +11,15 @@
 	<key>CFBundleIconFile</key>
 	<string>jacktrip</string>
 	<key>CFBundleIdentifier</key>
-	<string>org.psi-borg.qjacktrip</string>
+	<string>%BUNDLEID%</string>
 	<key>CFBundleInfoDictionaryVersion</key>
 	<string>6.0</string>
 	<key>CFBundleName</key>
-	<string>QJackTrip</string>
+	<string>%BUNDLENAME%</string>
```
dyfer, line 18: Does the `%BUNDLENAME%` ever get set? I might be missing something, but I can't find it.

psiborg112: Yep, in the assemble_app script (line 24):
`sed -i '' "s/%BUNDLENAME%/$APPNAME/" "$APPNAME.app/Contents/Info.plist"`

psiborg112:
> It would be good to discuss installer, signing etc in the context of GitHub actions. E.g. for the app bundle and installers I ended up not using your scripts but writing out script contents in the yml file directly (based on your scripts) - I'd like for everyone to be on the same page why this is done in a certain way, so that we can choose the best solution moving forward.

I'm not sure we can do signing securely as part of github actions. (This seemed to be the conclusion from past conversations about it, but maybe someone has found out differently now.) Which is fine - someone can just manually sign whenever we need to do a release. And it's all in the script already, just commented out. (Whoever's signing can just do what I do and keep a local copy with their credentials entered into it.)

It would be great if the github actions can just run the script, because then code only needs to be changed in on place. And having a script means that those who want to roll their own at home have access to those capabilities. But it depends how much they need to be customised to work in github. (And your current way of doing things may be fine. Once these are set up, there hopefully won't need to be a huge amount of change.)

macos/JackTrip.app_template/Contents/Info.plist
dyfer: Ah thanks, sorry I missed it!

dyfer:
> It would be great if the github actions can just run the script, because then code only needs to be changed in one place. And having a script means that those who want to roll their own at home have access to those capabilities. But it depends how much they need to be customised to work in github. (And your current way of doing things may be fine. Once these are set up, there hopefully won't need to be a huge amount of change.)

TL;DR I would really like to use the scripts in GHA! However `remove the exit line above` is not necessarily the best way to interact with the script programmatically :)

In the other PR I started making the changes to have the macos script completely configurable using command line switches, but then it was becoming much longer, I was modifying something you've been working with and I didn't know if my changes would be to everyone's liking. Then I looked at the windows script and that one has bunch of paths hardcoded.

I see that you've added application and bundle ID as arguments - that's great.  I think we also need:
- turn macdeployqt step on/off - I think it's not needed for the static build, but please let me know if that's not true
- turn package creation on/off - does the `.pkgproj` file still contain your certificate? (in my PR I'm using a modified pkgproj file, so if we wan't to keep the one with the certificate in, maybe `.pkgproj` this also should be configurable?)

For a good measure, I wouldn't mind if other parameters (codesigning params, apple credentials) could also be set externally, without modifying the script. But we need to test how the signing would work with GHA - i.e. the person who signs the release would download the artifact and sign it, but would they have to put it in the correct place in the build tree? Maybe we should split signing and make it run on the app bundle and the installer, place in a configurable location (or the same directory)?

On windows, the script has a bunch of paths hardcoded, these should be configurable as well (sorry if I'm not understanding and this is already possible, I'm not fluent in `.bat` scripting). You are also copying files and running windeployqt which again is probably also only needed for dynamic builds, so that should also be configurable.

A minor thing - I personally prefer to work in bash, but I understand if we want to keep the windows script as .bat. FWIW the build instructions in the [PDF in this repo](https://github.com/jacktrip/jacktrip/blob/main/WIN10BUILDINSTRUCTIONS.pdf) were showing how to do it in git bash... My other PR extends the build script (bash) so that it runs on windows (mingw) as well. This is mainly for GHA, but I wouldn't mid also documenting this as the recommended build procedure, instead of specifying qmake parameters by hand (current situation on windows, unless you use cmake or meson).

psiborg112:
> TL;DR I would really like to use the scripts in GHA! However `remove the exit line above` is not necessarily the best way to interact with the script programmatically :)

Totally. It's really just a matter of seeing what parameters we want to set and having sane defaults for most users. (Assuming for instance that they don't have packages installed, and just want to build an app bundle, for instance.)

> In the other PR I started making the changes to have the macos script completely configurable using command line switches, but then it was becoming much longer, I was modifying something you've been working with and I didn't know if my changes would be to everyone's liking. Then I looked at the windows script and that one has bunch of paths hardcoded.

I've put those paths into variables now so they're easier to change at least. Could be turned into parameters or we could try to detect where they're installed. Also, if anyone's building with msvc, we'll need an option to configure which compiler runtime libraries are included.

> I see that you've added application and bundle ID as arguments - that's great. I think we also need: 
>     * turn macdeployqt step on/off - I think it's not needed for the static build, but please let me know if that's not true

It's probably not needed, but it might not hurt to run it either. (My assumption would be that it looks at the binary to see what libraries need to be packaged. And if there aren't any needed then it won't add them to the bundle. But I haven't built using static qt in a long time - didn't see the point of going to the extra effort of building it once I started using the app bundle and installer. So worth testing for someone who does have a static qt build handy.)

>     * turn package creation on/off - does the `.pkgproj` file still contain your certificate? (in my PR I'm using a modified pkgproj file, so if we want to keep the one with the certificate in, maybe `.pkgproj` file also should be configurable?)

It should not contain my certificate. We should just use the one without my certificate. But this could relate to the question of how we handle signing in general.

> A minor thing - I personally prefer to work in bash, but I understand if we want to keep the windows script as .bat. FWIW the build instructions in the [PDF in this repo](https://github.com/jacktrip/jacktrip/blob/main/WIN10BUILDINSTRUCTIONS.pdf) were showing how to do it in git bash... My other PR extends the build script (bash) so that it runs on windows (mingw) as well. This is mainly for GHA, but I wouldn't mid also documenting this as the recommended build procedure, instead of specifying qmake parameters by hand (current situation on windows, unless you use cmake or meson).

I also prefer to work in bash, but I like having a script that can run natively with as few dependencies as possible.

dyfer:
I see a few options for script configurability. On macOS maybe we should have 3 scripts: create app bundle, create installer, sign. First two would run in GHA, the last one would be run by the person signing. (All three could be run locally if needed of course)

On Windows, maybe we also three scripts: set paths and store them in env vars, create app bundle (that would copy extra files if needed) and create the installer. That way we wouldn't run the first scrip in GHA because we'd set the paths according to GHA config.

What do you think about the scripts?

Re: mac/windeployqt
On one had I don't mind running it just in case, but in #294 I observed the following:
- I need to build qttools to get the deploy tool with the static qt; it takes 40-50% more time (though maybe that's not a big issue, since we do cache that
- on Windows I had a weird issue where windeployqt said that jacktrip is not a qt app... I wonder if that was because of the static build?

In any case, I thought that we could maybe not run it if we don't need it... i.e. run it only for dynamic builds. But I'm open to other ideas and to figuring out what the real issue with windeployqt is.

So... how should we proceed? I see at least two options:
- merge this PR
  2. update and merge #294 not using scripts yet
  3. update scripts (new PR) and change GHA to use them
- update scripts first (in this PR or another) so they can be used in GHA
  2. update #294 to use those scripts and merge it

psiborg112:
I'm inclined to go with the first option, since it gets us there faster and we can keep tweaking things as we go.

Thanks for the info about static qt. I was just hoping we could run it anyway and have it sort out the details to save on complexity within the script, but seems like that's not a great option. That said, I did just come up with a way to test whether a jacktrip binary is built using static Qt on Mac, so we can make the script auto-detect whether we need to run qtdeploy. (That's my one good idea for the day.) Will investigate a similar test on windows.

I think we could probably still keep everything in the one script, just with command line options. Just build the app bundle by default, and build the installer if a switch is set. It might make more sense to have the signing as a separate script though, but that will depend if we can sign an application already in an installer package. (Will investigate.)

psiborg112:
Also meant to add that we probably don't need the first script in windows. We can just check if the environment variables are already set and not modify them if they are.

dyfer (approved):
Thanks, let get this into dev!

psiborg112:
Mac bundle script is fully configurable now.

psiborg112:
@dyfer: Have just uploaded an updated script to the win_installer_script branch. It only sets path variables if they're not already set, and it checks to see if static qt has been used. Just to check though, are we building using mingw or msvc? The script assumes that the mingw tools for qt will at least be available. If they are, then the static check should work regardless of what we're actually building with. That said, if we're building with msvc then we need to change some of the dlls we're copying into the package. (Different compiler runtime libraries.)

dyfer:
> Mac bundle script is fully configurable now. Should be able to use it in the github scripts with just the -i flag set.

Great, thanks so much!

> @dyfer: Have just uploaded an updated script to the win_installer_script branch. It only sets path variables if they're not already set, and it checks to see if static qt has been used. 

Sounds good. Should I cherry pick that into #294 or are you planning to make a PR with it?

> Just to check though, are we building using mingw or msvc? The script assumes that the mingw tools for qt will at least be available. If they are, then the static check should work regardless of what we're actually building with. That said, if we're building with msvc then we need to change some of the dlls we're copying into the package. (Different compiler runtime libraries.)

We currently building with mingw and everything needed seems to be in PATH. 
I plan to add a switch for msvc so eventually it would be nice to add that configurability to the script. Not necessarily for this release though.

psiborg112:
I can probably just make a PR for it now. That's probably easiest. Let me know if you find any issues with it - happy to help with adjusting things.

dyfer:
Sounds good, feel free to open a PR.

For win script, I see at least one thing that I'd need for #294: I need to define a different list of files for the static installer. Right now I just created the `files_static.wxs`, though I tried the installer from GHA and it doesn't seem to work (quits with no message) so I probably messed something up. Maybe let's continue this conversation after you open the PR.

Another comment to both scripts I had (and forgot to mention) is that I'm not sure if we should be doing a check whether JT is build with a GUI. While not a common scenario, I can imagine someone wanting to install a non-gui version... Although on a mac that would be kind of weird if it's an app bundle. Still, I'm not sure if we should block it.

psiborg112:
Already merged, but adding the files_static thing will be an easy edit. I'm not sure what the point of the scripts would be with the non-gui version though, especially on mac. The main purpose of it is to build an app bundle and the non-gui version just wouldn't work in that scenario. I'm not sure exactly how it would operate, but best case scenario is that clicking the app icon would launch a terminal window showing the help screen that would quit out immediately. (I think the window would hang around showing '[process completed]' at the bottom, at least.) Windows would fare better at least, but the main advantage of the installer is that it sets the user up with start menu items, which once again would be of very limited use here. It's such a different use case that I'm not sure an installer makes sense. (And absolute worst case scenario would be someone without technical knowledge getting a hold of a non-gui installer and then not knowing what's gone wrong.)

dyfer:
> (And absolute worst case scenario would be someone without technical knowledge getting a hold of a non-gui installer and then not knowing what's gone wrong.)

Yeah, that's a serious problem. I haven't considered that as much as I should.

My thinking was mainly to be able to distribute dynamic lib non-gui builds - if we wanted to. On a mac an installer would make sure it's available in the terminal (but the app icon would be useless, which is of course suboptimal). On windows the action on click-open (terminal closing immediately) would definitely make it for a possible confusion; on a mac the terminal window at least stays visible after the command is done.

So yeah, maybe not worth the trouble.

Thanks again for the `files_static` addition! I'll work on updating #294 shortly.

PR #306, opened 2021-06-01
Rtaudio subproject fixes

ntonnaett:
Sorry @dyfer! There were some flags missing. I forgot to pass the defines to the compiler. But no linking errors occured.

### DIFF
--- a/.github/workflows/meson.yml
+++ b/.github/workflows/meson.yml
@@ -91,10 +91,10 @@ jobs:
       - name: build on Windows
         if: runner.os == 'Windows'
         run: |
-          meson --buildtype release builddir
+          meson --buildtype release -Drtaudio=enabled builddir
           cd builddir
-          meson configure -Ddefault_library=static -Drtaudio=enabled
-          meson compile
+          meson configure -Ddefault_library=both -Drtaudio:default_library=static -Drtaudio:wasapi=true -Dwingetopt:default_library=static
+          meson compile -v
       - name: upload artifacts
         uses: actions/upload-artifact@v2
         with:
--- a/.github/workflows/meson.yml
+++ b/.github/workflows/meson.yml
@@ -94,7 +94,7 @@ jobs:
           meson --buildtype release -Drtaudio=enabled builddir
           cd builddir
           meson configure -Ddefault_library=both -Drtaudio:default_library=static -Drtaudio:wasapi=true -Dwingetopt:default_library=static
-          meson compile -v
+          meson compile
       - name: upload artifacts
         uses: actions/upload-artifact@v2
         with:
--- a/meson.build
+++ b/meson.build
@@ -90,7 +90,13 @@ if host_machine.system() == 'windows'
 endif
 
 if compiler.get_id() == 'msvc'
-	wingetopt = cmake.subproject('wingetopt')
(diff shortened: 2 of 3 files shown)


dyfer:
Have you tried running this build on Windows? DependenciesGui shows some missing dependencies (inside `WS2_32.dll` and `DSOUND.dll`) and I can't run it because of that, despite Qt actually being found.

ntonnaett:
As I said. I don't have Windows. Both libraries are not part of jacktrip. So I don't understand what DependenciesGui is telling you. Could you post the output? Did you build it yourself?

ntonnaett:
Maybe you don't have DirectX installed?!

dyfer:
> As I said. I don't have Windows. Both libraries are not part of jacktrip. So I don't understand what DependenciesGui is telling you. 

I really can't be much help, aside from reporting what does and does not work, sorry.

> Could you post the output? 

I'll try later.

> Did you build it yourself?

No, I used the build from GHA.

> Maybe you don't have DirectX installed?!

Is that even needed in Windows 10? Also, dev JT build (with RtAudio) works fine.

BTW it doesn't seem like you're building with DirectSound, but one of the missing dependencies suggests that maybe you do? Would be worth double checking parsing options etc.

It does seem like you have all linker flags in place. However, I had a problem with linking on windows in qmake - these windows flags needed to be [specified](https://github.com/jacktrip/jacktrip/blob/rtaudio_subproject_fixes/jacktrip.pro#L143) _after_ `-lrtaudio`. This is with MinGW so maybe not relevant, but I thought I'd point it out. And also - I had to link JackTrip with these libraries, rtaudio itself was not enough.

I wonder if you could run JT in wine to test these things... though probably that poses its own set of problems.

ntonnaett:
I tried with wine and it's only complaining about missing Qt5 in the beginning. Copying from mingw doesn't work because jacktrip is compiled with msvc. Dev works fine because it has no backends built.

Right now I try to get the cmake flags for wingetopt right. Maybe it's easier to include those files in jacktrip at some point in the future.

dyfer:
> I tried with wine and it's only complaining about missing Qt5 in the beginning. Copying from mingw doesn't work because jacktrip is compiled with msvc. Dev works fine because it has no backends built.

Ah! Maybe it's a mistake on my end? The binary from this build picked up Qt in the MinGW folder. Trying now

ntonnaett:
I am cleaning up right now too. Maybe you want to wait. I found some things with wingetopt that weren't right.

ntonnaett:
Ready!

dyfer:
I'm so SORRY Nils I have sent you on a wild goose chase.

After Installing Qt 5.15 for msvc and adding that to PATH I can start this build of JT. Interestingly, DependenciesGui still complains about WS2_32 and DSOUND dlls, but the applications starts properly.

Just to re-iterate: I can start both the build I've downloaded previously, as well as the current build.

Sorry again.

dyfer (approved):
Thanks!

are you leaving `compile -v` in for now?

ntonnaett:
The current build has less warnings. So it was beneficial nevertheless.

I remove the -v. I don't think that we need this level of information for static qt.

PR #298, opened 2021-05-28
Documentation for 1.4.0

ntonnaett:
That's just a start. We might want to add to this PR or have several PRs.

PR #310, opened 2021-06-02
Dev plc for 1.4.0

cchafe:
ready for review!
-- new bufstrategy 3
-- implements an experimental receive buffering algorithm (sort of LIFO rather than FIFO) using a pool rather than ring
-- uses Burg AR prediction to fill in late packets
should be close to constant lag and runs fine at -q2
pool size adjusts dynamically adapting to std dev (jitter ms)
-V gives the running statistics
for example, call a server or client with --udprt --bufstrategy 3 -q2 -V
typical load seen in "top -H" is 35-40% per channel under the jacktrip thread (which is where Burg runs)
-q <some_large_number> gives a delayed output as needed
so far, only stereo testing, and only hub mode
survives pulling out and then replacing the ethernet cable


cchafe:
Todo's from yesterday's mtg -- clang-tidy and make the stdDev vs. poolsize relationship also depend on FPP

ntonnaett:
You mean clang-format. The command I posted yesterday wasn't totally right. You have to add -i to apply the change to the file:
`clang-format -i file.cpp file.h`

cchafe:
Right, thanks! Also to move implementations from .h to .cpp

SynthiaPayne:
do I need a pull request to say thank you to you all?
ha, it's inspiring to witness this work activity, thank you for 
everything. I would like to create a summary of major updates for 1.4, 
but I've lost track of it all. Is there one already I could review?

thanks again.
synthia

ntonnaett:
There is one in #298 `docs/About/CHANGELOG.md`

SynthiaPayne:
Thank you, Nils!

synth

src/PoolBuffer.h
```diff
@@ -0,0 +1,163 @@
...
+
+#include "AudioInterface.h"
+#include "RingBuffer.h"
+using std::vector;
+#include <math.h>
+
+class BurgAlgorithm
+{
+   public:
+    BurgAlgorithm() {}
```
ntonnaett, line 57: This looks like the constructor can be removed. This can potentially lead to faster code (Rule of zero).

cchafe:
Got it, thanks. All the mods are done. Undrafting the draft...

ntonnaett (approved):
I didn't test it. But I think others did. So I would be okay to squash merge this.

PR #311, opened 2021-06-02
Have wix use the right file list for static win build

psiborg112:
Needed for #294. (Doesn't include the actual static file list, just handles filename selection in the script.) Also give the windows script a little more flexibility in finding Qt and WiX. (Not tied to a specific version number.) And have the macos script search in some known locations for macdeployqt if it's not in the path.


dyfer (approved):
Wow! This is great! Thanks!

I haven't tested it but I think I will in #294 and resolve any issues then.

PR #315, opened 2021-06-04
gha mkdocs: run only on the main repo

dyfer:
I think this should work :) - I don't think we want to make it run on forks (i.e. I don't have GH pages activated on my fork and I wouldn't want a copy of the main JT pages there anyway).
If we wanted, we could make it configurable per repository (fork) but I don't think that's needed - let me know if you think otherwise.


ntonnaett (approved):
I will trust you with that :)

dyfer:
> I will trust you with that :)

Thanks! I'll fix it if it breaks anything :)

PR #317, opened 2021-06-04
Flatpak and appstream

ntonnaett:
Flatpak runs and is able to connect to our server. The flatpak manifest uses dev branch right now. That should be changed for a release. So what's missing is getting it into the flathub repository.

```bash
cd linux/flatpak/
flatpak-builder --user --install build-dir org.jacktrip.JackTrip.yml
flatpak run org.flatpak.Hello
```

https://docs.flatpak.org/en/latest/first-build.html


ntonnaett:
So looking at [this](https://github.com/flathub/flathub/wiki/App-Submission) it seams like I just open a PR when JackTrip 1.4.0 is released. Right now I use a git submodule called shared-modules that gives us a jack2 flatpak module. That will be unnecessary in the future when pipewire is updated in the flatpak platform (base layer where you put your app on top).

ntonnaett:
So we will have something like what SonoBus has: https://flathub.org/apps/details/net.sonobus.SonoBus

ntonnaett:
I should definitely mention and thank @tinywrkb for his help.

dyfer:
Thanks! Questions:

- would we run `flatpak-builder` in GHA ideally?
- would we want to serve the file ourselves or publish to flatpak repository?
  - if the latter, how to we publish the app exactly? Who hosts it?

linux/org.jacktrip.JackTrip.metainfo.xml
```diff
@@ -0,0 +1,37 @@
...
+
+  <url type="homepage">https://jacktrip.github.io/jacktrip/</url>
+  <project_group>JackTrip</project_group>
+
+  <provides>
+    <binary>jacktrip</binary>
+  </provides>
+
+  <releases>
+    <release version="1.4.0rc1" date="2021-06-17">
```
dyfer, line 34: We've recently removed all hardcoded version identifiers (e.g. [here](https://github.com/jacktrip/jacktrip/pull/303/files#diff-7aa732776e190add1ac6737c1475e0beef9a70ae0887646edd137b9b08bcbd38R22)), could we somehow avoid hardcoding a version in this file?

ntonnaett:
>     * would we run `flatpak-builder` in GHA ideally?

Probably not.

>     * would we want to serve the file ourselves or publish to flatpak repository?
>       
>       * if the latter, how to we publish the app exactly? Who hosts it?

I open a PR for flathub. Distribution's have their own flatpak repositories but flathub is the common one. I'm not quite sure if they produce nightly build for the dev versions. But I will figure this out.

The flatpak file in our repository will be more like the template for the real one in flathub. Or if you want to build your local flatpak.

dyfer:
Sounds good, so maybe we should indicate that what we have here is a template? Also, why do we need that submodule? 

Sorry for stupid question, I have no idea about flatpak.

ntonnaett:
The submodule is for the jack dependency. It's just a repository of common flatpak modules. This will eventually be removed when we can rely solely on the platform.

We shouldn't call it a template because it's fully functional.

PR #318, opened 2021-06-04
Automatic upload of release artifacts

dyfer:
In preparation for the release:

- added automatic uploads of artifacts to the GH release page
  - I'm not uploading Linux binaries due to issues with the static Linux build
- tagged builds have tag included in the filename
- single files have to be double-zipped when downloaded straight from actions (so that they show up as a regular zip on the release page)
- changed GHA to run on main + dev, PR and on schedule to refresh cache (on schedule will become active once merged into main)
- removed skip duplicates

_After this PR GHA won't run on all pushes, just PRs, after merges to dev/main and eventually "on demand"_ (after merging this to main, i.e. after the 1.4.0 release)

You can see artifacts in actions from a tagged commit form this branch here: https://github.com/dyfer/jacktrip/actions/runs/907802432

Here's a test release: https://github.com/dyfer/jacktrip/releases/tag/test.v1.3.9

Release can be created using GH web UI, or by pushing a new tag using git itself.

### DIFF
--- a/.github/workflows/jacktrip.yml
+++ b/.github/workflows/jacktrip.yml
@@ -250,3 +250,29 @@ jobs:
         with:
           name: JackTrip-${{ steps.set-version.outputs.version }}-${{ matrix.name }}-installer
           path: ${{ env.BUILD_PATH }}/${{ matrix.installer-path }}
+  # release - list of files uploaded to GH release is specified in the *upload* step
+  deploy_gh:
+    if: startsWith(github.ref, 'refs/tags/') # run on tagged commits
+    needs: build
+    runs-on: ubuntu-18.04
+    name: 'deploy release'
+    env:
+      DOWNLOAD_PATH: ${{ github.workspace }}/Install
+    steps:
+      - name: download artifacts
+        uses: actions/download-artifact@v2
+        with:
+          path: ${{ env.DOWNLOAD_PATH }} # no "name" paramter - download all artifacts
+      - name: upload to the release page
+        uses: softprops/action-gh-release@v1
+        with:
+          files: |
+            ${{ env.DOWNLOAD_PATH }}/JackTrip*Linux-x64/*
+            ${{ env.DOWNLOAD_PATH }}/JackTrip*macOS-x64/*
+            ${{ env.DOWNLOAD_PATH }}/JackTrip*macOS-x64-application/*
+            ${{ env.DOWNLOAD_PATH }}/JackTrip*macOS-x64-installer/*
+            ${{ env.DOWNLOAD_PATH }}/JackTrip*Windows-x64/*
+            ${{ env.DOWNLOAD_PATH }}/JackTrip*Windows-x64-installer/*
+          draft: true
+        env:
+          GITHUB_TOKEN: ${{ secrets.GITHUB_TOKEN }}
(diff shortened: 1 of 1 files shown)


ntonnaett:
Why do you want to remove skip-duplicates?

dyfer:
> Why do you want to remove skip-duplicates?

- I strongly believe we want to run GHA on primary branches, always.
  - runs in PR might be from the same codebase, but they don't have access to the same environment (github secrets, cache storage), at least when they come from a fork
  - we want dev artifacts (so always run on dev - I had to add [another check](https://github.com/jacktrip/jacktrip/pull/318/files#diff-400b21146a0aa3667e1c8f1d95c364ac50306b03359d195bff0e90725ecaabf8L19) to make that work, which is another piece of logic to amend skip-duplicates)
  - after this PR GHA won't run automatically on every push and won run on doc updates, so I really don't see a reason to skip anything
  - IIUC if we kept "skip-duplicates" it would make it to skip automated weekly builds (they run on main, so it would be the same code throughout the release) - we need that to "refresh" cache and avoid unnecessary long Qt builds

Other reasons are more personal:
- IMO it doesn't have any measurable benefit, while at the same time:
- it creates confusing behavior (some CI runs don't actually create artifacts etc) - especially for those not closely following our gha practices, future maintainers of the project, occasional collaborators)
- list of job within a run has an out-of-place step ("skip-artifacts" has no build-related function), which increases visual noise

ntonnaett:
Where do it avoid building duplicates? I see that it gets triggered by push to dev or main and PRs to any branch. It's extremely annoying to wait for the CI to finish. Especially the Windows build needs a lot of time. So getting it down to around half of the time by just not doing the same thing a second time, is really appreciated.

dyfer:
> Where do it avoid building duplicates? I see that it gets triggered by push to dev or main and PRs to any branch. 

Here: https://github.com/jacktrip/jacktrip/actions/runs/901209633
This is a typical merge before I made the change to always run on dev recently - the job is not re-run after merge, which results in no artifacts being associated with the dev branch.

> It's extremely annoying to wait for the CI to finish. Especially the Windows build needs a lot of time. So getting it down to around half of the time by just not doing the same thing a second time, is really appreciated.

GHA runs jobs run in parallel. Where is the extra waiting coming from?

I also really wanted to stress that:
- runs in PR might be from the same codebase, but they don't have access to the same environment
- after this PR GHA won't run automatically on every push and won't run on doc updates

ntonnaett:
I think you misread my question. I was referring to:
>after this PR GHA won't run automatically on every push and won't run on doc updates, so I really don't see a reason to skip anything

I answered the question myself now. If there's a PR it usually is a PR to main or dev. It triggers on the PR and a second time on the push. So the answer is that the push only occurs later when it gets merged to main or dev (with this change now). So the only left duplicates are when PRs get merged. And if they come from foreign repos, they aren't really duplicates because they are doing different things.

ntonnaett (approved):
Thank you!

dyfer:
> I think you misread my question. I was referring to:
> 
> > after this PR GHA won't run automatically on every push and won't run on doc updates, so I really don't see a reason to skip anything

Ah gotcha, sorry!

Issue #276, opened 2021-05-21
Use of compiler reserved names


ntonnaett:
Global names that begin with an underscore (followed by a capital letter) and names with double underscore are reserved for the compiler/preprocessor/implementer. We should change our preprocessor variables accordingly: __LINUX__, __WIN32__, __NO_GUI__, etc.

There are predefined variables for checking the platform.

PR #319, opened 2021-06-06
Replace cryptic error message on linux if no X or Wayland display

psiborg112:
If no x or wayland display is set, show the user a more informative error message than the qt plugin one that otherwise appears.

Also, replace \_\_NO_GUI\_\_ with NO_GUI (addresses part of #276)


ntonnaett (approved):
Great! Love it.

PR #320, opened 2021-06-07
Fix static Linux build in GitHub Actions

dyfer:
This should fix the Linux static build... tested on Ubuntu only.

~~It turned out, that in order to build static Qt that runs in X11, one needs to bundle the qt `xcb` platform plugin. This has been [disabled in Qt 5.15](https://doc.qt.io/qt-5/linux-requirements.html), but is still [possible in Qt 5.12](https://doc.qt.io/qt-5.12/linux-requirements.html).~~

It turned out that bundling xcb platform plugin was not necessary. Qt 5.15 seems to be incompatible with the distribution I was testing with (Ubuntu 20.04, running X11). 

Compatibility with Wayland is still TBD.

In this PR
- I made the qt version configurable in the build matrix
- I set Linux and macOS to use Qt 5.12
  - ~~on Linux, I've used build flags specifically for 5.12, namely `-qt-xcb`, which bundles the xcb plugin with the library; this is the main qt "fix" in this PR~~ (aside from bundling `libjpeg` and the like)
- on all platforms I removed `-skip qtquick3d -skip qtlottie -skip qtquicktimeline` when building static qt since they are not present in 5.12 and caused errors; the increased build time should be negligible
- on macOS I used the earlier Qt version so we support slightly older operating systems (macOS 10.12 I believe) - why not? I can revert that if desired
- Linux build now installs only the needed libraries when building qt, instead of installing qt from apt
- I kept Windows at Qt 5.15 since AFAIR only that version (or 5.14 maybe) fixes the issue with fractional display scaling

**EDIT: here's the link to the [Linux static build](https://github.com/jacktrip/jacktrip/suites/2934741456/artifacts/65930498)**

### DIFF
--- a/.github/workflows/jacktrip.yml
+++ b/.github/workflows/jacktrip.yml
@@ -30,7 +30,7 @@ jobs:
         include:
           - name: Linux-x64
             runs-on: ubuntu-18.04
-            static: true
+            static-qt-version: 5.12.11 # older Qt version so that we can compile in xcb platform plugin
             jacktrip-path: jacktrip
             binary-path: binary
 
@@ -48,7 +48,7 @@ jobs:
           - name: macOS-x64
             runs-on: macos-10.15
             rtaudio: true
-            static: true
+            static-qt-version: 5.12.11 # if set, it will enable static Qt build
             weakjack: true
             jacktrip-path: jacktrip # relative to build path
             binary-path: binary # directory relative to build path; triggers artifact upload for jacktrip binary
@@ -64,7 +64,7 @@ jobs:
             # suffix: installer
             runs-on: windows-2019
             rtaudio: true
-            static: true
+            static-qt-version: 5.15.2
             weakjack: true
             jacktrip-path: release/jacktrip.exe
             binary-path: binary # directory relative to build path; triggers artifact upload for jacktrip binary
@@ -72,13 +72,12 @@ jobs:
 
     env:
       BUILD_PATH: ${{ github.workspace }}/builddir
-      QT_VERSION: '5.15.2'
+      QT_VERSION: '5.15.2' # for dynamic qt installer using install-qt-action
       DEVELOPER_DIR: '/Applications/Xcode_11.7.app/Contents/Developer' # specify XCode version on macOS
-      QT_SRC_URL: 'https://download.qt.io/official_releases/qt/5.15/5.15.2/single/qt-everywhere-src-5.15.2.tar.xz'
       QT_SRC_PATH: ${{ github.workspace }}/../qt-static-src
       QT_STATIC_BUILD_PATH: ${{ github.workspace }}/../qt-static
-      QT_STATIC_CACHE_KEY: '5.15.2_v02' # update this to force rebuilding static Qt
-      QT_STATIC_OPTIONS: '-static -release -optimize-size -no-pch -nomake tools -nomake tests -nomake examples -opensource -confirm-license -skip webengine -skip qt3d -skip qtactiveqt -skip qtandroidextras -skip qtcharts -skip qtdatavis3d -skip qtdeclarative -skip qtdoc -skip qtgamepad -skip qtgraphicaleffects -skip qtimageformats -skip qtlocation -skip qtlottie -skip qtmultimedia -skip qtpurchasing -skip qtquick3d -skip qtpurchasing -skip qtquickcontrols -skip qtquickcontrols2 -skip qtquicktimeline -skip qtremoteobjects -skip qtscript -skip qtsensors -skip qtserialbus -skip qtserialport -skip qtspeech -skip qtsvg -skip qttools -skip qtvirtualkeyboard -skip qtwayland -skip qtwebchannel -skip qtwebengine -skip qtwebglplugin -skip qtwebsockets -skip qtwebview -skip qtxmlpatterns' # common for all platforms
+      QT_STATIC_VERSION: 'v02' # update this to force rebuilding static Qt
+      QT_STATIC_OPTIONS: '-static -release -optimize-size -no-pch -nomake tools -nomake tests -nomake examples -opensource -confirm-license -skip webengine -skip qt3d -skip qtactiveqt -skip qtandroidextras -skip qtcharts -skip qtdatavis3d -skip qtdeclarative -skip qtdoc -skip qtgamepad -skip qtgraphicaleffects -skip qtimageformats -skip qtlocation -skip qtmultimedia -skip qtpurchasing -skip qtpurchasing -skip qtquickcontrols -skip qtquickcontrols2 -skip qtremoteobjects -skip qtscript -skip qtsensors -skip qtserialbus -skip qtserialport -skip qtspeech -skip qtsvg -skip qttools -skip qtvirtualkeyboard -skip qtwayland -skip qtwebchannel -skip qtwebengine -skip qtwebglplugin -skip qtwebsockets -skip qtwebview -skip qtxmlpatterns' # common for all platforms
       
     steps:
       - uses: actions/checkout@v2
@@ -111,7 +110,7 @@ jobs:
           curl -L https://github.com/jackaudio/jack2-releases/releases/download/v1.9.16/jack2-macOS-v1.9.16.tar.gz -o jack2.tar.gz
           tar -xf jack2.tar.gz
           sudo installer -pkg jack2-osx-1.9.16.pkg -target /
-          if [[ "${{ matrix.static }}" != true ]]; then 
+          if [[ -z "${{ matrix.static-qt-version }}" ]]; then 
             brew install qt5
             brew link qt5 --force
           fi
@@ -135,37 +134,40 @@ jobs:
           fi
       - name: cache Qt
         id: cache-qt
-        if: runner.os == 'Windows' && matrix.static != true
+        if: runner.os == 'Windows' && '!matrix.static-qt-version'
         uses: actions/cache@v1
         with:
           path: ../Qt
           key: ${{ runner.os }}-QtCache-${{ env.QT_VERSION }}
       - name: install Qt
         uses: jurplel/install-qt-action@v2
-        if: runner.os == 'Windows' && matrix.static != true
+        if: runner.os == 'Windows' && '!matrix.static-qt-version'
         with:
           version: ${{ env.QT_VERSION }}
           arch: 'win64_mingw81'
           cached: ${{ steps.cache-qt.outputs.cache-hit }}
       - name: cache static Qt
         id: cache-static-qt
-        if: matrix.static == true
+        if: matrix.static-qt-version
         uses: actions/cache@v1
         with:
           path: ${{ env.QT_STATIC_BUILD_PATH }}
-          key: ${{ runner.os }}-QtStaticCache-${{ env.QT_STATIC_CACHE_KEY }}
+          key: ${{ runner.os }}-QtStaticCache-${{ matrix.static-qt-version }}_${{ env.QT_STATIC_VERSION }}
       - name: download Qt sources
-        if: matrix.static == true && steps.cache-static-qt.outputs.cache-hit != 'true'
+        if: matrix.static-qt-version && steps.cache-static-qt.outputs.cache-hit != 'true'
         shell: bash
         run: |
-          echo "Downloading Qt source"
+          QT_FULL_VERSION=${{ matrix.static-qt-version }}
+          QT_MAJOR_MINOR_VERSION=`echo $QT_FULL_VERSION | cut -d '.' -f1-2`
+          QT_SRC_URL="https://download.qt.io/official_releases/qt/$QT_MAJOR_MINOR_VERSION/$QT_FULL_VERSION/single/qt-everywhere-src-$QT_FULL_VERSION.tar.xz"
+          echo "Downloading Qt source from $QT_SRC_URL"
           curl -L $QT_SRC_URL -o qt.tar.xz
           echo "Unpacking Qt source"
           tar -xf qt.tar.xz
           # move qt to a known location
           mv qt-everywhere* $QT_SRC_PATH
       - name: build static Qt on Linux / macOS
-        if: matrix.static == true && steps.cache-static-qt.outputs.cache-hit != 'true' && runner.os != 'Windows'
+        if: matrix.static-qt-version && steps.cache-static-qt.outputs.cache-hit != 'true' && runner.os != 'Windows'
         run: |
           mkdir $QT_STATIC_BUILD_PATH
           # configure
@@ -177,7 +179,7 @@ jobs:
           echo "installing Qt"
           make install
       - name: build static Qt on Windows
-        if: matrix.static == true && steps.cache-static-qt.outputs.cache-hit != 'true' && runner.os == 'Windows'
+        if: matrix.static-qt-version && steps.cache-static-qt.outputs.cache-hit != 'true' && runner.os == 'Windows'
         shell: cmd
         run: |
           mkdir %QT_STATIC_BUILD_PATH%
@@ -189,7 +191,7 @@ jobs:
       - name: set Qt environment variables
         shell: bash
         run: |
-          if [[ "${{ matrix.static }}" == true ]]; then 
+          if [[ -n "${{ matrix.static-qt-version }}" ]]; then 
             echo "$QT_STATIC_BUILD_PATH/bin" >> $GITHUB_PATH
           elif [[ "${{ runner.os }}" == "macOS" ]]; then
             echo "`brew --prefix qt5`/bin" >> $GITHUB_PATH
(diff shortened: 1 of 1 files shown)


dyfer:
@ntonnaett @cchafe could you test whether the Linux static build works on Fedora?

ntonnaett:
Why is this a zip in a zip?

ntonnaett:
These are all shared linked libraries:
```bash
$ ldd jacktrip
	linux-vdso.so.1 (0x00007ffef611e000)
	libjack.so.0 => /usr/lib64/pipewire-0.3/jack/libjack.so.0 (0x00007f7b1fb1c000)
	libm.so.6 => /lib64/libm.so.6 (0x00007f7b1f9d8000)
	libpthread.so.0 => /lib64/libpthread.so.0 (0x00007f7b1f9b7000)
	libstdc++.so.6 => /lib64/libstdc++.so.6 (0x00007f7b1f798000)
	libgcc_s.so.1 => /lib64/libgcc_s.so.1 (0x00007f7b1f77d000)
	libc.so.6 => /lib64/libc.so.6 (0x00007f7b1f5ae000)
	libfontconfig.so.1 => /lib64/libfontconfig.so.1 (0x00007f7b1f55e000)
	libfreetype.so.6 => /lib64/libfreetype.so.6 (0x00007f7b1f49a000)
	libxcb-glx.so.0 => /lib64/libxcb-glx.so.0 (0x00007f7b1f47d000)
	libX11-xcb.so.1 => /lib64/libX11-xcb.so.1 (0x00007f7b1f478000)
	libxcb.so.1 => /lib64/libxcb.so.1 (0x00007f7b1f44d000)
	libXrender.so.1 => /lib64/libXrender.so.1 (0x00007f7b1f440000)
	libX11.so.6 => /lib64/libX11.so.6 (0x00007f7b1f2f7000)
	libSM.so.6 => /lib64/libSM.so.6 (0x00007f7b1f2ec000)
	libICE.so.6 => /lib64/libICE.so.6 (0x00007f7b1f2ce000)
	libxkbcommon-x11.so.0 => /lib64/libxkbcommon-x11.so.0 (0x00007f7b1f2c3000)
	libxkbcommon.so.0 => /lib64/libxkbcommon.so.0 (0x00007f7b1f27a000)
	libdl.so.2 => /lib64/libdl.so.2 (0x00007f7b1f273000)
	libglib-2.0.so.0 => /lib64/libglib-2.0.so.0 (0x00007f7b1f137000)
	libGL.so.1 => /lib64/libGL.so.1 (0x00007f7b1f0b0000)
	/lib64/ld-linux-x86-64.so.2 (0x00007f7b208c6000)
	libpipewire-0.3.so.0 => /lib64/libpipewire-0.3.so.0 (0x00007f7b1eff2000)
	libxml2.so.2 => /lib64/libxml2.so.2 (0x00007f7b1ee69000)
	libbz2.so.1 => /lib64/libbz2.so.1 (0x00007f7b1ee56000)
	libpng16.so.16 => /lib64/libpng16.so.16 (0x00007f7b1ee1d000)
	libz.so.1 => /lib64/libz.so.1 (0x00007f7b1ee01000)
	libharfbuzz.so.0 => /lib64/libharfbuzz.so.0 (0x00007f7b1ed31000)
	libbrotlidec.so.1 => /lib64/libbrotlidec.so.1 (0x00007f7b1ed23000)
	libXau.so.6 => /lib64/libXau.so.6 (0x00007f7b1ed1d000)
	libuuid.so.1 => /lib64/libuuid.so.1 (0x00007f7b1ed14000)
	libxcb-xkb.so.1 => /lib64/libxcb-xkb.so.1 (0x00007f7b1ecf4000)
	libpcre.so.1 => /lib64/libpcre.so.1 (0x00007f7b1ec7c000)
	libGLX.so.0 => /lib64/libGLX.so.0 (0x00007f7b1ec4a000)
	libXext.so.6 => /lib64/libXext.so.6 (0x00007f7b1ec35000)
	libGLdispatch.so.0 => /lib64/libGLdispatch.so.0 (0x00007f7b1eb7d000)
	liblzma.so.5 => /lib64/liblzma.so.5 (0x00007f7b1eb51000)
	libgraphite2.so.3 => /lib64/libgraphite2.so.3 (0x00007f7b1eb2e000)
	libbrotlicommon.so.1 => /lib64/libbrotlicommon.so.1 (0x00007f7b1eb0b000)
```

It looks like it's using X11 instead of wayland. That looks bad on my display. Is there something like -qt-wayland?

dyfer:
> It looks like it's using X11 instead of wayland. That looks bad on my display. Is there something like -qt-wayland?

Ooops, I skipped qtwayland module. Will try to fix.

Does it run OK otherwise?

> Why is this a zip in a zip?

That came in #318. Unfortunately due to limitations of GHA what we see in the artifact download is automatically zipped, even if the file was already zipped. But when uploading to GH release, that "exterior" zip is not present. If I had not zipped it first, the release artifacts for the binaries would be `jacktrip` instlead of `JackTrip-<version>-<platform>-binary.zip`. I could try to circumvent that by moving the zip creation to the release deployment step, if that bothers you and others a lot.

From #318:
> single files have to be double-zipped when downloaded straight from actions (so that they show up as a regular zip on the release page)

ntonnaett:
> 
> Does it run OK otherwise?

Didn't try more than to start it.

> > Why is this a zip in a zip?
> 
> That came in #318. Unfortunately due to limitations of GHA what we see in the artifact download is automatically zipped, even if the file was already zipped. But when uploading to GH release, that "exterior" zip is not present. If I had not zipped it first, the release artifacts for the binaries would be `jacktrip` instlead of `JackTrip-<version>-<platform>-binary.zip`. I could try to circumvent that by moving the zip creation to the release deployment step, if that bothers you and others a lot.

I was just wondering. If it's easy to fix it would be good. But it's just a small weirdness.

dyfer:
> Didn't try more than to start it.

Sounds good, thanks. Building another one now that might have wayland support (not sure though).

>  If it's easy to fix it would be good. But it's just a small weirdness.

Not an easy fix, just a bit complicated. If we can leave it for now let's do that.

dyfer:
@ntonnaett could you try [this build](https://github.com/jacktrip/jacktrip/suites/2943299284/artifacts/66165618)? I don't think it fully works (doesn't have egl bindings?) but it was _some_ wayland support. Would be good to know to what extent this works.

Another build is in progress now as well.

ntonnaett:
This is what I get with bout builds.
```bash
./jacktrip 
Warning: Ignoring XDG_SESSION_TYPE=wayland on Gnome. Use QT_QPA_PLATFORM=wayland to run on Wayland anyway.
```

As you said, there are wayland links now, but it looks like it's using X11. That means it's blurry because it can't scale correctly with 4k displays. 
The build connects to our jacktrip server.

dyfer:
how about [this one](https://github.com/jacktrip/jacktrip/suites/2943718070/artifacts/66180594)?

ntonnaett:
No. But I'm not sure if the linked libraries are actually the problem. If I run it with `QT_QPA_PLATFORM=wayland ./jacktrip` it runs with wayland. Everything is sharp. Except there's no window...

dyfer:
> No. But I'm not sure if the linked libraries are actually the problem. If I run it with `QT_QPA_PLATFORM=wayland ./jacktrip` it runs with wayland. 

Is this the case with all three builds in this thread? [1](https://github.com/jacktrip/jacktrip/suites/2934741456/artifacts/65930498), [2](https://github.com/jacktrip/jacktrip/suites/2943299284/artifacts/66165618) or [3](https://github.com/jacktrip/jacktrip/suites/2943718070/artifacts/66180594)?

Everything is sharp. Except there's no window...

Ugh... any idea which libraries we need linked/included? This is what the lates binary used in QT: https://github.com/jacktrip/jacktrip/runs/2776553306?check_suite_focus=true#step:11:301 (i.e. do we need ` XComposite EGL`? or one of the others?)

ntonnaett:
Could you bundle libxcb directly from debian? Or aren't they providing static libs? Did you install the list from https://doc.qt.io/qt-5/linux-requirements.html?

ntonnaett:
No, you need the qt wayland platform for having the windowless experience. I think we should stick to qt 5.15 because using an old version could be the reason for that. The change is just that xcb is not delivered by qt anymore. 
[This](https://lists.debian.org/debian-user/2016/06/msg00431.html) looks like the dev packages include static libraries.

dyfer:
> No, you need the qt wayland platform for having the windowless experience. I think we should stick to qt 5.15 because using an old version could be the reason for that. The change is just that xcb is not delivered by qt anymore.
> [This](https://lists.debian.org/debian-user/2016/06/msg00431.html) looks like the dev packages include static libraries.

Are you sure we need to go back to 5.15? I don't see any static `libxcb` packages in ubuntu (searched with `apt info *xcb*static*`)

ntonnaett:
If 5.15 fixes wayland we definitely shouldn't go back to 5.12.

pkg-config gives me different flags with --static for xcb on ubuntu. So I assume it's included in the dev package. But to be honest, I should concentrate on flatpak.

dyfer:
> If 5.15 fixes wayland we definitely shouldn't go back to 5.12.
> 
> pkg-config gives me different flags with --static for xcb on ubuntu. So I assume it's included in the dev package. But to be honest, I should concentrate on flatpak.

When I built static Qt 5.15 (with "system xcb") I couldn't run JT (on ubuntu with X11), getting:
```
qt.qpa.plugin: Could not find the Qt platform plugin "xcb" in ""
This application failed to start because no Qt platform plugin could be initialized(...)
```

All information I could find was suggesting that I need to bundle the xcb plugin with the app, which defies the purpose of a single binary build...

Wayland is in Qt since 5.8 or 5.9 so I really don't think 5.12 in itself should be an issue...

I'm all in favor of flatpak. But I thought I'd want to make the single static build at least useable - so far it doesn't even start.

Could you please clarify what "Except there's no window..." means? No borders? Or does it start fullscreen?

dyfer:
Ok, now we're back to not working under X11. But wayland platform plugins seem supported. @ntonnaett does [this build](https://github.com/jacktrip/jacktrip/suites/2945672048/artifacts/66229480) run on your system?

dyfer:
It seems that using "system xcb" option, downgrading Qt to 5.12 and making sure that egl libraries are available when building (and turning Wayland on) I got a static build (with system xcb) that runs on Ubuntu 20.04: https://github.com/jacktrip/jacktrip/suites/2946582071/artifacts/66253380
@ntonnaett could you try this one on wayland?

ntonnaett:
> It seems that using "system xcb" option, downgrading Qt to 5.12 and making sure that egl libraries are available when building (and turning Wayland on) I got a static build (with system xcb) that runs on Ubuntu 20.04 with X11: https://github.com/jacktrip/jacktrip/suites/2947206910/artifacts/66272649
> @ntonnaett could you try this one on wayland?

Stil now window borders when using wayland.

ntonnaett:
Maybe this is of interest:
```bash
$ ./jacktrip -platform wayland --gui
Warning: Ignoring XDG_SESSION_TYPE=wayland on Gnome. Use QT_QPA_PLATFORM=wayland to run on Wayland anyway.
Failed to load client buffer integration: wayland-egl

Using the 'xdg-shell' shell integration
```

dyfer:
> Maybe this is of interest:
> 
> ```shell
> $ ./jacktrip -platform wayland --gui
> Warning: Ignoring XDG_SESSION_TYPE=wayland on Gnome. Use QT_QPA_PLATFORM=wayland to run on Wayland anyway.
> Failed to load client buffer integration: wayland-egl
> 
> Using the 'xdg-shell' shell integration
> ```

Thanks, thanks unfortunate. I'll look into that. BTW could you try [this build](https://github.com/jacktrip/jacktrip/suites/2945672048/artifacts/66229480)? This one doesn't run on my ubuntu because of QT 5.15, but I wonder if it works better on Wayland.

ntonnaett:
> > Maybe this is of interest:
> > ```shell
> > $ ./jacktrip -platform wayland --gui
> > Warning: Ignoring XDG_SESSION_TYPE=wayland on Gnome. Use QT_QPA_PLATFORM=wayland to run on Wayland anyway.
> > Failed to load client buffer integration: wayland-egl
> > 
> > Using the 'xdg-shell' shell integration
> > ```
> 
> Thanks, thanks unfortunate. I'll look into that. BTW could you try [this build](https://github.com/jacktrip/jacktrip/suites/2945672048/artifacts/66229480)? This one doesn't run on my ubuntu because of QT 5.15, but I wonder if it works better on Wayland.

This works with wayland. It doesn't comply to the theme and buttons of the rest of my desktop but I think that's negligible. There's even a small JackTrip logo in the corner.

ntonnaett:
Oh…
```bash
$ ./jacktrip -platform wayland --gui
Warning: Ignoring XDG_SESSION_TYPE=wayland on Gnome. Use QT_QPA_PLATFORM=wayland to run on Wayland anyway.

$ ./jacktrip
Warning: Ignoring XDG_SESSION_TYPE=wayland on Gnome. Use QT_QPA_PLATFORM=wayland to run on Wayland anyway.
qt.qpa.plugin: Could not find the Qt platform plugin "xcb" in ""
This application failed to start because no Qt platform plugin could be initialized. Reinstalling the application may fix this problem.

Available platform plugins are: wayland-egl, wayland.

Abgebrochen (Speicherabzug geschrieben)

```

First one runs. Second actually crashes… Do you set xcb as default platform somewhere?

dyfer:
> First one runs. Second actually crashes… Do you set xcb as default platform somewhere?

That's consistent with behavior on my ubuntu 20.04 - it seems that Qt 5.15 expects xcb plugin that's too new or something... i.e. since I don't have wayland, I get the `qt.qpa.plugin: Could not find the Qt platform plugin "xcb" in ""` message.

So, if anything, the previous build (Qt 5.12 + system xcb) is the one that seems to work on both system, albeit it's unfortnately running through X11 on wayland. The second build (Qt 5.15 + system xcb) doesn't seem to work on latest ubuntu LTS so is a no-go, even if it works under wayland...

Not sure where to go from here. I mean these builds to be operational for testing etc and I'd love to offer a static build for this purpose. But if it's impossible maybe we should just stop building static linux build? FWIW it works (I think) without the gui.

ntonnaett:
https://packages.ubuntu.com/focal/amd64/libxcb1-dev/filelist

libxcb1-dev has static build included and is version 1.14 on Ubuntu 20.04 (1.9 required according to [her](https://doc.qt.io/qt-5/linux-requirements.html)). Did you install all those packages that are listed?

.github/workflows/jacktrip.yml
```diff
@@ -98,7 +99,12 @@ jobs:
         if: runner.os == 'Linux'
         run: |
           sudo apt-get update
-          sudo apt-get install --yes libjack-dev qt5-default qt5-qmake qttools5-dev
+          sudo apt-get install --yes libjack-dev
+          if [[ -n "${{ matrix.static-qt-version }}" ]]; then 
+            sudo apt-get install --yes libfreetype6-dev libxi-dev libxkbcommon-dev libxkbcommon-x11-dev libx11-xcb-dev libdrm-dev libglu1-mesa-dev libwayland-dev libwayland-egl1-mesa libgles2-mesa-dev libwayland-server0 libwayland-egl-backend-dev
```
ntonnaett, line 104: Add all those listed packages here?

dyfer: I think all the needed packages are included - at some point working on this I removed the ones that said they are already installed.

Just in case triggering a new build installing libxcb explicitly.

ntonnaett: Could you add those?

```
    libfontconfig1-dev
    libfreetype6-dev
    libx11-dev
    libx11-xcb-dev
    libxext-dev
    libxfixes-dev
    libxi-dev
    libxrender-dev
    libxcb1-dev
    libxcb-glx0-dev
    libxcb-keysyms1-dev
    libxcb-image0-dev
    libxcb-shm0-dev
    libxcb-icccm4-dev
    libxcb-sync0-dev
    libxcb-xfixes0-dev
    libxcb-shape0-dev
    libxcb-randr0-dev
    libxcb-render-util0-dev
    libxcd-xinerama-dev
    libxkbcommon-dev
    libxkbcommon-x11-dev
```

dyfer: > I think all the needed packages are included - at some point working on this I removed from `apt install` the ones that said they are already installed.
> 
> Just in case triggering a new build installing libxcb explicitly.

ntonnaett: It's libxcb-xinerama-dev and not libxcd-xinerama-dev. I just copy and pasted.

dyfer:
I think this is the best we can do for now - JT is built with static Qt 5.12 and I think all dependent libraries installed. There should be no issues with external dependencies (I think). This runs fine with the gui on ubuntu 20.04 (I presume 18.04 as well?) with X11, as well as with CLI on most platforms. On wayland it runs through X11 emulation.

cchafe (approved):

PR #321, opened 2021-06-08
Weakjack part2

ntonnaett:
This moves weakjack from the subprojects directory into JackTrip's own src directory because weakjack is not a subproject. The subprojects directory is a special meson directory that is protected to only contain subprojects. Weakjack is supposed to be part of your projects source code (copied or as a git submodule).

Additionally this PR adds a weakjack option for meson. Tested on Linux. This seems to be broken with QMake because of the dl dependency.

PS: QMake is slow… and links jacknet because of some reason.

### DIFF
--- a/.github/workflows/meson.yml
+++ b/.github/workflows/meson.yml
@@ -93,16 +93,16 @@ jobs:
         env:
           DEVELOPER_DIR: '/Applications/Xcode_11.7.app/Contents/Developer' # specify XCode version on macOS
         run: |
-          meson --buildtype release $BUILD_PATH
+          meson --buildtype release -Drtaudio=enabled $BUILD_PATH
           cd $BUILD_PATH
-          meson configure -Ddefault_library=static -Drtaudio=enabled -Dweakjack=true
+          meson configure -Ddefault_library=static -Drtaudio:jack=disabled -Dweakjack=true
           meson compile
       - name: build on Windows
         if: runner.os == 'Windows'
         run: |
           meson --buildtype release -Drtaudio=enabled builddir
           cd builddir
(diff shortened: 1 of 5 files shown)


dyfer:
I would prefer if git submodules were kept separately from source. Doesn't have to be in `subprojects` if that's reserved for meson, but maybe some other directory?

dyfer:
If you're interested in parity with qmake build, we also use weakjack on macos. But I don't think it's necessary, since at least for now we plan to use qmake for release builds.

dyfer:
> Tested on Linux. This seems to be broken with QMake because of the dl dependency.

This is not clear - what does and does not work?

ntonnaett:
It doesn't link because dynamic loading on Linux needs the dl dependency.

dyfer:
> It doesn't link because dynamic loading on Linux needs the dl dependency.

Do you mean that weakjack in qmake doesn't build properly on Linux?

ntonnaett:
> > It doesn't link because dynamic loading on Linux needs the dl dependency.
> 
> Do you mean that weakjack in qmake doesn't build properly on Linux?

yes

ntonnaett:
@dyfer Could you test again? I couldn't find a way to show me linked libraries for macOS executables on Linux.

dyfer:
jack dependency is still there on macOS, tested from [build #580](https://github.com/jacktrip/jacktrip/actions/runs/919748847)

ntonnaett:
Had an epiphany. RtAudio depended on Jack.

dyfer:
macOS build is good now. I probably should check Windows before the final approval...

PR #322, opened 2021-06-08
[Meson] Remove rtaudio patch and use upstream meson

ntonnaett:
RtAudio merged the meson build files. Let's use those now.


dyfer (approved):
Thanks!

PR #323, opened 2021-06-09
Add warning for command line arguments in GUI mode

psiborg112:
Alert the user that command line options are currently ignored in GUI mode. And if they manage to stumble upon --test-gui, let them know they need to remove it for jacktrip to run normally.


dyfer (approved):
Thanks! Works as advertised. Tested on macOS.

Issue #218, opened 2021-01-30
Removal of RtAudio


ntonnaett:
Our RtAudio backend seems to be broken right now and we encourage to use Jack on every platform. If nobody fixes the RtAudio backend, we should remove it.

dyfer:
I'd be interested in continuing work on RtAudio and I'd like to keep the remnants of it, with the idea that the support might be re-added in the future.

antonrunov:
I did some related work in my fork (https://github.com/antonrunov/jacktrip/tree/ui-draft-on-1.1). It is based on old jacktrip version where rtaudio technically worked, but there still were issues that I addressed.

There is also a patch for more recent jacktrip version (https://github.com/kltdwrds/jacktrip/commit/5468484837a99770e05ca77d64026b3d00ee537a) that can be helpful.

I think for many musicians jack dependency is one of the greatest downsides of jacktrip. Having a lightweight client that can work directly with native audio driver would be a significant advantage.

ntonnaett:
I made RtAudio compile again with meson. #233 
We should incorporate your links next and then test.

ntonnaett:
> I did some related work in my fork (https://github.com/antonrunov/jacktrip/tree/ui-draft-on-1.1). It is based on old jacktrip version where rtaudio technically worked, but there still were issues that I addressed.

With your changes we would be able to define different devices than the default one? Maybe we test with the default device first?

> There is also a patch for more recent jacktrip version ([kltdwrds@5468484](https://github.com/kltdwrds/jacktrip/commit/5468484837a99770e05ca77d64026b3d00ee537a)) that can be helpful.

This looks like what I did in my RtAudio branch.

dyfer:
Thanks for working on this.
I just wanted to add, that ultimately it would be desirable to optionally select independent input and output devices (internal soundcard on macOS and Windows typically presents itself as separate input/output device)

antonrunov:
@ntonnaett Yes my changes allow selecting non-default devices. But even with default devices the current implementation might work incorrectly if the default input and output are different physical devices (meaning they are not synced). In this case there might be periodical audio issues. I hit this issue on my mac.

ntonnaett:
[Build artifacts](https://github.com/jacktrip/jacktrip/actions/runs/584267135)

Those artifacts are build with RtAudio support. Maybe somebody wants to try if they work.

mattahorton:
[...]

PR #324, opened 2021-06-09
QJackTrip: add basic backend selection

dyfer:
Since we are now building JackTrip with rtaudio on macOS and Windows, I thought it would be nice to have a way to use it when running the GUI.

I've added a menu for choosing between Jack and RtAudio (or should that be called "native" or "direct"?) and added samplerate setting for RtAudio. I have not added audio device setting since this is not fully functional right now AFAIU.

I think I've added storing everything in settings, the dialog disappears when in hub server mode, and should be permanently hidden when built without rtaudio.

~~I've used Qt Creator to add new elements to the layout. The diff for `jacktrip.ui` looks bigger than the changes I've made, not sure if I've done this right.~~ I tried to clean this up and manually copied only the new section into the original `jacktrip.ui`. The diff looks much nicer, but when the UI is opened in Qt Creator, new elements overlap with the old ones. The previous version, that was done purely in and looks better in Qt Creator, can be found [here](https://github.com/dyfer/jacktrip/blob/gui-rtaudio-backup01/src/gui/qjacktrip.ui). Which one is better?

This is meant to be an addition to #218

Please review carefully, I'm not confident with my changes :)

image

### DIFF
--- a/src/gui/qjacktrip.cpp
+++ b/src/gui/qjacktrip.cpp
@@ -793,15 +793,6 @@ void QJackTrip::loadSettings()
         m_ui->channelRecvSpinBox->setValue(settings.value("ChannelsRecv", gDefaultNumOutChannels).toInt());
     }
     
-#ifdef __RT_AUDIO__
-    m_ui->backendBox->setCurrentIndex(settings.value("Backend", 0).toInt());
-    m_ui->sampleRateBox->setValue(settings.value("SampleRate", 48000).toInt());
-    if (m_ui->backendBox->currentIndex() == 0) {
-        m_ui->sampleRateBox->setEnabled(false);
-    }
-    // TODO: load device
-#endif
-    
     m_ui->autoPatchComboBox->setCurrentIndex(settings.value("AutoPatchMode", 0).toInt());
     m_ui->zeroCheckBox->setChecked(settings.value("ZeroUnderrun", false).toBool());
     m_ui->timeoutCheckBox->setChecked(settings.value("Timeout", false).toBool());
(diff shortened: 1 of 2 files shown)


src/gui/qjacktrip.cpp
```diff
@@ -856,6 +891,11 @@ void QJackTrip::saveSettings()
     settings.setValue("LastAddress", m_ui->addressComboBox->currentText());
     settings.setValue("ChannelsSend", m_ui->channelSendSpinBox->value());
     settings.setValue("ChannelsRecv", m_ui->channelRecvSpinBox->value());
+#ifdef __RT_AUDIO__
+    settings.setValue("Backend", m_ui->backendBox->currentIndex());
+    settings.setValue("SampleRate", m_ui->sampleRateBox->value());
+    // TODO: save device
+#endif
```
psiborg112, line 898: It would be good to place this in a separate group.

```
settings.beginGroup("Audio");
settings.setValue("Backend", m_ui->backendBox->currentIndex());
settings.setValue("SampleRate", m_ui->sampleRateBox->value());
settings.endGroup();
```

Maybe have it just above or below the RecentServers group.

psiborg112: (And a corresponding change will need to be made in load settings)

psiborg112 (commented):
Code looks good, but there are a few things I want to do with the UI. (Pretty sure I can't submit to this branch, so mind if I send an edited file that you can submit?) One thing I want to check though before I do, can the rtaudio sample rate really be anything? Or should we instead have a combo box there with some preset values?

dyfer:
> Code looks good, but there are a few things I want to do with the UI. (Pretty sure I can't submit to this branch, so mind if I send an edited file that you can submit?) 

Sending the file would be fine, but there are two other possibilities 
1) I push it to a branch in the main repo (and reopen a PR) or
2) you make a PR against my branch on my fork (probably more cumbersone)

> One thing I want to check though before I do, can the rtaudio sample rate really be anything? Or should we instead have a combo box there with some preset values?

I wasn't sure what are the supported SRs by JackTrip so I don't want to limit it initially. 
I think the ideal UI would be a dropdown list with 44100, 48000, 96000 and (custom) where the user could type that in. But for initial pass I thought that any number would be good enough. Or do you think we should do a dropdown instead?

src/gui/qjacktrip.cpp
dyfer: Done, is that good?

psiborg112:
I think that for less technical users, limiting their choice would be good. I've set up a combo box that contains a common set of sample rates (lifted from qjackctl) that defaults to 48000.

dyfer:
Sounds good. Should I make the change, or are you planning to? And should I push this to a branch in the main repo?

psiborg112:
Have made the change. Just testing a few things then can probably upload a branch. (Has all of your code changes in it.)

dyfer:
superseded by #326

PR #325, opened 2021-06-09
Prepare flatpak for flathub

ntonnaett:
This should prepare JackTrip for flathub. But then I decided to add a flatpak build GHA as @dyfer proposed in the first place.

You can install the artifact with:
```bash
flatpak install --bundle jacktrip.flatpak
```

And then I saw we can actually just deploy to flathub with this. I think that would be great to have. I assume we have to go through the manual process of getting JackTrip into flathub first and afterwards be able to automate this. Flathub only includes stable builds (but has a beta branch nonetheless…). We might want a dev flatpak build which this PR includes and a second one that builds main and deploys it to flathub. We might need two flatpak manifests then.

---

The build complains about the icons (wrong name). But after installing it seems gnome gets the icon. But I definitely need to add 64x64 and 128x128 for flathub.


dyfer:
Thanks! This is great (though I'm still learning about flatpak).

I think I'd prefer this job to be included in the main jacktrip.yml workflow. Would you be OK with that? The reason is that it's easier to make jobs depend on each other that way.

> You can install the artifact with:

Is this a command to be run locally in the source? Or does `jacktrip.flatpak` refer to the online location? Is JT in flatpak (flathub?) repository already? I'd appreciate more info as these are all new things to me.

As for the version identifier, we could presumably run a string substitution first (like Aaron's mac/win installer generator scripts), getting version from the linux build (or by grepping the source?). This could be later addition though.

dyfer:
> Flathub only includes stable builds (but has a beta branch nonetheless…).

This is unclear, so are we doing stable or dev? If we want stable builds, we should run this job on tags only.

ntonnaett:
> Thanks! This is great (though I'm still learning about flatpak).
> 
> I think I'd prefer this job to be included in the main jacktrip.yml workflow. Would you be OK with that? The reason is that it's easier to make jobs depend on each other that way, if we need to.

I wasn't sure about that because it triggers differently.

> > You can install the artifact with:
> 
> Is this a command to be run locally in the source? Or does `jacktrip.flatpak` refer to the online location? Is JT in flatpak (flathub?) repository already? I'd appreciate more info as these are all new things to me.

jacktrip.flatpak is the name of the artifact. As I said we have to go through the manual process of adding jacktrip to flathub.

There are .flatpak bundles that you can install as described. There are .flatpakref files that are an easy way to install a flatpak from a repository (including getting updates). If you already have the repository installed where your flatpak comes from, you can install directly from there.

> As for the version identifier, we could presumably run a string substitution first (like Aaron's mac/win installer generator scripts), getting version from the linux build (or by grepping the source?). This could be added later though.

That's more something like a changelog. So getting this right would be a lot of work. We would need a common place for the changelog and getting the syntax right (Markdown for the documentation and this appstream file). By the way this appstream file isn't just used by flatpaks but by installation with common package managers too. It's necessary for showing JackTrip in apps like gnome-software.

ntonnaett:
> > Flathub only includes stable builds (but has a beta branch nonetheless…).
> 
> This is unclear, so are we doing stable or dev? If we want stable builds, we should run this job on tags only.

Dev only as bundles from github (means no updates) and stable from flathub. If we deploy on tags we have to be sure that the appstream file gets updated before.

ntonnaett:
There are two profiles now. The default one is for stable releases and *development* is for development…
```bash
meson configure -Dprofile=development
```

It's necessary for having two application ids: org.jacktrip.JackTrip for the stable flatpak and org.jacktrip.JackTrip.Devel for the development version.

Before merge, I have to change the git branch from prepare_flatpak back to dev.

dyfer:
Do we want to get any of the flatpak stuff before the RC?

ntonnaett:
Would you be willing to test the flatpak file on Ubuntu?

dyfer:
> Is the release trigger right? It should trigger on release and prerelease. But should there be an additional branch: main too?

I don't know actually. I've only ever used "on push" and then checked for tags `if: startsWith(github.ref, 'refs/tags/')`

ntonnaett:
I decided for that because I read that only checking for tags could be problematic. But I didn't check why... Have to search for it again.

dyfer:
It's likely a correct to do it, it's called "on release" but I'm just saying that I haven't tested. WOuld be good to test with a test release.

ntonnaett:
org.jacktrip.JackTrip.Devel is now built from dev. That means that the Flatpak build won't work correctly until this is merged.

dyfer (approved):
Just pointing out that once we merge this, it will always be visible in [Actions](https://github.com/jacktrip/jacktrip/actions) even if me move this at some point. That's not a big deal, just want to make sure we know this before making a decision to keep that as a separate workflow.

PR #327, opened 2021-06-10
Use wix preprocessor

psiborg112:
Turns out that wix has a preprocessor. Use this to simplify installer creation. (Can substitute version info into the installer more easily, and don't need a separate files_static.wxs)


dyfer (approved):
Thanks!

PR #334, opened 2021-06-11
UI: add buffer size setting for RtAudio

### LINKED ISSUE #331: Missing buffer size setting for RtAudio in the GUI
I forgot to add buffer setting for the RtAudio gui implementation. We should add this soon.
Probably a dropdown menu with powers of 2 between 16 and 1024 (as suggested by Chris)

dyfer:
Add buffer size setting to RtAudio settings in the GUI. Fixes #331.

I've edited `qjacktrip.ui` by hand - feel free to correct as needed.


psiborg112:
Looks good. I'm making some more UI fixes at the moment, but I think we should merge this first. Ultimately I think this should probably go in a separate Audio tab, especially now that we're adding more options. (Once we add device selection we should definitely move it across, but I think it's fine here for the moment.)

psiborg112 (approved):

PR #336, opened 2021-06-11
Strip rc part of version for windows installer script

psiborg112:
Wix only supports version numbers of the form x.x.x. Strip the rc part of our version to build the installer.


dyfer (approved):
Thanks so much!

IIUC that means that the release candidate(s) and the final version will have the same version in the installer. Could this create any issues?

psiborg112:
You are correct. And it shouldn't create any issues.

PR #337, opened 2021-06-11
Check if jack is available

psiborg112:
If built with weakjack, check if jack is available. If not and the rtaudio backend has been built in, fallback on this and show a warning dialog which the user can choose to hide in the future. If no other backend is available, show a warning that cannot be suppressed.

Also, remove redundant code to change UI based on audio backend choice in the loadSettings function. (This is taken care of when the backendComboBox value is set via the connected signal and slot.)

### DIFF
--- a/src/gui/qjacktrip.cpp
+++ b/src/gui/qjacktrip.cpp
@@ -33,6 +33,9 @@
 #include <QFileDialog>
 #include <cstdlib>
 #include <ctime>
+#ifdef USE_WEAK_JACK
+#include "weak_libjack.h"
+#endif
 
 #include "../Limiter.h"
 #include "../Compressor.h"
@@ -47,7 +50,8 @@ QJackTrip::QJackTrip(QWidget *parent) :
     m_jackTripRunning(false),
     m_isExiting(false),
     m_hasIPv4Reply(false),
-    m_argc(1)
+    m_argc(1),
+    m_hideWarning(false)
 {
     m_ui->setupUi(this);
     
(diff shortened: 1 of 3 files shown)


src/gui/qjacktrip.cpp
```diff
@@ -832,10 +876,6 @@ void QJackTrip::loadSettings()
     m_ui->backendComboBox->setCurrentIndex(settings.value("Backend", 0).toInt());
     m_ui->sampleRateComboBox->setCurrentText(settings.value("SampleRate", "48000").toString());
     m_ui->bufferSizeComboBox->setCurrentText(settings.value("BufferSize", "128").toString());
-    if (m_ui->backendComboBox->currentIndex() == 0) {
```
dyfer, line 835: Yeah, I understand this shouldn't be needed, but when I first implemented it, for some reason I observed not having these boxes disabled on load. Maybe it was a build error or something. Works fine in this PR.

src/gui/qjacktrip.cpp
```diff
@@ -210,6 +214,46 @@ QJackTrip::QJackTrip(QWidget *parent) :
     } else {
         m_autoQueueIndicator.setText("Auto queue: disabled");
     }
+    
+#ifdef USE_WEAK_JACK
+//Check if Jack is actually available
+    if (have_libjack() != 0) {
+#ifdef __RT_AUDIO__
```
dyfer, line 221: IUUC correctly this only supports weakjack + rtaudio configuration, correct? While I don't think we plan to ship that, from a configuration standpoint it might be prudent to allow jack + rtaudio config as well (jack, not weakjack)

ntonnaett: I'm thinking about jack + rtaudio for flatpaks to use the pulseaudio backend.

psiborg112: This is to detect the situation where libjack isn't available on the system and can't be dynamically loaded by weakjack so the jack backend is unusable. All of my builds are dynamic, so the test you're suggesting would never happen without weakjack because of a segfault. Is the jack library itself being statically linked in the static builds?

ntonnaett: You can't (or you shouldn't) link jack statically. Your code looks good to me.

psiborg112: Yeah, that's what I thought. So we can only test for this situation if we're using weakjack.

dyfer: Got it. I originally thought that some of that ifdef guargs RtAudio-only functionality (which we want to maintain even if we're not using weakjack) but that wasn't correct.

Just to clarify one point:
I think @psiborg112 meant _loading_ jack dynamically, right? Sure, we shouldn't link jack statically, but when it's _linked_ dynamically (not loaded dynamically, i.e. not using weakjack) when the app wouldn't even start if jack is not present.

dyfer (approved):

PR #338, opened 2021-06-11
Prepare 1.4.0 rc.2

dyfer:
- ~~fix XML file, as suggested by @ntonnaett~~
- update version to 1.4.0-rc.2
- flatpak stable deployment only runs on the main branch (as requested by @ntonnaett)

~~EDIT: I removed changes to the flatpak XML file from this PR as they did not work. I don't know what's wrong there - [here's what I've tried](https://github.com/dyfer/jacktrip/blob/xml-change-copy/linux/org.jacktrip.JackTrip.metainfo.xml.in).~~

~~Flatpak GitHub Action failure should not interfere with the release candidate, so IMO we can go ahead with rc.2. I think @ntonnaett is on board with that :) ~~ Flatpak issue is now resolved

### DIFF
--- a/.github/workflows/flatpak-release.yml
+++ b/.github/workflows/flatpak-release.yml
@@ -6,6 +6,7 @@ jobs:
   flatpak:
     name: "Flatpak"
     runs-on: ubuntu-latest
+    if: startsWith(github.ref, 'refs/head/main')
     container:
       image: bilelmoussaoui/flatpak-github-actions:kde-5.15
       options: --privileged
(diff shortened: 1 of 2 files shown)


ntonnaett:
After #340 is merged the flatpak GHAs should work.

ntonnaett:
The change from the metainfo file should be removed from this PR.

dyfer:
> The change in the metainfo file should be removed from this PR.

Done!

ntonnaett (approved):

PR #339, opened 2021-06-11
Update CHANGELOG.md

dyfer:
Clarify QJackTrip -> JackTrip

I thought it might be good to include that in the changelog? I'm open to suggestions in terms of wording etc.

### DIFF
--- a/docs/About/CHANGELOG.md
+++ b/docs/About/CHANGELOG.md
@@ -3,7 +3,7 @@ Changelog
 
 1.4.0
 ---
-- (added) optional GUI from QJackTrip
+- (added) optional GUI from QJackTrip (the application is now called JackTrip, even though it includes the gui previously released as QJackTrip)
 - (added) authentication in hub server mode
 - (added) different number of sending and receiving channels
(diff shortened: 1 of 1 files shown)


ntonnaett:
I think this should be shorter.

dyfer:
> I think this should be shorter.

Sure. Suggestions?

ntonnaett:
- include GUI from QJackTrip in JackTrip
- merge GUI…

dyfer:
I would really like to indicate more clearly that QJackTrip is no more. But maybe we don't need to do it in the changelog, but instead on the release page

cchafe:
--this seems to make sense--
(added) include GUI from QJackTrip in JackTrip
--and then--
(changed) QJackTrip and JackTrip are now identical

dyfer:
How about this? (also, this is somewhat orthogonal to the release, we can tag rc.2 first as this will go on the github.io website anyway right after merging)

Also, if merging - please squash merge

cchafe (approved):

PR #341, opened 2021-06-12
Deploy fix

dyfer:
Unfortunately the deployment did not work... I think this is the missing part.

### DIFF
--- a/.github/workflows/jacktrip.yml
+++ b/.github/workflows/jacktrip.yml
@@ -4,6 +4,7 @@ on:
     branches:
       - main
       - dev
+    tags:
     paths-ignore:
       - 'docs/**'
       - 'mkdocs.yml'


ntonnaett:
I'm not sure about softprops/action-gh-release. Isn't it supposed to create a release if you for example tag something v*.*.*? But we already have a release or prerelease.

ntonnaett:
Wait… Why is it a draft?

dyfer:
> Wait… Why is it a draft?

If you push a tag with git (as opposed to creating a release with github web interface), the release action will create an actual release in GH and yes, it will initially be a draft. The idea was that if we want to update description, sign binaries, we do that after the draft is created, and then turn it into a non-draft.

The reason deployment didn't work is because GHA did not trigger at all when the tag was pushed, if you take a look at our actions.

The deployment was tested on a branch when I was working on it, you could see if/how it works in #294 The problem is that when testing I had to temporarily make it run on all branches, and I failed to test this again after merging. My bad.

After this is merged, I'll test again deployment on my fork and will confirm whether it works.

ntonnaett (approved):

dyfer:
Didn't work... I'll make another PR.

PR #345, opened 2021-06-12
GHA: deploy on ALL tags

dyfer:
I _think_ this should work...


dyfer:
This seems to be properly triggering [on my fork](https://github.com/dyfer/jacktrip/actions/runs/931558028), but let's wait for this job to finish to make sure the artifacts are uploaded etc.

ntonnaett (approved):

PR #349, opened 2021-06-16
added bufstrategy 3 (experimental) to UI choices

cchafe:
the name says it

### DIFF
--- a/src/gui/qjacktrip.ui
+++ b/src/gui/qjacktrip.ui
@@ -6,7 +6,7 @@
    <rect>
     <x>0</x>
     <y>0</y>
-    <width>403</width>
+    <width>409</width>
     <height>947</height>
    </rect>
   </property>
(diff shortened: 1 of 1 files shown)


src/gui/qjacktrip.ui
```diff
@@ -146,7 +146,7 @@ To connect to a hub server you need to run as a hub client.</string>
        </size>
       </property>
       <property name="currentIndex">
-       <number>0</number>
+       <number>2</number>
```
psiborg112, line 149: This should be 0 so that the app initially opens on to the first tab. (Qt Designer always saves the currently viewed tab as the default one. I've lost track of the number of times I've been stung by this myself.)

cchafe: How to edit this? Designer thinks it's set for 0 and manually editing the file isn't allowed.

psiborg112: If you save it while you're viewing the first tab, that'll do the job.

cchafe: Thanks! That did the trick, fixed in latest commit

psiborg112 (approved):

PR #351, opened 2021-06-24
docs: update Install.md

dyfer:
This PR updates information on downloading JackTrip from GitHub releases, as well as provides a link to the latest dev builds. 

Open for any and all suggestions!

EDIT: should I mention anything about Jack for macOS and Windows? JT should work with all Jack versions now, but also does not require it, if one is satisfied with the default input/output. This is least tested on Windows. What should go in the install instructions thought?

### DIFF
--- a/docs/Install.md
+++ b/docs/Install.md
@@ -14,16 +14,17 @@ On Linux the easiest way to install JackTrip is to use the distribution's packag
     sudo apt install jacktrip
     ```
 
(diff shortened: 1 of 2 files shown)


ntonnaett (approved):

dyfer:
I'll merge this for now in the interest of getting updated info up quickly, we can add information about Jack later.

PR #354, opened 2021-07-05
Dev jtf logos

cchafe:
would be great if someone could check these on a running instance after build & install for win and mac


ntonnaett:
I like the logo. But this would mix up JackTrip and the JackTrip Foundation.

dyfer:
One comment for macOS: the application bundle I've downloaded still has the old logo. I think `macos/jacktrip.icns` also needs to be put in `macos/JackTrip.app_template/Contents/Resources/jacktrip.icns`

dyfer (approved):
I've tried macOS and Windows builds and the new icons are shown properly there.

(I haven't tried Linux but I assume @cchafe has covered that?)

ntonnaett:
When install with meson I get the right icon when searching for JackTrip in GnomeShell but the running JackTrip has no icon at all.

ntonnaett:
Same with dev, so it shouldn't be a problem for this PR. Have to investigate.

Issue #355, opened 2021-07-07
AudioTester: stack use after scope


ntonnaett:
I've got another catch. jacktrip is build with `meson configure -Db_sanitize=address,undefined`.

The AudioTester object lives in Settings. A pointer is transferred to the jacktrip object. When app->exec() is run in main(), settings is already out of scope. That might be the problem.

```bash
./jacktrip -C sprawl --receivechannels 2 --sendchannels 1
Setting JACK Process Callback...
SUCCESS
---------------------------------------------------------
The Sampling Rate is: 48000
---------------------------------------------------------
The Audio Buffer Size is: 1024 samples
                      or: 4096 bytes
---------------------------------------------------------
The Number of Channels is: 1
---------------------------------------------------------
Using UDP Protocol
---------------------------------------------------------
Using JitterBuffer strategy 1
Server Address set to: sprawl Port: 61002
---------------------------------------------------------
UDP Socket Receiving in Port: 4464
---------------------------------------------------------
Waiting for Peer...
=================================================================
==6029==ERROR: AddressSanitizer: stack-use-after-scope on address 0x7fffbbd1e950 at pc 0x00000082fe8d bp 0x7f820789b780 sp 0x7f820789b778
READ of size 1 at 0x7fffbbd1e950 thread T1
    #0 0x82fe8c in AudioTester::getEnabled() ../src/AudioTester.h:91
    #1 0x82fe8c in AudioInterface::callback(QVarLengthArray<float*, 256>&, QVarLengthArray<float*, 256>&, unsigned int) ../src/AudioInterface.cpp:223
    #2 0x894c8d in JackAudioInterface::processCallback(unsigned int) ../src/JackAudioInterface.cpp:304
    #3 0x7f821e01221f in on_rtsocket_condition (/usr/lib64/pipewire-0.3/jack/libjack.so.0+0x1721f)
    #4 0x7f8215f3d70a in loop_iterate (/usr/lib64/spa-0.2/support/libspa-support.so+0x870a)
    #5 0x7f821bbf5512 in do_loop (/lib64/libpipewire-0.3.so.0+0x3d512)
    #6 0x7f821be68298 in start_thread (/lib64/libpthread.so.0+0x9298)
    #7 0x7f821bd90352  (/lib64/libc.so.6+0x100352)

Address 0x7fffbbd1e950 is located in stack of thread T0 at offset 1024 in frame
    #0 0x43fa6f in main ../src/main.cpp:160

[...]

PR #356, opened 2021-07-08
Fix stack use after scope

ntonnaett:
Fixs #355. No smart pointer needed.

### DIFF
--- a/src/main.cpp
+++ b/src/main.cpp
@@ -161,6 +161,7 @@ int main(int argc, char* argv[])
     QScopedPointer<QCoreApplication> app(createApplication(argc, argv));
     QScopedPointer<JackTrip> jackTrip;
     QScopedPointer<UdpHubListener> udpHub;
+    Settings settings;
 #ifndef NO_GUI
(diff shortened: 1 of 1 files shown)


psiborg112:
This works, but I think smart pointers would be the safer way to handle it in the long run. At the very least, I think there needs to be a brief comment in the code about why the variable is created there because it would be easy to reintroduce the bug if the code is ever restructured, or if someone decides they can "optimise" it.

ntonnaett:
Why do you think smart pointers are safer? I mean this goes out of scope when the whole application exits.

I will add a comment. The whole configuration has to be restructured at some point.

psiborg112:
Safer in terms of this bug being less likely to reappear if someone restructures the code in main, although obviously a warning comment will do the job. It just seems like bad practice to have code so much further down the chain being dependent upon the command line parsing object remaining resident in memory long after its work is done.

psiborg112:
(To clarify, I am still happy to approve this with just the warning comment.)

psiborg112 (approved):

PR #357, opened 2021-07-08
Unregister jack ports before closing client to prevent indirect memory leaks (with pipewire)

ntonnaett:
This prevents indirect memory leaks (at least in pipewire). Jack's
simple_client looks like this shouldn't happen and is probably a
bug in pipewire. But unregistering ports doesn't hurt.


psiborg112:
Yeah, my understanding is that it shouldn't strictly be needed, but it also doesn't hurt, and if it works around a discrepancy in pipewire's implementation then that's always a plus.

psiborg112 (approved):

PR #359, opened 2021-07-14
Include Qt5 with -isystem in Meson and fix some clazy warnings

ntonnaett:
This prevents warnings from tools like clazy that we can't fix. It's a really big cleanup. So I fixed already two for loops that I missed the last time.

### DIFF
--- a/meson.build
+++ b/meson.build
@@ -85,9 +85,9 @@ endif
 
 if get_option('nogui') == true
 	defines += '-DNO_GUI'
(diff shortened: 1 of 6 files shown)


ntonnaett:
@psiborg112 Could you look into the constexpr jacktrip_global.h commit? I changed QString to const char* to get rid of some clazy warnings. But we had problems before changing things away from QString.

cchafe (approved):

Issue #361, opened 2021-07-14
Default to Meson build for Linux in 1.4.0


ntonnaett:
We should encourage the use of the meson build for Linux. 

The advantages for the end user right now are:
- Manpage
- Icons
- Metainfo / Appstream data
- maybe LTO on some distributions

dyfer:
We've talked about this during the meeting on Jul 14.

I think this should be accompanied by the change to the CI, which should happen after qmake and meson builds are put into the same action (which I plan to work on). My preference would be to:
- release 1.4.0 first
- I plan to unify qmake and meson builds in CI after 1.4.0 (after we merge `dev` into `main`),
- then we merge this (possibly updating the changelog)
- In our - then unified - github actions script I'd switch our nightly Linux builds to use meson

ntonnaett:
This would lead to bad user experience on Linux. You wouldn't find JackTrip in your Desktop Enviroment without a desktop file. There is no reason to make this depend on CI changes.

dyfer:
I think the nightly/release builds should be built with the same instructions as the "default" in the instructions for consistency.

ntonnaett:
Package maintainers wouldn't use meson and end up with a package that doesn't show up in software centers or your application launchers. The only way for the enduser then is to start jacktrip from the terminal. So this should really happen before we call it 1.4.0. Is the change in the CI really big? Why not do it before the release?

ntonnaett:
Small Sidenote: Meson 0.59 sets up the visual studio environment automatically on Windows (for 64bit) now, if no other compiler toolchain is detected. So Windows use becomes easier now.

dyfer:
> a package that doesn't show up in software centers or your application launchers.

Since we are supporting 3 build systems, we could (should?) make all of them create the necessary files to provide desktop environment integration. What's needed in order to have that with the qmake build?

> Is the change in the CI really big?

Not really big, but the main CI now runs on main/dev and PRs, and not on every push. I guess that's not really a big deal, and/or I could change it to make it run on every push until 1.4.0 release. All this is because the manual trigger for the builds (which we agreed to use instead of builds on every push) only appears when the CI changes are on the default branch (`main` in our case).

I see the following options:
1. we release 1.4.0 as is with suboptimal linux user experience
2. I make the CI changes before 1.4.0 and we change nightly builds to meson (I'd suggest that we again build on all pushes, until 1.4.0 release)
3. make qmake build create proper desktop environment integration files

[...]

PR #363, opened 2021-07-14
Default to Meson for Linux

ntonnaett:
Maybe this is all we need for #361. But we could elaborate more.


ntonnaett:
Should we change the build script?

ntonnaett:
I will add a link from the QMake build.

PR #367, opened 2021-07-16
Backend tab

psiborg112:
Move the audio backend settings into their own tab. This will declutter the basic options tab and give us space to implement additional backend settings moving forward.


ntonnaett (approved):
This looks good and works for me with JACK. RtAudio is broken for me on dev too. Maybe it's a pipewire issue.

Issue #358, opened 2021-07-11
Limiter license mismatch


ntonnaett:
https://github.com/jacktrip/jacktrip/blob/551cb18b273085613c23dd82751672a567ecb974/faust-src/limiterdsp.dsp#L8

If this is the current base of our c++ code, all builds of JackTrip are GPLv3. We should document this before the release.
Transpiling a new C++ file would override all changes we did to that file.

dyfer:
(notes from the dev meeting Jul 14)
- let's add a compile switch to switch between GPL-licensed and MIT-licensed limiter (qmake/meson/cmake)
- it can default to GPL limiter
- (Marcin: I'd explicitly specify that option in the build system in so that we know that our automated builds using the GPL version, which has been used already for a while AND we are building with GUI which makes everything GPL)
- document building a MIT-licensed jacktrip (no GUI, no weakjack, MIT limiter)

ntonnaett:
For whom or what are we doing this effort of maintaining a MIT version? This feels quite unnecessary without knowing.

dyfer:
> For whom or what are we doing this effort of maintaining a MIT version?

Maybe @cchafe can answer that?

dyfer:
I started looking into this and it looks like this `jacktrip/faust-src/limiterdsp.dsp` file needs to be processed with faust in order to generate a different limiter into `limiterdsp.h`. This means that we can't change this at compile time, we'd need to run faust to generate a different limiter. 
It seems that we also use freeverb and zita which are LGPL, so a more permissive license than GPL, but what does that mean for the rest of the project? Do we need switches for them or are they implicitly compatible with MIT?
So, would we want to generate two limiters (MIT and GPL) using faust, and switch between them in our build systems?

ntonnaett:
As nobody came forward to explain why this is necessary, I don't see a reason for that. Just make JackTrip GPL. It's much easier.

LGPL would make JackTrip LGPL. You can't make a closed source project out of it then (only as a service...).

PR #370, opened 2021-07-21
[Linux] Add symbolic icon

### LINKED ISSUE #365: Symbolic monochrome icon missing
Gnome Shell shows a symbolic link of applications in the topbar. Right now JackTrip's icon gets converted to greyscale and this doesn't look right. We should add a black and white version of the logo. SVG should be sufficient.

ntonnaett:
This fixes #365 thanks to @codedoser. The symbolic icon is most prominently shown in the top bar of Gnome Shell when the JackTrip window is focused. But there might be other occurrences I'm not aware of.


dyfer:
This looks fine, but it made no change to the icon appearance on my system (Ubuntu 20.04, Gnome 3.36.8). Actually, even if I remove all the icons (48px, scalable, and this one symbolic), there's still a little icon on the top bar when JT is in focus (despite showing no app icon otherwise). Looks like it's cached somewhere - might be something about my system...

Have you tested this and does it appear correctly for you? Which version of Gnome have you tested with?

ntonnaett:
What icon do you see? The old one?

ntonnaett:
Did you logout and login again after installing the icon? I forgot (to mention) that gnome seems to create the icon cache at startup.

dyfer:
> Did you logout and login again after installing the icon? 

Yes, I even restarted.

> What icon do you see? The old one?

Yes, I see the old icon, but again it's weird, because I can see it even after 1) removing all icons manually and 2) restarting the system. Not sure where it's cached.

ntonnaett:
What prefix did you use? What Qt version are you using? I remember that there was a change to use those icon files instead of a qt resource. But I'm not sure how a qt resource icon could work on Linux. But this would be only one type of icon I think.

dyfer:
> What prefix did you use? 

`/usr/local`, seems that was the default

> What Qt version are you using? 

5.12.8

> I remember that there was a change to use those icon files instead of a qt resource. But I'm not sure how a qt resource icon could work on Linux. But this would be only one type of icon I think.

Not sure I understand. Is there a qt resource icon that might be in use instead? Where's that located?

ntonnaett:
Could you try using just `/usr` as prefix? Usually you wouldn't want this because it can conflict with your distribution's installation of jacktrip.

qt resources are precompiled resources into C++ or C source code. It has pros and cons. One con is that it is compiled into the binary. What is also a pro...

dyfer:
I'll try `/usr` and report back.
I skimmed through our source and I can see no qt resource icons, aside from one for windows that seems to be used only in cmake build, AFAICT.

ntonnaett:
https://github.com/jacktrip/jacktrip/blob/dev/src/gui/qjacktrip.qrc

dyfer:
> https://github.com/jacktrip/jacktrip/blob/dev/src/gui/qjacktrip.qrc

Ah, I missed that. Thanks, good to know.

dyfer:
In that case, wouldn't it be better to modify `icon.png` to be thicker (though still in color)? I'm not sure it would be, but maybe something to consider...?

ntonnaett:
This is the main icon on other platforms.

dyfer:
> This is the main icon on other platforms.

I think it's only used on Windows and my thought was that since it's small anyway it wouldn't be bad if it's thicker. It's just a suggestion, if you don't think that's a good approach that's fine.

As for this PR, it looks fine, but I really can't tell if it's doing anything on my system.

dyfer:
After installing in `/usr`, here's my result:
- symbolic icon shows up as the JT icon in the application launcher (instead of the "red" icon)
- the icon in the taskbar is the correct "red" icon
- the icon on the top bar is still the previous one (thin)

I think the first point is suboptimal, I'd expect to see the main JT icon when searching for it in the launcher. 

Are you seeing the correct behavior on your system with this change?

ntonnaett:
You can uninstall the last install with `ninja uninstall`.

Could you try again? I removed the symbolic from the filename.

dyfer:
I still would like to get the answer whether this works/worked on your system.

ntonnaett:
If it wouldn't work on my system I would try to fix that first. But Ubuntu should also work. At least the wrong launcher icon is fixed now.

dyfer:
Just so it's clear: with the latest update to this PR, _I see no change in behavior in relation to what this PR originally provided_. I get the symbolic icon in the launcher (undesirable), and "thin" icon in the top bar (no change from the regular `dev` build).

dyfer:
@ntonnaett I was trying to move around that symbolic icon between `/usr/` and `/usr/local` prefixes, installing JT in either, naming the icon with and without "-symbolic" and I can't reproduce having the symbolic icon show up in my application launcher anymore. (It still never showed up at the top bar, where it should've). Maybe there's something weird going on with my system. Someone else on Ubuntu should probably try this (maybe at some point I'll set up a VM with a clean install...). Sorry I can't be of more help.

ntonnaett:
Thanks for your help. I appreciate that. Last night I had an epiphany. Ubuntu uses X11 instead of Wayland. The symbolic icon that is chosen from the WM_CLASS name that is set by the application. That defaults to lowercase application id on wayland but the binary name on X11. What's missing here is that JackTrip sets the WM_CLASS. That should be possible by setting the desktop file's basename with QGuiApplication::setDesktopFileName(). But right now JackTrip is only a QCoreApplication. @psiborg112 Could you help? Is it possible to switch to QGuiApplication for the gui runtime? Or did you try that already?

dyfer:
I support fixing/adding this, but since it's a rather cosmetic issue and it might involve more work to get right, maybe we don't need this to happen in 1.4.0? IMO the users would benefit if we release 1.4.0 sooner with all its improvements already in. 

Other than this it seems we are ready for either another RC (due to the recent PR for closing jack ports) or possibly for the final release, what do you think?

ntonnaett:
Yes. I tried a lot today to fix this. No clue what's wrong with X11. Let's release 1.4.0. There's only #358 left after this.

ntonnaett:
I think we can merge this. So it works on wayland and fix X11 after the release.

dyfer:
All right. Should I add the icon copy to qmake or would you do that? We'd need to add something like this towards the end of `jacktrip.pro` (untested):

````
icon_symbolic.extra = cp $${PWD}/linux/icons/jacktrip-symbolic.svg $${OUT_PWD}/$${application_id}-symbolic.svg
icon_symbolic.CONFIG += no_check_exist
icon_symbolic.files = $${OUT_PWD}/$${application_id}-symbolic.svg
icon_symbolic.path = $${PREFIX}/share/icons/hicolor/symbolic/apps

INSTALLS += (...) icon_symbolic
```

ntonnaett:
Installation with QMake works as expected.

dyfer (approved):
Thanks! Sorry for the delay.

PR #372, opened 2021-07-30
Meson ci fixes

ntonnaett:
This is based on discussions in #361 

- Add a workaround for missing coreaudio linking flags in RtAudio's pkgconfig file
- ~~Add an option to set if RtAudio should be linked statically (even with package manager's rtaudio or default_library=shared)~~

### DIFF
--- a/meson.build
+++ b/meson.build
@@ -127,7 +127,11 @@ endif
 
 if host_machine.system() == 'darwin'
 	src += ['src/gui/NoNap.mm']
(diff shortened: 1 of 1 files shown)


ntonnaett:
Okay. I encountered [this](https://github.com/mesonbuild/meson/issues/8050) bug with this. I will remove the second commit.

dyfer (approved):
Thanks! I've build it locally and it works fine.

PR #385, opened 2021-08-23
Fix RtAudio device selection

### LINKED ISSUE #256: Rtaudio set device id broken
Unable to choose device to use over ID. Only the default audio set in system audio settings is used.
MacOS 10.15, dev-branch

dyfer:
This PR adds the following functionality:
- setting RtAudio input/output devices independently (from CLI, by name)
- listing audio devices
- selecting RtAudio devices from the GUI

I think this should complete #218; it also fixes #256 (more on that below). 

Here are some aspects I'd like to point the attention of reviewer(s) to:

- I decided to favor selecting device by name, instead of by ID. Device IDs might change when audio devices are plugged/unplugged and I strongly feel we should discourage/deprecate using ID. I have fixed selection by ID, but marked it as "deprecated" in the help. However, currently there's no warning posted if the user uses this deprecated argument. **Should we post a warning when the `--deviceid` argument is used?**
- I've added more `ifdef` guards around RtAudio code. Most notably, builds with no RtAudio will now _not recognize_ the `--rtaudio` and related flags. I think this is the right way to do it, but I'm open to suggestions.
- I'm not enforcing the use of duplex devices in any way. Mixing different input/output hardware devices is not well tested; it should be fine for built-in devices on all platforms, and internal devices are in fact the primary use for independent input/output functionality. On macOS it _might_ be fine to mix e.g. built-in and external device, but I'm not sure. On Windows, when using ASIO, the same device needs to be used for both input and output. Again, I'm not checking for this right now, so this should be clearly documented.
- Based on my experience with device selection in SuperCollider (and parsing device names in CLI), I suspect that on Windows selecting built-in devices from command line _on non-English OS installs_ might not work, since on that platform translated names of built-in devices might include non-ASCII characters and  we'd need to use `w_main` function and convert `wchar` to `char` to properly parse that in CLI. Let's cross that bridge if/when we get there. _Selecting devices from the GUI should work fine nonetheless._
- The list of devices in the GUI will not automatically update on un-/re-plugging the devices, but it will refresh when switching from Jack to RtAudio backend, or with the provided button at the bottom of that tab.


src/RtAudioInterface.cpp
```diff
@@ -66,43 +66,59 @@ void RtAudioInterface::setup()
...
+            cout << "Selecting default OUTPUT device" << endl;
+            deviceId_output = mRtAudio->getDefaultOutputDevice();
+        }
     }
+    
+    auto dev_info_input = mRtAudio->getDeviceInfo(deviceId_input);
+    auto dev_info_output = mRtAudio->getDeviceInfo(deviceId_output);
 
-    if (getNumInputChannels() > dev_info_input.inputChannels) {
+    if ((unsigned int)getNumInputChannels() > dev_info_input.inputChannels) {
```
ntonnaett, line 111: Let's use the more precise C++ static_cast<unsigned int>() instead of C cast.

src/Settings.cpp
```diff
@@ -140,9 +146,13 @@ void Settings::parseInput(int argc, char** argv)
...
          OPT_APPENDTHREADID},                        // Append thread id to client names
+#ifdef __RT_AUDIO__
         {"rtaudio", no_argument, NULL, 'R'},         // Run in JamLink mode
         {"srate", required_argument, NULL, 'T'},     // Set Sample Rate
-        {"deviceid", required_argument, NULL, 'd'},  // Set RTAudio device id to use
+        {"deviceid", required_argument, NULL, 'd'},  // Set RTAudio device id to use (DEPRECATED)
+        {"audiodevice", required_argument, NULL, OPT_AUDIODEVICE},  // Set RTAudio devices by name
+        {"listdevices", no_argument, NULL, OPT_LISTDEVICES},  // Set RTAudio device id to use
+#endif
         {"bufsize", required_argument, NULL, 'F'},   // Set buffer Size
```
ntonnaett, line 156: Is buffer size also RtAudio-only like sample rate?

src/jacktrip_globals.h
```diff
@@ -86,7 +86,7 @@ constexpr AudioInterface::audioBitResolutionT gDefaultBitResolutionMode =
 constexpr int gDefaultQueueLength              = 4;
 constexpr int gDefaultOutputQueueLength        = 4;
 constexpr uint32_t gDefaultSampleRate          = 48000;
-constexpr uint32_t gDefaultDeviceID            = 0;
+constexpr uint32_t gDefaultDeviceID            = -1;
```
ntonnaett, line 89: This isn't unsigned.

ntonnaett:
Thank you very much! Deprecating --deviceid sounds right. But then we should warn about it at least for one release and remove it afterwards.

I think JackTrip should fail if the selected device is not available.

IMHO we should not mention RtAudio at all in the future. I imagine a --backend option where you can choose jack, alsa, pulse, core, asio, wasapi, etc. If jack is chosen we use our own. For the rest we use RtAudio(RtAudio::API::WASAPI) or similar.

src/Settings.cpp
dyfer: Ah yes, I think so, I'll change this.
Just FYI while I tried to guard more of the RtAudio code, there might still be some RtAudio stuff in AudioInterface that's not guarded, we might want to try cleaning that up in the future.

src/jacktrip_globals.h
dyfer: Thanks for catching!

dyfer:
> Thank you very much! Deprecating --deviceid sounds right. But then we should warn about it at least for one release and remove it afterwards.

Sounds good, I'll add the warning.

 
> I think JackTrip should fail if the selected device is not available.

What kind of error should I throw?

> IMHO we should not mention RtAudio at all in the future. I imagine a --backend option where you can choose jack, alsa, pulse, core, asio, wasapi, etc. If jack is chosen we use our own. For the rest we use RtAudio(RtAudio::API::WASAPI) or similar.

I like that, let's do that in the future.

dyfer:
I've added throwing an error when the device is not found, but I've specified it as `RtAudioError::INVALID_DEVICE`, which I know means something slightly different (wrong device ID), however, conceptually that's similar IMO. Is that OK or should I throw something else?

dyfer:
I've just realized something: CLI audio device selection will break if the device name includes a comma (`,`), since I used that as a delimiter between input and output devices; however a comma in the device name is rather unlikely. Should we re-consider how parsing input/output name is done or is this fine? I'm mentioning this since any future CLI changes would be breaking changes.

ntonnaett:
We shouldn't use RtAudio's exceptions. And IIRC state of the art is throwing std::errors. Looking at `git grep throw` it looks that we mostly throw std::runtime_error() with a message. If @psiborg112 doesn't say differently I would go with that.

ntonnaett:
> I've just realized something: CLI audio device selection will break if the device name includes a comma (`,`), since I used that as a delimiter between input and output devices; however a comma in the device name is rather unlikely. Should we re-consider how parsing input/output name is done or is this fine? I'm mentioning this since any future CLI changes would be breaking changes.

As I would like to switch to cxxopts in the future, the general parsing looks fine to me. Is it possible that those names include a comma? Do we have to consider all APIs here :fearful:?

dyfer:
> As I would like to switch to cxxopts in the future, the general parsing looks fine to me. 

Sounds good.

> Is it possible that those names include a comma? Do we have to consider all APIs here 😨?

I _guess_ it's possible, but I _think_ it's unlikely.

I'm guessing it's possible since I've seen commas in PortAudio's device names, but only on Windows, only with WDM-KS, and only for Bluetooth devices (`"headset (@path,something)..."`). OTOH, my current observation with RtAudio is that the device names look pretty sane so far. On Windows they are short for WASAPI, seem standard for ASIO, but on macOS they are a bit longer as they include manufacturer for some reason (e.g. `Apple Inc.: Built-in Microphone`).

I have no idea how device names look with ALSA and PulseAudio, I haven't looked at those in the context of RtAudio (or PortAudio for that matter).

ntonnaett:
Okay. I think that excludes the comma from possible delimiters. We should look into all APIs what the possible characters are. Maybe we end up with something weird like `"`.

dyfer:
> Okay. I think that excludes the comma from possible delimiters. We should look into all APIs what the possible characters are. Maybe we end up with something weird like `"`. Edit: or `/`

Just so it's clear: this was _only_ with PortAudio and WDM-KS and bluetooth headset. On the same system, this device shows a sane name under RtAudio+WASAPI (`"Headset (<brand name>"`).

I can see `/ ` in device names already so this one is definitely a no-go. `"` would be confusing IMO (since we double quote device names already. In PortAudio/bluetooth/WDM-KS I've also seen `@`, `#`, `%`,`\`,`;`, but I think these all are really really unlikely with RtAudio...

ntonnaett:
Oh, why do we quote them? The user has to escape spaces with and without quotes.

dyfer:
> Oh, why do we quote them? The user has to escape spaces with and without quotes.

I didn't have to escape spaces, that's why I decided to go for quoting.

ntonnaett:
Is this a feature of getopt()?

dyfer:
> Is this a feature of getopt()?

I don't really know. I hoped it would work and it did. Even `"blah <space> blah","something else"` is treated as a single argument (which is what I'm using for device names now).

ntonnaett:
I see. It seems that the quoting is handled by the shell itself. You learn something new every day.

ntonnaett:
Can you omit the quotes when there is no space in the name?

dyfer:
I don't know, I have no devices without spaces...

ntonnaett:
Omitting the quotes works. And if jack is not running when listing devices it just shows me PulseAudio's.

dyfer:
I had hoped it would work without quotes. 

So is the last thing to consider that delimiter in device selection?

ntonnaett:
I think so. If I understand this correctly, the shell just removes the quotes from the string. So `"device name 1","device,name 2"` becomes a C++ string with `device name 1,device,name 2` in it. If this is the case, I don't see how we could parse this correctly. But you said that you don't know any API that we use where we would end up in this situation. So I think I'm okay with this.

Issue #295, opened 2021-05-27
Use SPDX License identifiers


ntonnaett:
https://spdx.dev/ids/

PR #390, opened 2021-08-26
Update License

dyfer:
Based on the discussion on the last meeting, here is the proposed update to our license text:
- `LICENSE.md` became `LICENSES/GPL-3.0-only.txt`
- `LICENSE_nogui` became `LICENSES/MIT.txt`
- added `LICENSES/LGPL-3.0-only.txt`
- `LICENSE.md` contains a short overview of licenses used throughout the project.

Note: I've updated the paths to comply with [REUSE](https://reuse.software/spec/), after seeing Nils's work in #394 

I'm open to any suggestions of wording of the new license file (@cchafe @mikedickey @ntonnaett  @psiborg112)

I this this should "fix" #358
#295 will help further clarify the license of individual files

EDIT:
TODO based on the discussion/links
- [x] Provide copy of license(s) with builds - binaries and application bundle
- [x] Add licenses inside macOS application bundle (`LICENSE.md` and `LICENSES/` copied to `Contents/Resources/`)
- [x] Display license in macOS installer
  - The installer now displays a Read Me screen with the contents of `README.md` and License screen with the contents of concatenated `LICENSE.md` file and the individual license files from `LICENSES/`
  - NOTE: In order to do this, I had to change package type from "raw package" to "distribution". Does this look OK @psiborg112 ?
  - These text files with fixed column width don't look that great, maybe it would be good to un-wrap lines before creating the installer...
  - [x] unwrap lines (TBD)
- [x] Display license in Windows installer
  - I am manually generating the RTF file with license text and it doesn't look great - particularly I couldn't figure out how to un-wrap the lines or remove the markdown header; but I still prefer that to having multiple license files. I'm open to suggestions about this
- [x] Add licenses to Windows install (install in Program Files)
- [x] Add copy of LGPL License
- [x] Add notice about the usage LGPL:
  - [x]  repo
  - [x]  binary distribution
  - [x] help (cli)
  - [x] about (gui)?
    - @psiborg112 in the about window I've also changed "This is beta software (...)" into "This is free software (...)". Is that OK?
- [x] Add instructions for downloading Qt sources 
  - ~~it should be _from our download location_, per [Qt FAQ](https://www.qt.io/faq/tag/qt-open-source-licensing)~~
  - ~~I haven't seen other projects using Qt providing a full download of their own...~~
  - added link to Qt repository
- [x] add `QT_OPENSOURCE` define to meson 
  - As I don't know how to check for the distribution in meson, I've added the define without checking for the license, with the appropriate note


dyfer:
I've addressed most of the issues and updated the description.

I still need to add licenses to the installers and I'm not sure what's the best way. Do we want to include licenses in the installers themselves or do we want to zip them together with files? Both approaches have their own technical caveats.

Something to consider: I've used plain text for all licenses for consistency. OK? Or do we want markdown? The original license is in plain text. 

Also what do we think about adding `.txt` extension to plaintext files? I know it's a convention to not have the extension, but it's always cumbersome to view them on macOS and Windows.

@ntonnaett I don't know how to use [this environment variable](https://github.com/jacktrip/jacktrip/pull/390/files#diff-f06d6f6734999eb5e8b10108b9e68361d1537fa93d8caa9f586534113eea6ce7R19) (and set the `#define` I've added). After reading [this](https://github.com/mesonbuild/meson/issues/9) it seems that meson doesn't like reading env vars (?) Any idea how to address this?

dyfer:
@ntonnaett I see your WIP PR puts license texts in `LICENSES/`, should I do that here? Since license location is used elsewhere (currently in docs, but I'm also thinking I'll add it to installers) it's better to put these files where they should be right away

ntonnaett:
> @ntonnaett I see your WIP PR puts license texts in `LICENSES/`, should I do that here? Since license location is used elsewhere (currently in docs, but I'm also thinking I'll add it to installers) it's better to put these files where they should be right away

Maybe it's better to do it now. The names of the files have to match what in that branch because it's all automated with REUSE.

dyfer:
All right, I think I'm done rebasing and fixing this for now... Please refer to PR description for details.
TODO: 
- add license text to installers on macOS and Windows (I'm thinking it should be displayed before the installation)
- make license files "installed" into Program Files on Windows

dyfer:
I think this PR should be ready for another review, I think I've addressed everything.

If anyone has a better idea how to handle [RTF creation](https://github.com/jacktrip/jacktrip/pull/390/files#diff-96246e12390fbd4340b9a7885724d23008cbc64fe29d2442c4cc1cdf08e32c1aR28-R38), I'd be happy to see that. Some notes on that below.

<details>

Currently, all newlines in the original text files are converted to RTF newlines (added `\line`). Additionally, paragraph markers (`\par`) are added between each of the license files.

Ideally, I think the following should happen:
- single newline should be removed, if there's no double space preceding it ("un-wrap" original text)
- double space should be replaced with newline - add `\line`
- multiple whitespace not at the end of the line should be left unchanged
- paragraph markers should be added as they are now
- empty lines should be replaced with `\line`
- markdown heading marker (`# `) should be removed
- (I might be forgetting about something still)

I couldn't figure out how to achieve all that using windows batch scripting and built-in tools. It's likely possible, but I feel I've spent more time than I wish on that already...
</details>

ntonnaett:
That became surprisingly huge. There are two small missing pieces that I see right now. The license of the limiter is GPL3 only and the gui is GPL3 or later. That's the reason there are two GPL3 license files in my reuse branch. But even in the GPLv3 only file there is a paragraph about the "or later" option that can be specified in the source code. So I think renaming to a generic `GPL-3.0.txt` should be fine.

The parsing of help2man is broken because copyright and licenses etc. are in one paragraph now. That should be easily fixable with some linebreaks.

ntonnaett:
Okay, that manpage thing didn't happen with this PR.

dyfer:
> That became surprisingly huge. 

Yeah, I know... if desired I can break this up into multiple PRs?

> The license of the limiter is GPL3 only and the gui is GPL3 or later. That's the reason there are two GPL3 license files in my reuse branch. But even in the GPLv3 only file there is a paragraph about the "or later" option that can be specified in the source code. So I think renaming to a generic `GPL-3.0.txt` should be fine.

Should I just rename the GPL file or would the text also change?

ntonnaett:
The content of GPL-3.0-only.txt and GPL-3.0-or-later.txt is the same. So renaming that file is enough.

I was just surprised of the size. The PR is fine as it is. It's one thing that gets resolved. For me this looks mergeable after the renaming. But I didn't test Windows and macOS.

dyfer:
I've renamed the file as requested.

On one had, I'd love @psiborg112 to look at this since he worked on macOS and Windows installers before... OTOH things can also be changed/fixed later if desired, I don't think I've made any changes that are hard to revert/build upon.

ntonnaett (approved):
Let's have a release candidate. :)

dyfer:
Okay, I'm going to merge this. If there's anything to be fixed, we can fix it later.

Issue #386, opened 2021-08-24
Flatpak builds


dyfer:
Currently our GitHub Actions flatpak builds are not operating properly, AFAICT. Particularly, they do not check out the exact branch/commit that they are triggered from. I also don't think these builds were widely tested.

Since FlatPak distribution doesn't seem fully tested/finished, I'd suggest to not include those in the 1.4.0 release and keep their GHA yaml files on a separate branch, but remove them from `dev` for now, so that they don't make it into `main` when we release 1.4.0. @ntonnaett what do you think about that?

If we want to use FlatPak for the release, we should further test it/fix it IMO. I'm not sure how much effort we want to put into that before 1.4.0 release though.

ntonnaett:
That there are only builds for main and dev is known. But we shouldn't trigger at all for PRs. That should be changed.

Did you try to run the flatpak builds?

Flatpaks should be delivered by flathub in the end. So they shouldn't end up on the Release page. But I think those GHAs should stay as a CI Check.

dyfer:
I think I've tried a flatpak build but I couldn't run it on Ubuntu since it expected PipeWire. I was able to install it though.

My main reservation is that the github branch is specified directly in the flatpak's yaml file, so it has no connection to what branch/commit/tag the CI is triggered from. I'd really like to see this being dynamically set by GHA runner to the current commit SHA, maybe in a similar way to how we update `org.jacktrip....desktop.in` when building.

As for when they are triggered - if FlatPak is viable to be installed directly, we could consider triggering that on PRs, similarly to what we do with regular builds, to make testing easier. 

I also think that if we want to provide FlatPak, it would be nice to also provide Snap, the latter being more ubuntu-oriented. But maybe that's not a good idea?

ntonnaett:
We would need to write a script then that generates the flatpak yaml file. There would be this scripted flatpak generator for PRs and the deployment flatpaks that automatically upload to flathub for stable and dev. Sounds good.

For the Pipewire requirement we should add a note in the metainfo description for flatpaks only. I have to think about how to detect that we build as flatpak.

dyfer:
Here are more comments I have on flatpak:
- Looking at how seamless the experience with flatpak is on Fedora, I would really like to also offer Snap (or something) for Ubuntu users
- When I opened the dev flatpak in GNOME software, the description was in German (my system language/localization setting is English)
[...]

PR #398, opened 2021-09-07
Templated flatpak generation

ntonnaett:
This change fixes the flatpak-devel GHA to work with PRs.

Maybe this can also be the basis to generate a changelog in the future.

### DIFF
--- a/linux/flatpak/org.jacktrip.JackTrip.Devel.yml.j2
+++ b/linux/flatpak/org.jacktrip.JackTrip.Devel.yml.j2
@@ -1,0 +1,30 @@
+app-id: org.jacktrip.JackTrip.Devel
+runtime: org.kde.Platform
+runtime-version: '5.15'
+sdk: org.kde.Sdk
+command: jacktrip
+finish-args:
+  # X11 + XShm access
+  - --share=ipc
+  - --socket=x11
+  - --socket=fallback-x11
+  - --device=dri
+  # Wayland access
+  - --socket=wayland
+  # Needs network access
+  - --share=network
+  # Pipewire/Jack
+  - --filesystem=xdg-run/pipewire-0
+  - --system-talk-name=org.freedesktop.RealtimeKit1
+  - --socket=system-bus
+modules:
+  - shared-modules/linux-audio/jack2.json
+  - name: jacktrip
+    buildsystem: meson
+    config-opts:
+      - -Dprofile=development
+    sources:
+      - type: git
+        url: {{ env['REPO'] }}
+        branch: {{ env['REF'] }}
+        commit: {{ env['COMMIT'] }}
--- a/.github/workflows/flatpak-devel.yml
+++ b/.github/workflows/flatpak-devel.yml
@@ -26,6 +26,6 @@ jobs:
         COMMIT: ${{ github.sha }}
     - uses: bilelmoussaoui/flatpak-github-actions/flatpak-builder@v3
       with:
-        bundle: jacktrip-dev.flatpak
+        bundle: jacktrip-dev-${{ github.sha }}.flatpak
         manifest-path: linux/flatpak/org.jacktrip.JackTrip.Devel.yml
         cache-key: flatpak-builder-${{ github.sha }}
--- a/.github/workflows/flatpak-devel.yml
+++ b/.github/workflows/flatpak-devel.yml
@@ -14,6 +14,16 @@ jobs:
     - uses: actions/checkout@v2
       with:
         submodules: true
+    - name: Generate flatpak manifest
+      uses: cuchi/jinja2-action@v1.2.0
+      with:
+        template: linux/flatpak/org.jacktrip.JackTrip.Devel.yml.j2
+        output_file: linux/flatpak/org.jacktrip.JackTrip.Devel.yml
+        strict: true
+      env:
+        REPO: https://github.com/${{ github.repository }}
+        REF: ${{ github.ref }}
+        COMMIT: ${{ github.sha }}
     - uses: bilelmoussaoui/flatpak-github-actions/flatpak-builder@v3
       with:
         bundle: jacktrip-dev.flatpak
--- a/linux/org.jacktrip.JackTrip.metainfo.xml.in
+++ b/linux/org.jacktrip.JackTrip.metainfo.xml.in
@@ -1,7 +1,7 @@
 <?xml version="1.0" encoding="UTF-8"?>
 <!-- Copyright 2021 Nils Tonnätt <nils.tonnaett@posteo.de> -->
 <component type="desktop-application">
-  <id>org.jacktrip.JackTrip</id>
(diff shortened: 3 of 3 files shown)


ntonnaett:
@dyfer If you don't know any problems with flatpak-release.yml then this should resolve #386. Can you confirm?

dyfer:
I like this approach with setting the branch (once it works).

What do you think about replacing the flatpak-release with this as well? It seems that `app-id` and config options would additionally need to be set in the template, but otherwise we would have only a single yaml template file. 
The actual GHA yaml file could also be consolidated between the devel and the release, but maybe it makes conditional run more tricky (`on:` section) so this could stay as two separate files.
Or would you rather leave the release flatpak not using a template approach?

ntonnaett:
Okay, why can't I just get two variables for what is checked out right now in GHA??? I stop for today. 

I tried `${GITHUB_REF#refs/heads/}`. Didn't work. Don't know why.

I thought if we deploy to flathub directly from the github action in the future, a separate release gha would be better.

dyfer:
> Okay, why can't I just get two variables for what is checked out right now in GHA??? I stop for today.

Yeah, I don't know, it looks like it ought to work

> I tried `${GITHUB_REF#refs/heads/}`. Didn't work. Don't know why.

Ah, to use that directly in the yaml config (as opposed to the script itself, i.e. in the "run" section), you'd need to use `${{ github.ref }}` but then I don't know if you can trim `refs/heads` from it (maybe that's not needed though?)

> I thought if we deploy to flathub directly from the github action in the future, a separate release gha would be better.

I see, good point.
One conceptual issue with the release flatpak is that it should be built from a defined tag, not from the tip of `main` (while we will aim to only push final releases to main, the actual release is tied to the tag, not to the tip of the branch). I don't know how this should be solved and how flatpak ecosystem deals with that.

ntonnaett:
Using the commit sha seems to work :)

ntonnaett:
SHA doesn't work when force-pushing... So I tried github.ref now. This seems to work for force-push. Hope this works out now.

dyfer:
> SHA doesn't work when force-pushing... So I tried github.ref now. This seems to work for force-push. Hope this works out now.

Weird, but I see this issue happens elsewhere... seems maybe to do with caching, but I don't understand on what stage (github servers? action runner? the action itself?) If you don't have time to investigate the SHA approach (which IMO would be ideal), this looks fine as is.

I have some comments on flatpak that I'll post in #386

dyfer:
Aside from SHA vs REF, I have questions/comments about few more details:
- what does the `development` profile do? Is this a debug build? It would be nice to document it somewhere (not necessarily in this PR)
- Could the flatpak name be changed to match other builds, to something like `JackTrip-<SHA>-flatpak` or `JackTrip-<SHA>-flatpak-dev` (the non-dev version wouldn't have the SHA in the name so mabe "-dev" is not needed?)
  - ideally the sha would be a "short" sha (I want to change this for other builds as well) but it's a bit cumbersome to do that, so I can change it later from regular sha to short when I make the change to other builds

ntonnaett:
The profile is needed to have different application ids for stable and dev. Maybe we want a third, that doesn't end up in flathub.

linux/flatpak/org.jacktrip.JackTrip.Devel.yml.j2
```diff
@@ -0,0 +1,29 @@
...
+  # Pipewire/Jack
+  - --filesystem=xdg-run/pipewire-0
+  - --system-talk-name=org.freedesktop.RealtimeKit1
+  - --socket=system-bus
+modules:
+  - shared-modules/linux-audio/jack2.json
+  - name: jacktrip
+    buildsystem: meson
+    config-opts:
+      - -Dprofile=development
```
dyfer, line 25: My question was what does this flag do to the build? Could this be documented somewhere?

ntonnaett: It's setting the app id to org.jacktrip.JackTrip.Devel instead of org.jacktrip.JackTrip, so that you can install the stable and dev versions side by side. This should probably be documented somehow.

dyfer: I was asking about `-Dprofile=development`, isn't that a build/meson setting?

ntonnaett: Yes. That's what I'm talking about. It's called profile because it's common in other projects. And I don't like the idea of setting the app id as a string. What's confusing you about it?

dyfer: Sorry, maybe I should've looked into meson source instead of asking you - after looking there I understand what's happening. The connection between `-Dprofile=development` and application id is definitely unclear unless you know what's in `meson.build`.

In case anyone else is looking at this: my original question was what does the "development" profile change in the build system. The answer is that it changes `application_id`, used in the installation step in the Linux desktop integration files; it does _not_ change anything in the cpp build itself, i.e. does not trigger a debug build.

dyfer:
This looks good, I just wanted to possibly consider two things:
- is it really not possible to use SHA? if not, that's okay, but it would really be optimal.
- would you consider changing the flatpak name? I can also make a PR later for this.

linux/flatpak/org.jacktrip.JackTrip.Devel.yml.j2
ntonnaett: No problem. Generally, if you don't know what configure options do, you can just type `meson configure` in your build directory:
```
  Project options                          Current Value                              Possible Values                            Description
  ---------------                          -------------                              ---------------                            -----------
  jack                                     auto                                       [enabled, disabled, auto]                  Build with JACK Backend
  nogui                                    false                                      [true, false]                              Build without graphical user interface
  profile                                  default                                    [default, development]                     Choose build profile / Sets desktop id accordingly
  rtaudio                                  enabled                                    [enabled, disabled, auto]                  Build with RtAudio Backend
  wair                                     false                                      [true, false]                              WAIR
  weakjack                                 false                                      [true, false]                              Weak link JACK library
```

So maybe the description for profile could be improved. What do you think?

ntonnaett:
What is the benefit of using the sha?

dyfer:
> What is the benefit of using the sha?

It identifies the exact commit that the build was triggered from. While not likely, the branch could've been altered between the time the build is triggered and when the action actually checks is out.

ntonnaett:
I decided to use both. And now force push worked. Maybe it's just random. Added the whole sha to the filename too. Maybe we can shorten it in the future.

dyfer (approved):
Looks great, thank you!

dyfer:
It looks like specifying SHA might not have worked in the end? https://github.com/jacktrip/jacktrip/runs/3571966360?check_suite_focus=true

PR #400, opened 2021-09-09
Dev plc final

cchafe:
different approach to the poolbuffer -- results are getting more promising
only tested on linux / jack
will get some more updates before version 1.4 no doubt


ntonnaett:
Do you want to elaborate on the new approach? Might be quite interesting.

Could you run clang-format over it, but mark everything that shouldn't be formatted (maybe some comments) with:
```c++
// clang-format off
// comment
// some_code() {
//     intended = 0;
// }
// clang-format on
```

ntonnaett:
The use of macros makes it really hard to comprehend. PACKETSAMP looks like you actually want something like std::accumulate that you could easily parallelize.
There are a lot of changes in other files than PoolBuffer.* that probably would go away when clang-formatting them.

ntonnaett:
I think plc-test-tmp should go into the tests directory and the tmp could probably be removed.

ntonnaett:
Every new should have a corresponding delete. In most cases it's better to not allocate that way at all but define and initialize them as members.

src/JackTrip.h
```diff
@@ -300,6 +300,10 @@ class JackTrip : public QObject
     {
         mDataProtocolReceiver = DataProtocolReceiver;
     }
+    virtual int getBufferStrategy()
```
ntonnaett, line 303: =>  int getBufferStrategy() const

ntonnaett: and noexcept

cchafe:
Thanks -- PACKETSAMP is written out longhand as **for (int s = 0; s < mFPP; s++)** -- simply iteration over a packet -- at this stage that can be expanded.
I forgot to remove the plc-test-tmp directory, will do that.
Ditto, deletes!
I'm not sure how to avoid new when doing the inner part of a 2D array, for example,
vector<int8_t*> mIncomingDat -- how to create the storage for the int8 arrays if doing that in the class definition?
for (int i = 0; i < mMaxPoolSize; i++) {
        int8_t* tmp = new int8_t[mBytes];
        mIncomingDat.push_back(tmp);
    }

src/PoolBuffer.cpp
```diff
@@ -32,42 +32,71 @@
...
 using std::endl;
 using std::setw;
 
-#define STDDEVINDOW     200  // packets
-#define STDDEV2POOLSIZE 30.0
-#define MAXPOOLSIZE     6000  // insanely large pool, 2 sec FPP16
-#define HIST            6     // at FPP32
-#define OUT(x, ch, s)   sampleToBits(x, ch, s)
+#define HIST          6  // at FPP32
+#define OUT(x, ch, s) sampleToBits(x, ch, s)
```
ntonnaett, line 88: HIST can be a constexpr member variable. OUT doesn't gain anything.

cchafe: thanks!

ntonnaett:
> Thanks -- PACKETSAMP is written out longhand as **for (int s = 0; s < mFPP; s++)** -- simply iteration over a packet -- at this stage that can be expanded.
> I forgot to remove the plc-test-tmp directory, will do that.
> Ditto, deletes!
> I'm not sure how to avoid new when doing the inner part of a 2D array, for example,
> vector<int8_t*> mIncomingDat -- how to create the storage for the int8 arrays if doing that in the class definition?
> for (int i = 0; i < mMaxPoolSize; i++) {
> int8_t* tmp = new int8_t[mBytes];
> mIncomingDat.push_back(tmp);
> }

Theoretically you could just use std::vector<std::vector<int8_t>>. But performance is important. The current approach and those nested std::vectors would allocate all over the place and using this leads to a lot of jumping in memory. Allocating the memory one dimensional would be faster. 

I have to do some research. Yesterday I've read about mspan when I looked in to Eigen3 and Blaze. Because using one dimensional arrays as twodimensional can lead to ugly bugs too.

src/JackTrip.h
cchafe: ...thanks, done. What should the noexcept syntax look like for this?

ntonnaett: int getBufferStrategy() const noexcept

ntonnaett: I think the correct order is cv specifier, reference, exception, (override/final)

ntonnaett:
Aren't those tests worth keeping?

src/PoolBuffer.cpp
```diff
@@ -353,46 +352,42 @@ void PoolBuffer::processChannel(int ch, bool glitch, int packetCnt, bool lastWas
...
+            //                                ( (!ch) ? cd->mPrediction[s] : (
+            //                                (s)?0.0:-0.2) )
+            //                              :
+            //                                ( (!ch) ? ( (lastWasGlitch) ?
+            //                                cd->mXfadedPred[s] : cd->mTruth[s] )
+            //                                        : cd->mTruth[s]),
+            //                            ch, s);
+
+            for (int s = 0; s < mFPP; s++)
+                sampleToBits((glitch)
```
ntonnaett, line 369: Is this double loop correct?

cchafe: nope, good catch

src/PoolBuffer.h
```diff
@@ -77,25 +83,29 @@ class ChanData
 class StdDev
 {
    public:
-    StdDev(int w);
+    StdDev(int w, int id);
     void reset();
-    void tick();
+    double tick();
     QElapsedTimer* mTimer;
```
ntonnaett, line 89: If you remove the asterisk, you shouldn't need any new.

src/PoolBuffer.h
```diff
@@ -142,18 +155,35 @@ class PoolBuffer : public RingBuffer
...
-    unsigned int mIncomingCnt;
+    int mIncomingCnt;
     vector<int> mIndexPool;
     int mQlen;
     int mGlitchCnt;
     int mGlitchMax;
     vector<ChanData*> mChanData;
+    vector<sample_t> mTail;
     StdDev* stdDev;
+    StdDev* stdDev2;
```
ntonnaett, line 172: StdDev stdDev{ (int)(floor(48000.0 / (double)mFPP)), 1 };

src/PoolBuffer.h
cchafe: done, good catch here, too

src/JackTrip.h
ntonnaett: Would you remove the virtual? Or is there a reason for it?

src/PoolBuffer.cpp
```diff
@@ -84,13 +112,14 @@ PoolBuffer::PoolBuffer(int /*sample_rate*/, int channels, int bit_res, int FPP,
         mBitResolutionMode = AudioInterface::audioBitResolutionT::BIT32;
         break;
     }
-    mHist            = 6 * 32;                // samples, from original settings
+    mMinPoolSize     = mPoolSize;
```
ntonnaett, line 115: It looks like mMinPoolSize and mMaxPoolSize are superfluous now.

src/PoolBuffer.cpp
```diff
@@ -104,119 +133,163 @@ PoolBuffer::PoolBuffer(int /*sample_rate*/, int channels, int bit_res, int FPP,
     }
     mLastWasGlitch = false;
     mOutgoingCnt   = 0;
-    for (int i = 0; i < mPoolSize; i++) {
+
+    mMaxPoolSize = mPoolSize;
+    for (int i = 0; i < mMaxPoolSize; i++) {
         int8_t* tmp = new int8_t[mBytes];
         mIncomingDat.push_back(tmp);
     }
```
ntonnaett, line 141:
```suggestion
    for (int i = 0; i < mMaxPoolSize; i++) {
        mIncomingDat.emplace_back(new int8_t[mBytes]);
    }
```

Let's stay with this approach for now, but add a delete loop to the destructor. I have a solution for this in mind with a std::array that is wrapped in a template class, so that you can do `array[m][n]`.

ntonnaett: And add `mIncomingDat.resize(mMaxPoolSize);` before the loop, so we don't reallocate all the time.

src/PoolBuffer.cpp
cchafe: yup, removed

ntonnaett: Great!

src/PoolBuffer.h
ntonnaett: Thanks!

dyfer:
A bystander comment: there's quite a bit of whitespace change in this PR. Do we want to put these in? E.g. changes like [this](https://github.com/jacktrip/jacktrip/pull/400/files#diff-3b399c558dca5ac7f9b8abc5fbb969e06444cc0647558250b6e2ac062a7ad4ccR82) or [this](https://github.com/jacktrip/jacktrip/pull/400/files#diff-3b399c558dca5ac7f9b8abc5fbb969e06444cc0647558250b6e2ac062a7ad4ccR422)

ntonnaett:
> A bystander comment: there's quite a bit of whitespace change in this PR. Do we want to put these in? E.g. changes like [this](https://github.com/jacktrip/jacktrip/pull/400/files#diff-3b399c558dca5ac7f9b8abc5fbb969e06444cc0647558250b6e2ac062a7ad4ccR82) or [this](https://github.com/jacktrip/jacktrip/pull/400/files#diff-3b399c558dca5ac7f9b8abc5fbb969e06444cc0647558250b6e2ac062a7ad4ccR422)

That's what I've meant with missing clang-format.

cchafe:
(converted to draft)
I found a better non-pool way, so I'd like to refactor and call it something like PacketDispatcher not that that rolls off the tongue.

cchafe:
Working well, PoolBuffer renamed to Regulator, clang-formatted, dev merged. I think it's ready.
-q is usually in terms of packets but here for --bufstrategy 3, it is now in terms of msec of jitter tolerance (integer number).
Running with -I 1 will print the max interpacket arrival interval and -q should be chosen accordingly.
Briefly, the UDP receiver calls Regulator::pushPacket on every new arrival. It is then timestamped and stored in a slot.
The JACK audio callback calls Regulator::pullPacket which returns an audio buffer. That will contain either
-- zeros if it's too early in the connection
-- the oldest packet within the -q msec tolerance range
-- a cross-faded packet bridging an overrun
-- a predicted packet filling an underrun

TODO's are broadcast output, mixed FPP's, but that's after 1.4

PR #401, opened 2021-09-10
Metainfo summary fix

ntonnaett:
Removes the German translation that doesn't work as expected and Remove some capitalization (https://blogs.gnome.org/tbernard/2021/09/07/ready-for-software-41/).


dyfer (approved):
Thanks!

PR #404, opened 2021-09-11
GitHub Actions: use short SHA in artifact names

dyfer:
This changes artifact names to use short SHA (first 7 digits)
I've also attempted to change the name for Flatpak dev build. @ntonnaett is the capitalization OK? I wanted to make it consistent with other artifacts.

### DIFF
--- a/.github/workflows/flatpak-devel.yml
+++ b/.github/workflows/flatpak-devel.yml
@@ -23,8 +23,10 @@ jobs:
       env:
         REPO: https://github.com/${{ github.repository }}
         REF: ${{ github.ref }}
+    - name: Add SHORT_SHA environment variable
+      run: echo "SHORT_SHA=`echo ${GITHUB_SHA::7}`" >> $GITHUB_ENV
     - uses: bilelmoussaoui/flatpak-github-actions/flatpak-builder@v3
       with:
-        bundle: jacktrip-dev-${{ github.sha }}.flatpak
+        bundle: JackTrip-Devel-${{ env.SHORT_SHA }}.flatpak
         manifest-path: linux/flatpak/org.jacktrip.JackTrip.Devel.yml
         cache-key: flatpak-builder-${{ github.sha }}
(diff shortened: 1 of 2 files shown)


dyfer:
Also @ntonnaett would would be think about renaming flatpak to `JackTrip-<commit>-dev`?

.github/workflows/flatpak-devel.yml
```diff
@@ -23,8 +23,10 @@ jobs:
       env:
         REPO: https://github.com/${{ github.repository }}
         REF: ${{ github.ref }}
+    - name: Add SHORT_SHA environment variable
+      run: echo "SHORT_SHA=`echo ${GITHUB_SHA::7}`" >> $GITHUB_ENV
```
ntonnaett, line 27: Why do you write into a file? Shouldn't  something like the following work?
```bash
export SHORT_SHA=`echo ${GITHUB_SHA::7}`
```

dyfer: No, with github actions the env variables set with `export` do not stick to the next step. Adding them the `$GITHUB_ENV` is the way to make them available later.
An alternative to this syntax would be to use `set-output` functionality like we do [here](https://github.com/jacktrip/jacktrip/blob/dev/.github/workflows/jacktrip.yml#L99) but I think what I proposed is a bit shorter (we don't need to specify step id). 
However, if you prefer, we could do `set-output`, it's not that different.

ntonnaett: Interesting.

ntonnaett:
Personally I would prefer the naming to be `JackTrip-Devel-commit.flatpak`, because I see Devel more to be part of the name. If you have a lot of those files in one directory it would also be easier to search for the right category.

.github/workflows/flatpak-devel.yml
dyfer: Any preference on the current approach vs `set-output`?

ntonnaett: I prefer the current approach.

ntonnaett (approved):
Thanks!

PR #405, opened 2021-09-12
QMake macOS: add "visibility=hidden" to linker flags for static builds

### LINKED ISSUE #393: GHA: macOS linker warnings
See: https://github.com/jacktrip/jacktrip/runs/3438184759?check_suite_focus=true#step:14:118

dyfer:
Should fix #393, but it doesn't seem to.


dyfer:
This doesn't seem to solve #393 so I'm closing it for now.

PR #406, opened 2021-09-15
Bundle RtAudio in QMake

dyfer:
I would like us to consider to re-bundle RtAudio with JackTrip, like we had around v.1.1. The main goal is to build RtAudio on Windows (to avoid cumbersome library installation), however it seems it does build on all 3 platforms. Of course it is still possible to use system library. Also, I know meson downloads RtAudio, but getting meson environment is also not that straightforward on Windows - QMake/MinGW is still the easiest to install.

This PR attempts to bring back building RtAudio into qmake, with following changes:
- RtAudio is updated to the current version
- RtAudio is a proper submodule now
- I'm not 100% sure about my qmake implementation (I'm not using a build script that comes with RtAudio)
  - RtAudio is still built somewhat separately from the rest of the project, using QMake's ["add compiler" functionality](https://stackoverflow.com/questions/27683777/how-to-specify-compiler-flag-to-a-single-source-file-with-qmake)
  - on Linux and macOS RtAudio uses the same compiler flags as the rest of the project, which is probably fine
  - on Windows it uses different flags, since RtAudio doesn't build if `..._WINNT=` and `WIN32...MEAN` are set
- RtAudio builds on Linux, but I haven't tested if this runs. I think this is something to investigate after releasing 1.4
  - My idea is that for the static binary that we release we could add RtAudio (if it performs adequately with PulseAudio or Alsa). In that case RtAudio could be built without Jack, so that we can keep using weakjack and have jack as an optional requirement also on Linux. I'm fine if we don't want to do this, it's just an option. And again, this should wait after 1.4, as it should be well tested before releasing it into the wild.

For now I've added qmake builds to GitHub Actions next to our regular builds; eventually I think we should replace Windows GHA build from using VCPKG's RtAudio to use bundled RtAudio. We don't have to change macOS build, but we could (?)

macOS build runs, but I haven't tested it beyond that. I haven't tried running Linux build. ~~Windows build seems to not run (I thought it had, but now I can't run it anymore)~~ It seems that it's an issue with RtAudio. Using an older commit of RtAudio fixes the Windows build.

The PR has also some housekeeping commits, if desired I can take them out and put them in a separate PR.

TODO:
- [ ] test Windows build
  - ~~figure out why it doesn't run...~~
- [ ] test macOS build 
- [ ] test Linux build
- [ ] review if my implementation of qmake build of RtAudio is clean/proper enough
- [ ] clean up GitHub Actions
  - [ ] decide if we leave both bundled and system rtaudio builds for all platforms (for testing, no artifact upload)
  - [ ] change the default Windows build to use bundled RtAudio
- [ ] optionally update qmake build docs (these don't document the build script well anyway, so maybe that should be a separate task/PR)
- [ ] have meson use the rtaudio submodule instead of downloading it, if that's OK with @ntonnaett (?)

### DIFF
--- a/.github/workflows/jacktrip.yml
+++ b/.github/workflows/jacktrip.yml
@@ -172,10 +172,10 @@ jobs:
         if: runner.os == 'Windows'
         shell: bash
         run: |
-          choco install jack --version=1.9.17
-          choco install openssl
+          choco install jack --version=1.9.17 --no-progress
+          choco install openssl --no-progress
           if [[ "${{ matrix.rtaudio }}" == true ]] && [[ "${{ matrix.build-rtaudio }}" != true ]]; then 
-            choco install pkgconfiglite
+            choco install pkgconfiglite --no-progress
             vcpkg install rtaudio[asio] --triplet=x64-mingw-static
             echo "PKG_CONFIG_PATH=$VCPKG_INSTALLATION_ROOT\installed\x64-mingw-static\lib\pkgconfig" >> $GITHUB_ENV
           fi
--- a/.github/workflows/jacktrip.yml
+++ b/.github/workflows/jacktrip.yml
@@ -36,6 +36,15 @@ jobs:
             qt-static-cache-key: 'v22' # required for static qt; update this to force rebuilding static Qt
             jacktrip-path: jacktrip
             binary-path: binary
+            
+          - name: Linux-x64-bundled-rtaudio
+            runs-on: ubuntu-18.04
+            rtaudio: true
+            build-rtaudio: true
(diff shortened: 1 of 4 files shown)


dyfer:
Small update here: It seems that the issue with Windows build is actually an issue with RtAudio. Reverting to an older version of RtAudio "fixes" it. However, after yesterday's dev meeting I'm investigating an alternative way to build RtAudio - by creating a separate .pro file and including the whole directory in jacktrip.pro. I'll probably submit a new PR when ready and we can compare.

dyfer:
Closing in favor or https://github.com/jacktrip/jacktrip/pull/415

PR #411, opened 2021-09-17
Update installer scripts

psiborg112:
Add an option to specify the binary location in the MacOS packaging script to allow easier signing of downloaded github artifacts.

Allow upgrades with Windows installer.

Use a smart pointer for the AudioTester so that we don't need to keep the Settings object around.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -3,8 +3,9 @@ set(CMAKE_OSX_DEPLOYMENT_TARGET 10.13)
 project(QJackTrip)
 
 set(nogui FALSE)
-set (rtaudio TRUE)
-set (weakjack TRUE)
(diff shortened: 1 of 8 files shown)


dyfer:
Thanks @psiborg112 
My personal preference would be to have separate commits for cpp changes and installer/script changes, but it's not a big deal.
I can confirm that with this PR one can install Windows version over the previous one.

dyfer (approved):
Thanks!

dyfer:
@cchafe since this fixes an issue with Windows installer, maybe we should merge this before making rc.5?

psiborg112:
@dyfer yeah, usually I'd split them out, but just didn't have time in this case.

PR #415, opened 2021-10-09
Bundle RtAudio in QMake v4

dyfer:
This is a variation on #414, with following differences:

- submodule is in `externals/rtaudio/`, so one level further than other submodules
- a separate .pro file is at `rtaudio.pro`, i.e. in the top directory


ntonnaett:
This shouldn't be mixed with meson. You can use `meson subprojects download` if you want to bundle all subprojects with jacktrip.

cchafe (approved):

PR #344, opened 2021-06-12
Formatting everything before release 1.4.0

ntonnaett:

PR #416, opened 2021-10-09
bumped to rc.6, updated copyrights

cchafe:
… to 2021 where appropriate and changed all 2008 notices to be 2008-2021


ntonnaett:
What's wrong with those changes?

ntonnaett:
There were changes after #344.

dyfer:
I see, thanks for explanation.
It would be beneficial to document the formatting workflow: what's expected of the contributors, how to configure it etc. That documentation is currently lacking.

And once we agree that everything is formatted, we should definitely add a GH action to check for it (not necessarily to correct it, but that's definitely possible as well). In the future we could have one automatic GH action to check formatting (for every PR) and one on-demand action to format the PR (I have no experience with the latter approach)

PR #420, opened 2021-10-29
Updated a few defaults for the GUI

mikedickey:
Updated a few defaults for the GUI that will work better for the vast majority of users:

* Enable auto jitter buffer
* Hub Client mode instead of P2P Client


cchafe (approved):

cchafe:
builds and works, good stuff

PR #421, opened 2021-10-29
[RtAudio] Set mAudioBufferSize in jacktrip after RtAudio setup

ntonnaett:
#419 might not have been enough to fix the possible buffer size mismatch with RtAudio (and ASIO) because the jacktrip class has its own buffersize variable. This should fix it.


mikedickey (approved):

mikedickey:
It worked for me!

PR #422, opened 2021-10-30
Moving authentication options to advanced tab

mikedickey:
Moving username/password fields for authentication to advanced tab, since most users will likely assume this is asking for the Virtual Studio credentials.


cchafe (approved):
didn't test -- but should be fine

PR #428, opened 2021-11-04
Add clang-format check to Github Actions

ntonnaett:
* adds clang-format Github Action
* style change: add more linebreaks for better debuggability
* add .clang-format-ignore file and add externals directory

### DIFF
--- a/.clang-format
+++ b/.clang-format
@@ -16,10 +16,10 @@ AllowAllParametersOfDeclarationOnNextLine: true
 AllowShortEnumsOnASingleLine: true
 AllowShortBlocksOnASingleLine: Empty
 AllowShortCaseLabelsOnASingleLine: false
-AllowShortFunctionsOnASingleLine: All
-AllowShortLambdasOnASingleLine: All
-AllowShortIfStatementsOnASingleLine: WithoutElse
-AllowShortLoopsOnASingleLine: true
+AllowShortFunctionsOnASingleLine: Inline
+AllowShortLambdasOnASingleLine: Empty
+AllowShortIfStatementsOnASingleLine: Never
+AllowShortLoopsOnASingleLine: false
 AlwaysBreakAfterDefinitionReturnType: None
 AlwaysBreakAfterReturnType: None
 AlwaysBreakBeforeMultilineStrings: true
--- a/.clang-format-ignore
+++ b/.clang-format-ignore
@@ -1,0 +1,1 @@
+externals/*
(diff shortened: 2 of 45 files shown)


dyfer:
I know this is a draft, here are some thoughts for consideration:
- I like the idea of having clang-format check
- I'm not sure if I like it in a separate file - we could have it as part of the main .yaml build matrix as the first job; but that could be either way really
- I'm hesitant to pull in a not-very-widely-used action for that, especially that it's in the "alpha" version. Could we do this using our own script instead?

ntonnaett:
The reason why this is a draft is only because my local clang-format on Fedora 35 produces something different than what both clang-format GHAs produce that I tried. I don't know why.
But I actually would prefer: https://github.com/marketplace/actions/clang-format-check

Including this into the main workflow might be better. Reduces work.

dyfer:
Thanks for the explanation

> The reason why this is a draft is only because my local clang-format on Fedora 35 produces something different than what both clang-format GHAs produce that I tried. I don't know why. 

My impression is that different versions of clang-format produce slightly different results (so if we want to enforce formatting, we need to stick to the same version). Or are you getting different results with the same version of clang-format?

> But I actually would prefer: https://github.com/marketplace/actions/clang-format-check

Hm, that looks better I think, but still I wonder if doing this ourselves wouldn't be better? I just want to avoid pulling in actions that for some reason fail and stop being supported, or end up being a security concern down the road. But maybe my worries are not warranted :)

> Including this into the main workflow might be better. Reduces work.

Sounds good

ntonnaett:
Both are version 13 but on Fedora it's a rc. So maybe this is already fixed.

I think we could just run run-clang-format.py with the right flags instead of pulling in another action.

ntonnaett:
I actually don't like that our clang-format style allows putting scopes in one line (after if, function(), etc.pp.). I'll change that.

ntonnaett:
* DoozyX's GHA is better (although marked as pre-release) and I don't want to maintain my own clang-format GHA.
* I've learned that additional linebreaks can help debugging and changed our style because of that.
* IMHO the jacktrip.yml is already big enough. I like the format checker being in another file.

psiborg112:
Outside of the Github actions changes, there are only formatting changes to the code, right?

ntonnaett:
Yes

psiborg112:
Then I think this looks good to go.

psiborg112 (approved):

PR #429, opened 2021-11-04
update headline to mention "Core"

forsythac:
improves consistency with the jacktrip.org landing page, which call this page "Core"

### DIFF
--- a/docs/index.md
+++ b/docs/index.md
@@ -1,4 +1,4 @@
-# Welcome to JackTrip
+# Welcome to JackTrip Core Software
 
 JackTrip is a Linux, macOS, or Windows multi-machine audio system used for network music performance over the Internet.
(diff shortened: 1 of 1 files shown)


dyfer:
I like the consistency but maybe we should also change this elsewhere in this project? AFAICT "... Core" is not used anywhere in the application itself. If we want to be consistent, it maybe should be...? (places to include it would be command line prompt `-v` and "about" in the gui, possibly elsewhere). What do others think?

ntonnaett:
JackTrip Core is not the name of the software but one part of jacktrip.org. For me this change doesn't make sense, especially if you are a jacktrip user that doesn't belong to the jacktrip.org community. Maybe JackTrip Core should link to https://github.com/jacktrip instead?

PR #431, opened 2021-11-10
Fix for OSX installer failures

### LINKED ISSUE #430: 1.4.0 osx installer fails when missing /usr/local/bin
Running the OS X installer on Big Sur (11.6) Intel. After it reaches the last step and seems to be running some package verification step it displays the message: "The Installation Failed" (see screenshot attached).

Despite the error message, the installer actually seems to have worked correctly. JackTrip is installed in the Applications folder, and appears to work just fine.

jacktrip_installation_failed
..

mikedickey:
Fixes #430

create /usr/local/bin directory if it doesn't already exist

### DIFF
--- a/macos/package/link.sh
+++ b/macos/package/link.sh
@@ -1,4 +1,5 @@
 #!/bin/sh
 
-rm /usr/local/bin/jacktrip
+mkdir -p /usr/local/bin
+rm -f /usr/local/bin/jacktrip
 ln -s "$2"/Contents/MacOS/jacktrip /usr/local/bin/jacktrip
--- a/src/jacktrip_globals.h
+++ b/src/jacktrip_globals.h
@@ -44,7 +44,7 @@
 /// \todo Add this namespace
 // namespace JackTrip
 
-constexpr const char* const gVersion = "1.4.0";  ///< JackTrip version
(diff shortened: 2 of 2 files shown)


ntonnaett:
Could you bump the version to 1.4.1?

ntonnaett:
In #430 you used sudo. Isn't that needed for the installer?

mikedickey:
> In #430 you used sudo. Isn't that needed for the installer?

I'm not sure, but was assuming that since the script was executing successfully without sudo before (assuming the directory existed), the installer must already be running it with elevated permissions.

dyfer (approved):
I tested this on a Big Sur installation with non-existing `/usr/local/bin` and can confirm that the destination folder was created correctly. The installer itself is elevating privileges with a GUI prompt before performing the installation.

dyfer:
The only outstanding question is whether we merge this into `main` or create a separate release branch. I'm happy to go either way, but I think that merging to main might be easier (if we are not planning a long "patch release" cycle). Does anyone have objections agains merging this into `main` directly? (@cchafe @ntonnaett)

ntonnaett:
I would prefer merging this directly.

cchafe (approved):

PR #434, opened 2021-11-22
Verbose mode in GUI

psiborg112:
Enable the verbose flag to be set from the command line while using the GUI. (jacktrip --gui -V)

Also detect whether the program has been run from the command line on windows so that we don't detach ourselves from the console. (Otherwise we won't get any output on the terminal.)

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -5,6 +5,7 @@ project(QJackTrip)
 set(nogui FALSE)
 set(rtaudio TRUE)
 set(weakjack TRUE)
+add_compile_definitions(NO_JTVS)
 add_compile_definitions(QT_OPENSOURCE)
 
 if (nogui)
@@ -111,6 +112,7 @@ if (NOT nogui)
     src/gui/qjacktrip.cpp
     src/gui/about.cpp
     src/gui/messageDialog.cpp
+    src/gui/textbuf.cpp
     src/gui/qjacktrip.qrc
   )
 
--- a/jacktrip.pro
+++ b/jacktrip.pro
@@ -232,7 +232,8 @@ HEADERS += src/JackAudioInterface.h \
 !nogui {
 HEADERS += src/gui/about.h \
            src/gui/messageDialog.h \
-           src/gui/qjacktrip.h
+           src/gui/qjacktrip.h \
+           src/gui/textbuf.h
 }
 
 rtaudio|bundled_rtaudio {
@@ -270,7 +271,8 @@ SOURCES += src/JackAudioInterface.cpp \
 !nogui {
 SOURCES += src/gui/messageDialog.cpp \
            src/gui/qjacktrip.cpp \
-           src/gui/about.cpp
+           src/gui/about.cpp \
+           src/gui/textbuf.cpp
 }
 
 !nogui {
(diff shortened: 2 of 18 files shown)


psiborg112:
There may be some sort of versioning issue with RtAudio that's causing the meson builds to fail. (No changes were made to RtAudioInterface.cpp but that's where the problem lies.)

cchafe:
(this is named "vebose" -- but no need to fix the typo)
The version number reports 1.4.0 so might need to merge dev to this first.
Works great in linux.

psiborg112:
Was driving me insane trying to find out where the typo was until I realised it was just in the branch name. (Will just keep the typo so I don't have to recreate the pull request.) Have merged dev into this. (Must not have updated my local copy of the dev branch before basing this off of it.)

ntonnaett:
Could we have a checkbox for this in the GUI?
I'll fix the Meson RtAudio fail.
Do we want to start with the rapid releases and make this version 1.5.0 because it's a feature?

ntonnaett:
Meson build is fixed in #436

psiborg112:
I was thinking I would hold off from adding a checkbox for this in the GUI just yet because the program has to be run from the command line for the diagnostic information to be seen. So for most people running the app from the dock or start menu, it's a checkbox that would just do nothing other than being confusing. Ideally, the next step here would be redirecting std::cout to a window, but that's going to be a little more involved. (That change would probably be worthy of a release as a new feature, but not this one.)

psiborg112:
And it's accessible from the GUI now.

ntonnaett:
Nice. So this becomes 1.5.0 in the end?! Would we need a release candidate phase for this? Should we release 1.4.2 now?

Why do we need the JTVS gui? Shouldn't the macro be reversed? JTVS instead of NO_JTVS?

ntonnaett:
Just read the commit message. So it's about the login credentials. Maybe we could have this in one gui? So you can choose between JackTrip native and Virtual Studio login.

psiborg112:
Maybe. Definitely worth considering. I think this will work for the moment though.

psiborg112:
(Will need to find out more about how the virtual studio login works first. But it would make sense to be able to just select between the two of them in the same group box.)

psiborg112:
And maybe this should just become a 1.4.2 release now if others are happy with that?

ntonnaett:
I wondered why there are those redundant commits. But dev is behind main. So I will fast forward dev to main first.

cchafe:
releasing this as 1.4.2 sounds good -- I'll wait a day in case there's any more feedback

cchafe:
I built it on FC34 and got 3 warnings for integer comparisons in gui/messageDialog.cpp
lines 42, 79, 87
different signedness: ‘int’ and ‘quint32’
That's all I see -- works great.

psiborg112:
Let me quickly fix that then. (I assumed that the QVector size would be an unsigned integer, but it's not.)

cchafe:
Looks like there's one more showing with the same qmake compile:
src/gui/messageDialog.cpp:42:26: warning: comparison of integer expressions of different signedness: ‘qint32’ {aka ‘int’} and ‘quint32’ {aka ‘unsigned int’} [-Wsign-compare]
   42 |     for (qint32 i = 0; i < streamCount; i++) {

psiborg112:
Oops. Typo. There should have been a u after that q. Thanks for picking that up.

cchafe (approved):

PR #439, opened 2021-11-26
Help screen for app bundle script

psiborg112:


cchafe (approved):
not something I can test but assume it works
I figure it should make it into 1.4.2

PR #441, opened 2021-12-05
[GHA] Fix weekly cron job entry

ntonnaett:
I just recognized that we have gha builds every minute right now. So I thought there might be someone who's tampering with our stuff. But our cron job entry is just wrong. Right now every minute on sundays a jacktrip build is started. I think one time a week instead of 1440 times is sufficient.

https://crontab.guru/#0_0_*_*_0


cchafe (approved):
good catch

dyfer:
Hahaha sorry about that! Thanks for catching and fixing it.

PR #442, opened 2021-12-18
Generate release data from yaml file for MkDocs and AppStream

ntonnaett:
Release data in the AppStream Metainfo file is a requirement for adding JackTrip to the Flathub repository. This PR moves the changelog data to a yaml file from where the changelog.md for Mkdocs and the metainfo file for flatpak are propagated.

PR #443, opened 2021-12-18
bump version and a couple other edits to jacktrip_globals.h

cchafe:
the name says it

### DIFF
--- a/src/jacktrip_globals.h
+++ b/src/jacktrip_globals.h
@@ -30,7 +30,7 @@
 //*****************************************************************
 
 /**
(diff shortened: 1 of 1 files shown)


ntonnaett:
Let's merge this after #442 and then make a release!

cchafe:
Yup, let's do that and if should then go all the way with flatpak

ntonnaett:
For 1.4.3 adding JackTrip to Flathub should be possible.

ntonnaett (approved):

PR #453, opened 2022-01-03
Comply with macro rules

### LINKED ISSUE #276: Use of compiler reserved names

ntonnaett:
Fixes #276 

Right now this doesn't introduce a semantic change. But we might want to consider changing from __linux__ to __unix__. I don't think we have any linux specific parts in jacktrip right now.


cchafe (commented):
For consistency (and less characters to type) how about this
convention:

_WIN32
_APPLE
_LINUX
_RT_AUDIO
_NO_JACK
_MANUAL_POLL

ntonnaett:
Maybe I should have explained those choices. Identifiers that include double underscores or begin with an underscore followed by a capital letter are reserved for the c++ implementation/compiler: https://en.cppreference.com/w/cpp/language/identifiers

__linux__ (Linux including derivates like Android), _WIN32 (32bit and 64bit) and __APPLE__(macOS and iOS) are predefined macros. They are already set by the compiler. That's the reason I've deleted our own macros for platform identification.

ntonnaett:
One problem that would arise with using `__unix__` is that CYGWIN sets `__unix__` (because POSIX for Windows) and `_WIN32`. FreeBSD just sets `__LINUX__` [right now for the jacktrip port.](https://cgit.freebsd.org/ports/tree/audio/jacktrip/Makefile?id=fa54a01d138292073d7441c36db12dbeaaf37ec5) So they could change this to lowercase.

cchafe:
Good to know. Should we merge as is -- or more to do?

ntonnaett:
I'll add the changelog entry. And then we can merge.

ntonnaett:
Changelog's updated.

cchafe (approved):

PR #456, opened 2022-01-04
Patcher update

psiborg112:
Forgot to take patcher mode 0 (server to clients) into consideration with the upmix update. Also ran Patcher.cpp through clang-format. (So some minor format tidying.)

### DIFF
--- a/src/Patcher.cpp
+++ b/src/Patcher.cpp
@@ -36,6 +36,7 @@
  */
 
 #include "Patcher.h"
+
 #include <QVector>
 
 void Patcher::setPatchMode(JackTrip::hubConnectionModeT patchMode)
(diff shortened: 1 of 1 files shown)


psiborg112:
Also, maybe before merging this we might want to consider adding patcher modes that are equivalent to client fan out/in and full mix but which include the server in the mix. (Seems that mode 0 is otherwise causing some confusion. See here: [https://groups.google.com/g/jacktrip-users/c/HXVC5noYufA/m/O_3jOEqwDAAJ](https://groups.google.com/g/jacktrip-users/c/HXVC5noYufA/m/O_3jOEqwDAAJ)

psiborg112:
To add to my last comment, see #459.

psiborg112:
Superseded by #459

PR #459, opened 2022-01-04
Options to include server in patching

psiborg112:
Add an option to include the server in the patching for client fan out/in and full mix modes. This allows an ensemble member to more easily run a server and their own audio on the same machine.

I've technically implemented these as patch modes 6 and 7, but am using a checkbox in the GUI as a modifier to the existing patch modes. Haven't implemented it on the command line yet and am open to suggestions on whether to add them to the -p list or to use a solution similar to the one in the GUI (ie, an additional switch).


cchafe (approved):
tested GUI and CLI -- works well

PR #463, opened 2022-01-04
Fix additional clazy warning

psiborg112:
Must have been introduced since a previous scan.


ntonnaett:
It might be possible to also generate clazy warnings like in #462.

ntonnaett (approved):

psiborg112:
That could be useful. Just looking at level 2 clazy warnings at the moment, and we have quite a few to say the least. (Have clazy integrated into KDevelop now.) Have fixed the low hanging fruit of old style connects, and thinking of tackling qstring allocations (mostly involves getting rid of implicit conversions from char* to QString).

PR #464, opened 2022-01-07
Fix clazy level 2 warnings

psiborg112:
Fixes hundreds of warnings from the clazy level 2 checks. The only check not done is the copyable-polymorphic check. (A lot of the warnings thrown by this are in the faust generated code, and so are probably safe to ignore.) From the remaining checks there are only 7 warnings left, and they're in parts of the code that I'm less familiar with. (So I didn't want to break anything.)


ntonnaett (commented):
I'm not sure about commenting out all the unused JMess functions. If those JMess functions are commented out your IDE won't suggest them to you when coding. Are we really still using JMess while having the Patcher class now? I'm unfamiliar with this code. Maybe the remaining JMess functions should go to the Patcher class.

psiborg112:
I don't think we really want them suggested by the IDE. A number of them (writeOutput, parseXML, and connectPorts) didn't do anything as it was - the entire body of the function was already commented out. And the disconnectAll function shouldn't be used: it wipes out the entire connection graph, including connections that are not at all related to jacktrip. (It's part of the reason why I rewrote the patching code.)

The only thing we're using the JMess class for now is for the reserved TUB mode. It could probably be moved over to the Patcher class at some stage, but it might be good to keep special, reserved modes separate.

ntonnaett:
I was actually thinking about removing the TUB mode because we don't use it anymore. But I can do this later.

If those other JMess functions are useless for us, let's remove them instead.

psiborg112:
If others are happy for me to remove those commented out functions from the JMess class then I'm down with that.

ntonnaett (approved):

Issue #468, opened 2022-01-08
Putting qmake and meson builds in the same github actions workflow?


dyfer:
A while ago I thought about putting meson and qmake (and, possibly, in the future cmake) builds in the same gha workflow. My thinking is that this would allow easier switching between which build system we use for releases etc.
This could eventually supersede the current meson separate workflow. What do you think about this @ntonnaett?

I made two proposed workflows: [option A](https://github.com/dyfer/jacktrip/actions/runs/1670362806/workflow) and [option B](https://github.com/dyfer/jacktrip/actions/runs/1670365711/workflow). Here's a quick comparison:

**Option A** has a straightforward build configuration in the build matrix, e.g.
```yaml
system-rtaudio: false
bundled-rtaudio: true
nogui: false
weakjack: false
```
but the [build steps](https://github.com/dyfer/jacktrip/actions/runs/1670362806/workflow#L351-L399) have quite a bit of switches to support this and I'm worried this would be a burden to maintain down the road.

**Option B** specifies build configuration directly in the matrix, e.g.
```yaml
system-rtaudio: false
...
meson-config: '-Ddefault_library=static -Drtaudio:jack=disabled -Drtaudio:default_library=static -Dweakjack=true'
```
so on one hand it's more transparent in relation to the general developer documentation, but one needs to be more specific/careful when creating new build configurations. In return [build steps](https://github.com/dyfer/jacktrip/actions/runs/1670365711/workflow#L326-L338) are much shorter and the overall yaml file is shorter too.

Which option seems better?

----
Also, I noticed some fancy clang-tidy commenter in #462 that depends on the meson workflow currently. If my suggestion to unify qmake and meson builds gets eventually merged, I'd move the clang-tidy "insecure" part to a new workflow.

ntonnaett:
I'd prefer option A. I think this would make it much more clear in the end what we actually test. We would probably need to talk about the switches we use. For example, if I set bundled-rtaudio to true with meson I might want to also set rtaudio-static to true.

Yesterday I was thinking about removing `on push` from the meson workflow. `on pull_request` already triggers when pushing to an open PR branch. But would that break uploading the release artifacts?

dyfer:
Thanks @ntonnaett, option A sounds good.

[...]

PR #469, opened 2022-01-08
[GHA] Incorporate meson build in the main GHA file

dyfer:
As discussed in #468, add meson builds to `jacktrip.yaml` in order to have CI for all build systems in one workflow.

TODO:

- [x] include build system in all artifact names
- [x] use separate (cleaner) artifact names for releases
- [x] move static analysis out of meson build
- [x] remove meson workflow
- [x] add compiler names to the job names
- [x] add rtaudio to linux meson build


ntonnaett:
The names should have the same structure for all and ordered by the most common things (os-arch-buildsystem-options).
The meson-build workflow needs to be removed and the clang-tidy workflow connected to the jacktrip workflow. The analysis steps are missing too.

dyfer:
> The names should have the same structure for all and ordered by the most common things (os-arch-buildsystem-options). 

The reason I left the default build names as they were was because these become artifact names for the release. Do you think it's desirable to have the release artifact like `macOS-x64-qmake-static`?

We can also rename artifacts before uploading to the release, I guess.

> The meson-build workflow needs to be removed and the clang-tidy workflow connected to the jacktrip workflow. The analysis steps are missing too.

I was planning to do this in a separate PR. Also I'd prefer to move the analysis part to a separate workflow, since it's not strictly related to building/testing, but if you prefer I can put it in the build workflow.

ntonnaett:
clang-tidy needs a build that generates some files. We could add Windows and macOS support too when integrating it in jacktrip.yml. Right now we are only analyzing a subset, because we use preprocessor macros.

clang-tidy has nothing to do with clang-format except that it belongs to the same toolset.

We should rename the artifacts in the end. The source packages have also bad names right now.

dyfer:
> clang-tidy needs a build that generates some files. We could add Windows and macOS support too when integrating it in jacktrip.yml. Right now we are only analyzing a subset, because we use preprocessor macros.

Sorry, I'm not super familiar with this. Where are these files generated? During build process? In meson only? can they be generated without a full build? (e.g. clang-tidy only "target" or something like that)?

ntonnaett:
In our case it needs the whole build, because moc genereates header files. It's meson only. If we put this in another workflow we end up with another build workflow. Because we want to have this for all parts of our code. And if we turn them off via our defines, they don't get checked.

In the meson-build workflow [the analysis is done](https://github.com/jacktrip/jacktrip/blob/7217310372bcf4210a2109a0def4dde5b5c80efc/.github/workflows/meson.yml#L108), the fixes get saved in a file and uploaded as an artifact for the use of the clang-tidy workflow later.

ntonnaett:
Right now it just comments about the changes that are happening in the PR. We might want to change that to have a clang-tidy clean dev and main all the time.

dyfer:
OK, thanks for the explanation. I'll add it to `jacktrip.yaml`

dyfer:
Instead of renaming artifacts at the upload stage, which seems to be cumbersome to implement without hardcoding names in multiple places, I though it would be easier if tagged builds use alternate names for artifacts. I hope that's OK. You can see a test release with the current approach here: https://github.com/dyfer/jacktrip/releases/tag/v1.5.0-testNewArtifactUpload-v1

dyfer:
I've added clang-tidy also, but couldn't [test it on my fork](https://github.com/dyfer/jacktrip/pull/1) since I think I'd need to get the change in `clang-tidy.yml` (so it runs after the `build` workflow) into main first...?

dyfer:
All right, I think it's okay now: https://github.com/dyfer/jacktrip/pull/2 (though commends were not triggered since I didn't make proper changes on the main branch?)

BUT! I had to add `fetch-depth: 0` to `actions/checkout@v2`, otherwise git diff was [having issues](https://github.com/dyfer/jacktrip/runs/4751485737?check_suite_focus=true#step:24:16). Why is that? What does checkout depth have to do with `git diff`?

ntonnaett:
That was new to me too. Github doesn't check out the whole history of your branch. I think they call it shallow clone. But if you don't have a history you can't diff against your base branch. It's wiped from history.

ntonnaett:
I was irritated that it set Qt environment variables although it didn't build Qt. But that step also sets other variables like the new CLANG_TIDY_*. That step [should be renamed](https://github.com/dyfer/jacktrip/blob/a441b1cde9bca6c62dfd9ae27b3d450d6c5b5b12/.github/workflows/jacktrip.yml#L354).

ntonnaett:
It seems that at the analyze build step it doesn't resolve the environment variable and fails (silently).

ntonnaett:
Could we add the compiler name to the check name? Right now meson builds with msvc on Windows. Would be great if you see this just by looking at the name.

ntonnaett:
Can we enable RtAudio for the Linux Meson build, so that part gets check too by the static analyzer?

dyfer:
> I was irritated that it set Qt environment variables although it didn't build Qt. But that step also sets other variables like the new CLANG_TIDY_*. That step [should be renamed](https://github.com/dyfer/jacktrip/blob/a441b1cde9bca6c62dfd9ae27b3d450d6c5b5b12/.github/workflows/jacktrip.yml#L354).

I don't understand. This step sets only Qt-related variables, and not only when building Qt (i.e. adds homebrew Qt's `bin` path to the `$PATH`). 

Are you saying that `CLANG_TIDY_*` vars should be set there? Why?

ntonnaett:
Oh no! Now you are confused, because I was confused. The output of this step misguided me. I just looked at the wrong places. Sorry!

The reason I've looked there in the first place was, because it shouldn't be necessary to set those variables, if it's not our own Qt5 build. It wasn't necessary in the meson workflow, if I'm not confused again… Is it even necessary for the static Qt build? On Linux (and other Unixes) libraries usually get installed in the right places. Maybe that whole step is just necessary for static Qt builds on Windows.

ntonnaett:
We should have a variable for the compiler name. It's important for Windows. A meson build (msvc) with shared vcpkg rtaudio (mingw) wouldn't link correctly.

I know you've invested a lot of time in the static qt build process. But what do you think about switching to MSYS2 on Windows and asking for a static build on homebrew?

dyfer:
> Oh no! Now you are confused, because I was confused. The output of this step misguided me. I just looked at the wrong places. Sorry!

No worries! Confusion compounded :)

> The reason I've looked there in the first place was, because it shouldn't be necessary to set those variables, if it's not our own Qt5 build. It wasn't necessary in the meson workflow, if I'm not confused again… Is it even necessary for the static Qt build? On Linux (and other Unixes) libraries usually get installed in the right places. Maybe that whole step is just necessary for static Qt builds on Windows.

Hm, I'd need to trace this back. I _believe_ something was not right with the static build if this was not set, not sure on which platform. And on macOS this might have been needed when making a package to run `macdeployqt` maybe? Now I see the `assemble_app` script should find qt on its own so I'm not sure anymore (and in the end we haven't run macdeployqt in CI...). If you feel this is worth investigating let's open an issue after this PR is done.

> We should have a variable for the compiler name. It's important for Windows. A meson build (msvc) with shared vcpkg rtaudio (mingw) wouldn't link correctly.

In principle I don't disagree to have a compiler name specified explicitly, but for just the vcpkg issue I think we already have an appropriate [matrix entry](https://github.com/jacktrip/jacktrip/pull/469/files#diff-400b21146a0aa3667e1c8f1d95c364ac50306b03359d195bff0e90725ecaabf8R172) (which I'll try shortly).

> I know you've invested a lot of time in the static qt build process. But what do you think about switching to MSYS2 on Windows and asking for a static build on homebrew? Just an idea.

Sure, I would be a bit sad do see my work into static Qt go, but if there's a better solution then let's consider it. However:
- what does MSYS2 give us exactly?
- I still think that qmake+mingw is the _easiest_ build env to set up on Windows (*) so I believe we should keep supporting that...
- I don't think we can get static qt from homebrew, please correct me if I'm wrong
- having said all this - we _could_ consider making an app package, at least on macOS and Windows (using `macdeployqt`) which uses official (shared) qt libs. This seems to be less straightforward on Linux (i.e. creating a "portable" app) but I understand that Fedora-adjacent distros as addressed with Flatpak. For Debian/Ubuntu-adjacent it would be nice to create a PPA or something before abandoning a static build

(*) now with rtaudio added back to qmake, setting up qmake build system in windows is quite simple: official qt installer to get qt + mingw and git installer to get git with bash. Then adjust env variables as needed (`$PATH`) and run the build script in bash. I'm not saying this is the best overall dev environment to work in, but it's the smallest one to set up. No need for python, chocolatey, vcpkg, python packages.

dyfer:
> Could we add the compiler name to the check name? Right now meson builds with msvc on Windows. Would be great if you see this just by looking at the name.

I assume this would be `clang` for macOS and `msvc` or `mingw` for Windows, right? And `gcc` or `g++` for Linux?

ntonnaett:
Let's have `clang`, `gcc` and `msvc`. `clang` on macOS, `msvc` and `gcc` on Windows and `gcc` on Linux. mingw could also be used with clang. So `gcc` is more precise. I think other combinations are not relevant.

dyfer:
> Can we enable RtAudio for the Linux Meson build, so that part gets check too by the static analyzer?

Would you like bundled or system lib? It seems that system lib is slightly older version (5.0.0) - not sure if that matters for static analysis.

dyfer:
> We should have a variable for the compiler name. It's important for Windows. A meson build (msvc) with shared vcpkg rtaudio (mingw) wouldn't link correctly.

Oops, the matrix entry was there, but the implementation was not. Correcting now.

ntonnaett:
> > The reason I've looked there in the first place was, because it shouldn't be necessary to set those variables, if it's not our own Qt5 build. It wasn't necessary in the meson workflow, if I'm not confused again… Is it even necessary for the static Qt build? On Linux (and other Unixes) libraries usually get installed in the right places. Maybe that whole step is just necessary for static Qt builds on Windows.
> 
> Hm, I'd need to trace this back. I _believe_ something was not right with the static build if this was not set, not sure on which platform. And on macOS this might have been needed when making a package to run `macdeployqt` maybe? Now I see the `assemble_app` script should find qt on its own so I'm not sure anymore (and in the end we haven't run macdeployqt in CI...). If you feel this is worth investigating let's open an issue after this PR is done.
> 

Let's do this afterwards. I think it would be best if we have our CI workflows as similar as possible to our documentation.

> > We should have a variable for the compiler name. It's important for Windows. A meson build (msvc) with shared vcpkg rtaudio (mingw) wouldn't link correctly.
> 
> In principle I don't disagree to have a compiler name specified explicitly, but for just the vcpkg issue I think we already have an appropriate [matrix entry](https://github.com/jacktrip/jacktrip/pull/469/files#diff-400b21146a0aa3667e1c8f1d95c364ac50306b03359d195bff0e90725ecaabf8R172) (which I'll try shortly).
> 

Yeah, I saw them. I think there were two that are similar for vcpkg and something else...

> > I know you've invested a lot of time in the static qt build process. But what do you think about switching to MSYS2 on Windows and asking for a static build on homebrew? Just an idea.
> 
> Sure, I would be a bit sad do see my work into static Qt go, but if there's a better solution then let's consider it. However:
> 
>     * what does MSYS2 give us exactly?
> 
>     * I still think that qmake+mingw is the _easiest_ build env to set up on Windows (*) so I believe we should keep supporting that...

MSYS2 it is ;) You install MSYS2, install your dependencies with pacman, start a MSYS2/mingw shell and build jacktrip.

> 
>     * I don't think we can get static qt from homebrew, please correct me if I'm wrong
> 

Static qt is missing from homebrew right now. But I found an issue where they said that they add static builds to libraries when somebody asks.

>     * having said all this - we _could_ consider making an app package, at least on macOS and Windows (using `macdeployqt`) which uses official (shared) qt libs. This seems to be less straightforward on Linux (i.e. creating a "portable" app) but I understand that Fedora-adjacent distros as addressed with Flatpak. For Debian/Ubuntu-adjacent it would be nice to create a PPA or something before abandoning a static build
> 

A PPA would be a good idea. The official repository already has version 1.5.0 for the next Ubuntu version. We might just ask the package maintainer if they could create a PPA for the latest jacktrip version. Fedora's RPM package is currently at 1.4.3. Arch is at 1.5.0. So I I think Ubuntu is the only one that's missing. Fedora and Arch already build their packages with Meson.

> 
> (*) now with rtaudio added back to qmake, setting up qmake build system in windows is quite simple: official qt installer to get qt + mingw and git installer to get git with bash. Then adjust env variables as needed (`$PATH`) and run the build script in bash. I'm not saying this is the best overall dev environment to work in, but it's the smallest one to set up. No need for python, chocolatey, vcpkg, python packages.

I'll write a MSYS2 build tutorial, if I find some time. Could get difficult.

dyfer:
>  Is it even necessary for the static Qt build?

I think I remember now. This variable is necessary when building static Qt _if the build is read from cache_ at the very least.

dyfer:
> MSYS2 it is ;) You install MSYS2, install your dependencies with pacman, start a MSYS2/mingw shell and build jacktrip.

> I'll write a MSYS2 build tutorial, if I find some time. Could get difficult.

All right, I'll revisit my opinion after trying MSYS2 when you write a tutorial :)

ntonnaett:
> > Can we enable RtAudio for the Linux Meson build, so that part gets check too by the static analyzer?
> 
> Would you like bundled or system lib? It seems that system lib is slightly older version (5.0.0) - not sure if that matters for static analysis.

Bundled, because the system version is too old. I already filed a bug about that. It shouldn't matter that much. I mean the JackTrip version in Ubuntu is built with that version…

ntonnaett:
> > Is it even necessary for the static Qt build?
> 
> I think I remember now. This variable is necessary when building static Qt _if the build is read from cache_ at the very least.

That makes sense. Could you add a comment about that?

dyfer:
> Bundled, because the system version is too old. 

Because I tried system library first I found an issue with the rtaudio package name in the script. Thanks :)

dyfer:
> > > Is it even necessary for the static Qt build?
> > 
> > 
> > I think I remember now. This variable is necessary when building static Qt _if the build is read from cache_ at the very least.
> 
> That makes sense. Could you add a comment about that?

Actually... I think I wasn't exactly right. All these variables were needed to find `qmake` in various qt scenarios. So I added a commend and a switch so that it runs on qmake builds only

dyfer:
I think I addressed all your comments @ntonnaett. Anything that I've missed?

ntonnaett:
One last thing. Could we turn on the jack backend on windows for meson?

dyfer:
> One last thing. Could we turn on the jack backend on windows for meson?

Hm. It seems that meson is [not finding jack](https://github.com/jacktrip/jacktrip/runs/4766433354?check_suite_focus=true#step:17:177) which seems to also have been the case in your [original meson build](https://github.com/jacktrip/jacktrip/runs/4758434969?check_suite_focus=true#step:12:154). Maybe it's an issue with meson/windows/pkgconfig? Or we need to add some search paths somewhere? 

FWIW I don't think that regular Jack install on Windows has the pkgconfig entry, in qmake we [search for it in predetermined paths](https://github.com/jacktrip/jacktrip/blob/dev/jacktrip.pro#L147) (as inelegant as it is)...

ntonnaett:
Yes, I forgot about that. Let's use MSYS2^^

ntonnaett (approved):

ntonnaett:
I was already thinking about adding jack-headers as a git submodule for weakjack.

dyfer:
Thanks for the review!

ntonnaett:
Thanks for this huge effort!

PR #473, opened 2022-01-10
Use bundled rtaudio for Windows release builds

dyfer:
This is mostly a housekeeping PR aiming to optimize our GHA builds.

The main change I'm suggesting is to use bundled rtaudio in static Windows release builds (currently we use the build from vcpkg, which does get built every time anyway). I'll test this build before marking the PR ready for review to make sure there are no regressions.

Reasoning: 
Using bundled rtaudio is much easier on Windows than installing the library using vcpkg (at least until a better solution is implemented, like moving to MSYS2). Thus I'd like to **use bundled rtaudio as the default approach for building the release artifacts** on Windows. Using rtaudio from vcpkg is still supported and tested for (in the shared library Windows build). 

This PR also removes one build from our already considerable build matrix, while - I believe - still testing for most practical built configurations.

Other changes in this PR include changing macOS and Linux `bundled_rtaudio` builds to use shared qt library to avoid unnecessary builds of Qt when cache is not available, e.g. on releases. 

Once rtaudio is more tested on Linux and we decide it's a viable option for that platform, we could add bundled rtaudio to the Linux static build (if we keep providing it), but that's to be considered in the future.

### DIFF
--- a/.github/workflows/jacktrip.yml
+++ b/.github/workflows/jacktrip.yml
@@ -132,22 +132,8 @@ jobs:
             build-system: meson
             meson-library-type: both
             
-          - name: Windows-x64-qmake-gcc-static
-            release-name: Windows-x64
-            runs-on: windows-2019
-            system-rtaudio: true
-            bundled-rtaudio: false
-            nogui: false
-            weakjack: true
-            vcpkg-triplet: x64-mingw-static
-            static-qt-version: 5.15.2
-            qt-static-cache-key: 'v04'
-            jacktrip-path: release/jacktrip.exe
-            binary-path: binary
-            installer-path: installer
-            build-system: qmake
-            
           - name: Windows-x64-qmake-gcc-static-bundled_rtaudio
+            release-name: Windows-x64
             runs-on: windows-2019
             system-rtaudio: false
             bundled-rtaudio: true
@@ -155,8 +141,9 @@ jobs:
             weakjack: true
             static-qt-version: 5.15.2
             qt-static-cache-key: 'v04'
-            # jacktrip-path: release/jacktrip.exe
-            # binary-path: binary
+            jacktrip-path: release/jacktrip.exe
+            binary-path: binary
+            installer-path: installer
             build-system: qmake
 
           - name: Windows-x64-qmake-gcc-shared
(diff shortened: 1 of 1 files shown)


dyfer:
I have done basic test with the Windows build, using both ASIO2ALL and a dedicated ASIO driver, at 128 and 32 FPP, 48k. Seems to work fine. I'd like someone else to test the Windows build before we merge this, just to be sure. Thanks!

ntonnaett:
Maybe we could get something like a test group together. The user community is quite separated from the developer community. What do you think, @forsythac and @mikedickey?

forsythac:
Will do @ntonnaett - just caught up with Marcin about testing this. We'll give it a spin on Windows and write back with the results

mikedickey:
I tested out 3aa1175 on my Windows 11 machine, and it seems to work great.

mikedickey (approved):

forsythac:
already merged, but just wanted to close the loop and say that I also tested this on Windows 10 without issue. Focusrite ASIO drivers, tested at 16, 32, 64, 128, and 1024. Thanks all

mattahorton:
Sorry to come in very late to this, but @dyfer I've got a question.

> Other changes in this PR include changing macOS and Linux bundled_rtaudio builds to use shared qt library to avoid unnecessary builds of Qt when cache is not available, e.g. on releases.

Does this mean now that we will not have a build for users that includes both Qt and RtAudio? It is very critical that there be at least one build that can be used without installing any other dependencies. 

I could be _completely_ misunderstanding, so just tell me if that's the case.

dyfer:
> Does this mean now that we will not have a build for users that includes both Qt and RtAudio? It is very critical that there be at least one build that can be used without installing any other dependencies.

We still provide builds that include RtAudio (on macOS and Windows) that do not require any dependencies and there's no plan to stop offering these I believe.

What this change means is that in our automated build system the builds that use "bundled rtaudio" (that means rtaudio library compiled from source with the project) now use shared qt library. _These builds are not aimed at releasing_, but just to test whether building with bundled rtaudio works. Rtaudio library is still included in the release builds on macOS and Windows, but on macOS the library itself comes from homebrew (i.e. it's not built from the "bundled" source code).

mattahorton:
Gotcha. Thanks for the explanation.

PR #478, opened 2022-01-26
Fix GUI command line warning

psiborg112:
The dialog that warns that most command line options are ignored with the --gui switch was not being shown in some cases when verbose mode was already set in the GUI. Also, remove manual formatting of dialog box contents.


ntonnaett (approved):
Looks good. But didn't test it.

Issue #477, opened 2022-01-25
Remove RTAudio warning (or make friendlier)


forsythac:
With the inclusion of RTaudio, more and more people are using the JackTrip Core app without having ever installed JACK

Right now, we give a very informative but intimidating error message to new users:

![image](https://user-images.githubusercontent.com/3578658/150874429-3a98f660-9094-4f15-8503-be850454fa0a.png)

We've had a decent number of people think that their installation was not successful due to this message. What's the value of this warning? Can we remove it, since everything is working as expected?

ntonnaett:
They would probably wonder why some features don't work. But I must say that message looks bad.

We already have places where the gui just changes according to what backends are available. Maybe it's sufficient if we remove the hub server mode for RtAudio? Maybe this is already the case? Maybe we can add a help button somewhere to explain what they are missing out without JACK.

psiborg112:
I'm happy for someone to make it friendlier if they want, but I don't think it should be removed. I understand that a lot of people don't need JACK, but there are plenty of use cases where it's still important, and I've lost so much time to troubleshooting issues with my students because they just totally ignored that warning message.

And yes, we currently remove the hub server mode choice if only RtAudio is available.

psiborg112:
If you do want to remove it, can I request that whoever does it simply disables it from running by using a preprocessor directive so that I can keep it in my builds.

```
#ifdef NO_JTVS
(code that shows the warning)
#endif
```

SynthiaPayne:
Today I was helping someone who a) never got passed that RtAudio 
warning, b) had no idea what RtAudio is, c) thought the error message 
meant that his computer would not run the app, and d) thought that the 
error message meant that the app was not installed. All totally 
understandable for an inexperienced user.

I wonder if some explanation of the difference between JACK and RtAudio 
might help. Like, "JackTrip needs a backend app to hand off the audio it 
will send and receive from the internet. This backend app can be RtAudio 
or JACK, however, the JackTrip server mode is only available with JACK."

or something like that...

synth

psiborg112:
If the main concern is that people don't think the app will run then we should just address that directly. Might I suggest:

An installation of JACK was not found. JackTrip will still run using a different audio library (RtAudio) but some more advanced features, like the ability to run your own hub server, will not be available.

(If you install JACK at a later stage, you can re-enable these features by changing the settings in the Audio Backend tab.)

ntonnaett:
[...]

PR #479, opened 2022-01-27
RtAudio warning changes

psiborg112:
Remove the RtAudio warning for official builds and tweak it for NO_JTVS builds (the binaries maintained by @psiborg112).

Additionally, save if RtAudio has been used as a fallback option rather than selected as a deliberate choice by the user. That way, if JACK is installed later, it can be reinstated as the default backend. (NO_JTVS builds)

### DIFF
--- a/src/gui/qjacktrip.cpp
+++ b/src/gui/qjacktrip.cpp
@@ -233,6 +233,7 @@ QJackTrip::QJackTrip(int argc, QWidget* parent)
     m_ui->basePortSpinBox->setVisible(false);
     m_ui->autoPatchGroupBox->setVisible(false);
     m_ui->requireAuthGroupBox->setVisible(false);
+    m_ui->backendWarningLabel->setVisible(false);
 
 #ifdef RT_AUDIO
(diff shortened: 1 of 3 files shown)


psiborg112:
Initial fix for #477 but might require additional changes depending on further discussion there.

psiborg112:
Without JACK support as in with NO_JACK set? The GUI was never designed to work in that configuration - it relies on using weakjack at a bare minimum. (Setting NO_JACK should maybe set NO_GUI in our build configs.) So yeah, the interface will be entirely inaccurate in that case.

ntonnaett:
Maybe we could get rid of NO_JACK and just toggle between dynamically and weakly linking jack. I already thought about adding the jack headers as a submodule. That would make it quite easy. It might even clean up the source code a bit.

ntonnaett:
But the main point for this PR in my message was actually that the warning in the backend tab is too subtle IMHO. Maybe we could emphasize it a little.

ntonnaett (approved):
I think the message could be designed better, but this shouldn't block this PR.

PR #481, opened 2022-01-28
Don't connect our UDP socket

### LINKED ISSUE #474: UDP timeout not working
When __MANUAL_POLL__ is not defined as default case, "Disconnect after 10 seconds of no network activity" option is not working as intended. I tested on version 1.5.1.

psiborg112:
One of the quirks of the pre-IPv6 networking code was that we called connect on our UDP socket. The benefit was that we could just call send rather than sendto and it would only accept packets arriving from the specified destination.

I abandoned it for IPv6 because adapters commonly have multiple addresses and there was no guarantee that packets would be arriving back from the same address. I kept it for IPv4 because if it ain't broke... But it is broken. It seems to cause issues with both epoll and kqueue causing them to report that the socket is ready for reading when there's no data on it. Fixes #474.

(Also, clean out some of the very old and very broken 'two sockets on a single port' code.)

### DIFF
--- a/src/UdpDataProtocol.cpp
+++ b/src/UdpDataProtocol.cpp
@@ -337,7 +337,7 @@ functions. DWORD n_bytes; WSABUF buffer; int error; buffer.len = n; buffer.buf =
         n_bytes = ::sendto(mSocket, buf, n, 0, (struct sockaddr*)&mPeerAddr6,
                            sizeof(mPeerAddr6));
     } else {
-        n_bytes = 
+        n_bytes =
             ::sendto(mSocket, buf, n, 0, (struct sockaddr*)&mPeerAddr, sizeof(mPeerAddr));
     }
     return n_bytes;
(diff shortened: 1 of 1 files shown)


src/UdpDataProtocol.cpp
```diff
@@ -388,7 +337,8 @@ functions. DWORD n_bytes; WSABUF buffer; int error; buffer.len = n; buffer.buf =
         n_bytes = ::sendto(mSocket, buf, n, 0, (struct sockaddr*)&mPeerAddr6,
                            sizeof(mPeerAddr6));
     } else {
-        n_bytes = ::send(mSocket, buf, n, 0);
+        n_bytes =
+            ::sendto(mSocket, buf, n, 0, (struct sockaddr*)&mPeerAddr, sizeof(mPeerAddr));
```
ntonnaett, line 341: Both times we cast to struct sockaddr*. Could it be that we don't need to distinguish between IPv4 and IPv6 at this point?

ntonnaett: Nevermind.

ntonnaett (approved):

PR #482, opened 2022-01-29
Update MacOS packaging script

psiborg112:
Sign the generated package if it isn't already and a certificate is provided.


dyfer (approved):
Thanks @psiborg112 
It looks good "on the screen" and I asses it doesn't break anything since the package is still properly created in GitHub Actions. However, I am not able to test notarization itself as I do not have the required dev account.

PR #486, opened 2022-01-31
Initial attempt to get JackTrip on FreeBSD

psiborg112:
Change preprocessor definitions in an initial attempt to get JackTrip running on FreeBSD

### DIFF
--- a/src/main.cpp
+++ b/src/main.cpp
@@ -262,8 +262,8 @@ int main(int argc, char* argv[])
 #ifndef _WIN32
                 setupUnixSignalHandler(UdpHubListener::sigIntHandler);
 #else
-                isHubServer = true;
-                SetConsoleCtrlHandler(windowsCtrlHandler, true);
+            isHubServer = true;
+            SetConsoleCtrlHandler(windowsCtrlHandler, true);
 #endif
                 udpHub->start();
             } else {
--- a/src/main.cpp
+++ b/src/main.cpp
@@ -279,13 +279,13 @@ int main(int argc, char* argv[])
 #ifndef _WIN32
                 setupUnixSignalHandler(JackTrip::sigIntHandler);
 #else
-                std::cout << SetConsoleCtrlHandler(windowsCtrlHandler, true) << std::endl;
+            std::cout << SetConsoleCtrlHandler(windowsCtrlHandler, true) << std::endl;
 #endif
 #ifdef WAIRTOHUB  // WAIR
                 jackTrip->startProcess(
                     0);  // for WAIR compatibility, ID in jack client name
 #else
-                jackTrip->startProcess();
+            jackTrip->startProcess();
 #endif  // endwhere
             }
 
--- a/src/main.cpp
+++ b/src/main.cpp
@@ -97,7 +97,7 @@ QCoreApplication* createApplication(int& argc, char* argv[])
             return new QCoreApplication(argc, argv);
         }
 #else
-#ifdef __linux__
+#if defined(__linux__) || defined(__FreeBSD__)
         // Check if X or Wayland environment variables are set.
         if (std::getenv("WAYLAND_DISPLAY") == nullptr
             && std::getenv("DISPLAY") == nullptr) {
@@ -108,7 +108,7 @@ QCoreApplication* createApplication(int& argc, char* argv[])
                       << std::endl;
             std::exit(1);
         }
-#endif  // __linux__
+#endif
         return new QApplication(argc, argv);
 #endif  // NO_GUI
     } else {
@@ -123,7 +123,7 @@ void qtMessageHandler([[maybe_unused]] QtMsgType type,
     std::cerr << msg.toStdString() << std::endl;
 }
 
-#if defined(__linux__) || defined(__APPLE__)
+#ifndef _WIN32
 static int setupUnixSignalHandler(void (*handler)(int))
 {
     // Setup our SIGINT handler.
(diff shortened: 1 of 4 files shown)


psiborg112:
@ntonnaett, clang-format is proposing some changes here in main that don't seem to make sense. It's pulling the #else part of a couple of preprocessor directives back an indent, and I'm not sure why. (This 'correction' was in the code previously, but I thought it was an oversight and manually adjusted things in this upload.)

yurivict:
Somehow the patches don't apply to 1.5.1.

psiborg112:
They won't. You'll need to pull this branch from the repository. (If you have git installed then you can just use the following command line: git clone --branch freebsd https://github.com/jacktrip/jacktrip.git)

yurivict:
The ```freebsd``` branch builds usimg cmake and then the link fails:
```
ld: error: undefined symbol: RtAudio::~RtAudio()
>>> referenced by RtAudioInterface.cpp
>>>               CMakeFiles/jacktrip.dir/src/RtAudioInterface.cpp.o:(RtAudioInterface::~RtAudioInterface())
>>> referenced by RtAudioInterface.cpp
>>>               CMakeFiles/jacktrip.dir/src/RtAudioInterface.cpp.o:(RtAudioInterface::~RtAudioInterface())
>>> referenced by RtAudioInterface.cpp
>>>               CMakeFiles/jacktrip.dir/src/RtAudioInterface.cpp.o:(RtAudioInterface::getDefaultDevice(bool))
>>> referenced 11 more times

ld: error: undefined symbol: RtAudio::RtAudio(RtAudio::Api)
>>> referenced by RtAudioInterface.cpp
>>>               CMakeFiles/jacktrip.dir/src/RtAudioInterface.cpp.o:(RtAudioInterface::setup())
>>> referenced by RtAudioInterface.cpp
>>>               CMakeFiles/jacktrip.dir/src/RtAudioInterface.cpp.o:(RtAudioInterface::getDefaultDevice(bool))
>>> referenced by RtAudioInterface.cpp
>>>               CMakeFiles/jacktrip.dir/src/RtAudioInterface.cpp.o:(RtAudioInterface::printDeviceInfo(unsigned int))
>>> referenced 4 more times

ld: error: undefined symbol: RtAudio::openStream(RtAudio::StreamParameters*, RtAudio::StreamParameters*, unsigned long, unsigned int, unsigned int*, int (*)(void*, void*, unsigned int, double, unsigned int, void*), void*, RtAudio::StreamOptions*, void (*)(RtAudioError::Type, std::__1::basic_string<char, std::__1::char_traits<char>, std::__1::allocator<char> > const&))
>>> referenced by RtAudioInterface.cpp
>>>               CMakeFiles/jacktrip.dir/src/RtAudioInterface.cpp.o:(RtAudioInterface::setup())
c++: error: linker command failed with exit code 1 (use -v to see invocation)
```

rtaudio-5.1.0 is installed, which provides ```rtaudio.pc```.

psiborg112:
Use the ./build script for a qmake build. Or meson. (I'm pretty sure no one other than me uses cmake to build the project so it's only tested on my development set up.)

The fact that it builds though is very encouraging.

yurivict:
It builds fine with meson.

When you would merge this branch and make a release I would be happy to pick it up into the port.

src/main.cpp
```diff
@@ -97,7 +97,7 @@ QCoreApplication* createApplication(int& argc, char* argv[])
             return new QCoreApplication(argc, argv);
         }
 #else
-#ifdef __linux__
+#if !(defined(_WIN32) || defined(__APPLE__))
```
ntonnaett, line 100:
```suggestion
#if defined(__unix__)
```
That's more clear, I think. On https://erato1004.tistory.com/entry/Predefined-Compiler-Macros it states that macOS sets `BSD` but not `__unix__`.

ntonnaett:
The rest looks good to me. A small side quest(ion): Why don't we use SO_REUSEPORT on linux? It's supported since kernel version 3.9.

src/main.cpp
psiborg112: Sure. I thought MacOS set __unix__, but seems it doesn't.

psiborg112: Although on the plus side the current option means we are only ever relying on the state of _WIN32, \_\_APPLE\_\_, and \_\_linux\_\_

ntonnaett: That's true. I wanted to change that because I initially misread it.

ntonnaett: Maybe we could change `!` to `not`? Shouldn't that work?

psiborg112: Yeah, totally understand that. (Reordered some later ones for that exact same reason.) If you'd prefer \_\_unix\_\_ then I'm happy to go with that

psiborg112: Yeah, totally understand that. (Reordered some later ones for that exact same reason.) If you'd prefer \_\_unix\_\_ then I'm happy to go with that.

ntonnaett: My main concern is the `!`. So I would be okay with `not` but I would prefer the `__unix__`.

psiborg112: Let's go with \_\_unix\_\_ then.

psiborg112:
Because I guess we don't really want SO_REUSEPORT. We don't need multiple sockets to be bound to the port, we just want to be able to bind to it when it's in TIME_WAIT. I guess the real question is whether we still need SO_REUSEPORT on MacOS?

ntonnaett (approved):

ntonnaett:
We should open a discussion/issue about the SO_REUSEPORT.

PR #488, opened 2022-02-01
macOS arm64

dyfer:
I've tried to build JackTrip on macOS 11.6 under Xcode 13.2 on m1 hardware (arm64)

TL;DR:
JackTrip builds and runs fine on arm64 mac (aka Apple Silicon / M1).

The good:
- with this change JackTrip builds
- works fine with Jack and RtAudio

~~The bad~~
- ~~doesn't work with RtAudio~~

**There is no issue with RtAudio**, something was wrong with my system. RtAudio backend works fine after restarting.

~~I'm not sure if we want to merge this PR in the current form. While it provides the minimum to detect the architecture and build in qmake, we might also consider being able to set architecture (e.g. to cross-compile or compile universal binary)~~ EDIT: I think it's fine as it is. I think cross-compiling is a topic to be considered at a later stage.

Also: 
- meson build worked fine on arm64 with no changes
- I haven't tested CMake

All in all, I'm somewhat hopeful about being able to provide a native arm64 build at some point. Whether we can make it a universal binary before moving to Qt 6, that I'm not sure. But we could offer two binaries on macOS for the time being.

### DIFF
--- a/build
+++ b/build
@@ -94,6 +94,10 @@ MCMD=make
 elif [[ $platform == 'macosx' ]]; then
   QCMD=qmake
   QSPEC=macx-clang
+  arch=`arch`
+  if [[ "$arch" == 'arm64' ]]; then
+    QSPEC=macx-clang-arm64
+  fi  
   # if qmake is not in path, try homebrew qt5
   if ! command -v $QCMD &> /dev/null; then
     # echo "qmake not found in \$PATH, searching for homebrew qt5"


build
```diff
@@ -94,6 +94,10 @@ MCMD=make
 elif [[ $platform == 'macosx' ]]; then
   QCMD=qmake
   QSPEC=macx-clang
+  arch=`arch`
```
ntonnaett, line 97: How does this work, if you define arch as `arch` and then check if it's `arm64`? Why isn't the architecture set automatically?

dyfer: There's an `arch` command (`/usr/bin/arch`) which returns architecture, similar to how we get [platform name](https://github.com/jacktrip/jacktrip/blob/main/build#L71).

> Why isn't the architecture set automatically?

I was building off of how the script was already set up with explicitly set architecture. It seems we can omit it, at least on macos. I'll update the PR.

ntonnaett: What the… Why this syntax?

ntonnaett (approved):

dyfer:
Sorry, I'm not sure if this addition was needed after all. It doesn't seem to break anything, but trying it again I seem to be able to build on arm64 even without this addition (I get an error `Could not find qmake spec 'macx-clang'` but the build continues and creates an arm64 binary). I'm not sure why the build failed in the first place, it could've been something local in my system. Sorry again for the noise.

I'll likely revisit this again if/when we look into releasing arm64 or "universal" binaries. For now, it builds fine on both architectures on macOS.

PR #492, opened 2022-02-18
Fix false error message (M1 builds)

psiborg112:
The current JACK universal build (1.9.20) doesn't cleanly close the client on arm64 and triggers the jack_on_shutdown callback when it forcibly removes it. This workaround prevents the GUI from incorrectly reporting that the JACK server has shutdown.

### DIFF
--- a/src/JackTrip.cpp
+++ b/src/JackTrip.cpp
@@ -985,6 +985,8 @@ void JackTrip::tcpTimerTick()
 //*******************************************************************************
 void JackTrip::stop(const QString& errorMessage)
 {
+    // Take a snapshot of sJackStopped
+    bool serverStopped = sJackStopped;
     mStopped = true;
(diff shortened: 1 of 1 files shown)


ntonnaett:
I don't understand what you are saying. When closing jacktrip, JACK is shut down? If this is an upstream bug, it would be good to have a link to the upstream issue.

ntonnaett (approved):

psiborg112:
JACK keeps running, but it feels the need to destroy the client, triggering the callback and causing us to wrongly report that JACK was shutdown. It sometimes results in segfaults as well. Sorry, I thought I'd included a link to the upstream bug, but hadn't. It's here: [https://github.com/jackaudio/jack2/issues/841](https://github.com/jackaudio/jack2/issues/841)

PR #493, opened 2022-02-18
plc

cchafe:
Tested with fedora 33, 34, 35, pipewire, and >>wifi<< which works well
(I've only tested with qmake build)

NOTE: the queue buffer length is in terms of ms -- not number of buffers

to test, just switch on in the GUI Jitter Buffer : Use buffer strategy : 3 (experimental)

Ring buffering has been replaced by "Regulator" with pushPacket called from network thread and pullPacket from audio thread

Turn on Display IO Stats to see PUSH and PULL timing info
............................................................
2022-02-18T15:40:34 jackloop32.stanford.edu send: 0/0 Pull underruns: 0
PUSH -- SD avg/last: 0.195 / 0.242 	 mean/min/max: 0.666 / 0.001 / 3.694
PULL -- SD avg/last:  0.07 / 0.065 	         mean/min/max: 0.666 / 0.541 / 0.803 
 tot: 11924


cchafe:
looks like I need to mess with the clang check, sigh...

ntonnaett:
The Meson build has to be fixed first. There's something wrong with the Regulator class construction / destruction.

Linux nogui build fails too, although it shows up as passed: https://github.com/jacktrip/jacktrip/runs/5254962097?check_suite_focus=true
@dyfer We have a CI problem ;)

ntonnaett:
Why is the Regulator class inherting QObject? IIRC you removed that from PoolBuffer because it made problems.

Could you run clang-format on the changed files? https://jacktrip.github.io/jacktrip/DevTools/Formatting/

cchafe:
All set. 
However... there is a difference in performance between qmake and meson
versions which is audible and shows up in the stats (look at both sets of standard deviations -SD- and at Pull -mean/min/max-). This is on a pipewire client with 941ea89

Push stats reflect the incoming jitter
Pull stats are the internal audio engine callback timings

Both tests are using the GUI and this is the equivalent CLI call
jacktrip -C jackloop32.stanford.edu --bufstrategy 3 -I 1 --udprt

///////////////////////////////// qmake -- this is ok
./build
PIPEWIRE_LATENCY="32/48000" builddir/jacktrip

2022-02-19T09:04:27 jackloop32.stanford.edu send: 0/0 Pull underruns: 0
PUSH -- SD avg/last: 0.122 / 0.085 	 mean/min/max: 0.666 / 0.001 / 1.511
PULL -- SD avg/last: **0.089** / 0.091 	 mean/min/max: 0.666 / 0.529 / 0.744 
 tot: 50672

///////////////////////////////// meson -- audible artifacts
meson builddir
cd builddir; ninja
PIPEWIRE_LATENCY="32/48000" builddir/jacktrip

2022-02-19T09:07:01 jackloop32.stanford.edu send: 0/0 Pull underruns: 0
PUSH -- SD avg/last: 0.846 / 0.824 	 mean/min/max: 0.666 / 0.001 / 2.564
PULL -- SD avg/last: **0.888** / 0.864 	 mean/min/max: 1.115 / 0.014 / 2.663 
 tot: 50535

ntonnaett:
It's probably a comparison between release and debug build. Try `meson --buildtype=release builddir` when creating a build directory or `meson configure -Dbuildtype=release` in an existent build directory.

src/Regulator.h
```diff
@@ -51,12 +51,26 @@
 using std::vector;
```
ntonnaett, line 51: Not changed currently but could you remove this. It's not a good idea in a header file because it can change the behavior of external code.

dyfer:
> Linux nogui build fails too, although it shows up as passed: https://github.com/jacktrip/jacktrip/runs/5254962097?check_suite_focus=true
@dyfer We have a CI problem ;)

Yeah, this is weird. 
I'm guessing it's a problem with with the build script or with `make`, somehow we must be getting a `0` exit code despite failure or ???
Sorry, I won't have time to look closely at this in the next few days, but I will later if that's still an issue.

cchafe:
that was it -- built release version of meson and it's comparable in the same test
made the std:: modification

src/Regulator.h
```diff
@@ -40,23 +40,37 @@
...
-using std::vector;
-#include <math.h>
+
+//#define GUIBS3
+#ifdef GUIBS3
+#include "herlpergui.h"
+#include "ui_herlpergui.h"
+#endif
+
+#include <QWidget>
```
dyfer, line 60: Doesn't the `nogui` build fail because of this include? Is it needed?

cchafe:
good catch, hope that's the fix

src/Regulator.h
dyfer: seems this fixes the `nogui` build, thanks!

ntonnaett:
How to use clang-format with QtCreator (IIRC that's what you are using): https://www.vikingsoftware.com/blog/using-clang-format-with-qtcreator/

cchafe:
should be ready now, please feel free to approve and merge

src/Regulator.h
```diff
@@ -40,23 +40,37 @@
...
-    void train(vector<long double>& coeffs, const vector<float>& x);
-    void predict(vector<long double>& coeffs, vector<float>& tail);
+    void train(std::vector<long double>& coeffs, const std::vector<float>& x);
+    void predict(std::vector<long double>& coeffs, std::vector<float>& tail);
+
+   private:
+    // the following are class members to minimize heap memory allocations
+    std::vector<long double> Ak;
+    std::vector<long double> f;
+    std::vector<long double> b;
```
ntonnaett, line 73: Those will end up on the heap because the size is unknown. To be totally sure that this ends up on the stack is to use std::array with the maximum possible size. With C++20 std::span would probably be the way to go.

Nevertheless this shouldn't stay in the way of merging as this feature stays experimental.

src/JackTrip.cpp
```diff
@@ -615,8 +615,14 @@ void JackTrip::onStatTimer()
     }
     bool reset = (0 == pkt_stat.statCount);
     RingBuffer::IOStat recv_io_stat;
-    if (!mReceiveRingBuffer->getStats(&recv_io_stat, reset)) {
-        return;
+    if (getBufferStrategy() != 3) {
```
ntonnaett, line 618: This looks unnecessary

src/UdpDataProtocol.cpp
```diff
@@ -808,7 +808,12 @@ void UdpDataProtocol::receivePacketRedundancy(
             }
             src = dst;
         }
-        if (!mJackTrip->writeAudioBuffer(src, host_buf_size, gap_size)) {
+        int ok = true; // send audio buf to
+        ok = (mJackTrip->getBufferStrategy() !=3) ? // ring or jitter
```
ntonnaett, line 812: This can be one.

src/JackTrip.cpp
ntonnaett: Looking at how this is presented I realize now that it's not clear what I mean. The if and else blocks are the same.

src/UdpDataProtocol.cpp
cchafe: that differs in the last arg -- gap_size in the original case, last_seq_num in the new case

src/JackTrip.cpp
cchafe: thanks -- historical -- removed that

cchafe:
let's leave aside the vector vs. array vs. span stuff for now -- it's a deep change which might need a rework after going to mixed buffer sizes anyway

src/Regulator.h
cchafe: A quick look at this and the solution is going to be non-trivial

src/UdpDataProtocol.cpp
ntonnaett: I've meant `int ok = (mJackTrip…`. `true` in line 811 will be immediatley overridden. And it looks like it would be bool. Should it be bool?

ntonnaett (approved):

PR #496, opened 2022-02-28
RtAudio: handle names with commas in the command line

### LINKED ISSUE #495: RtAudio device names with commas are not handled
Originally reported over email. 
Example:
```sh
jacktrip -C cm-toast.stanford.edu -z --udprt --rtaudio --srate 48000 --bufsize 256 --audiodevice "Rogue Amoeba Software, Inc.: Loopback Audio 2","Rogue Amoeba Software, Inc.: Loopback Audio 1"
```

dyfer:
Fixes #495 

We had no way of specifying RtAudio devices on the command line if their names included commas, which we use as a delimiter between input and output device.
The issue is about distinguishing comma separating the devices and the comma that's part of the device name. 

When investigating this, I noted that when `--audiodevice "name, with comma","name, with comma"` is passed to JT, we get `name, with comma,name, with comma` (without quotes) from the command line parser, so matching `","` is not an option, AFAICT.

I think the solution is to have an alternative separator. I've chosen `\\,` (well technically it's `\,` but we need double escape in the shell). I'm open to suggestions if we want it to be something different - it just needs to be reasonably unlikely to also come up in the device name itself. One alternative I was considering was two or three commas in a row, but I'm not sure if that's better.

Additionally, I've changed parsing so that comma + space is treated as part of a single device name, so that `Rogue Amoeba Software, Inc.: undley-closse` should be handled properly when used as a single (input/output) device. 

If there is a device name that uses a comma but no space after comma, it can still be used but it needs to be specified as both input and output device using the alternative delimiter, e.g. `--audiodevice "device name,no space after comma"\\,"device name,no space after comma"`. Not ideal, but workable I think?

I've updated command line parsing, help output and "get command line" function in the gui. 

Any and all feedback is welcome. :)

### DIFF
--- a/src/Settings.cpp
+++ b/src/Settings.cpp
@@ -816,6 +816,7 @@ void Settings::printUsage()
          << endl;
     cout << " --audiodevice \"input-output device name\"" << endl;
     cout << " --audiodevice \"input device name\",\"output device name\"" << endl;
+    cout << " --audiodevice \"input device name, with comma\"\\\\,\"output device name, with comma\"" << endl;
     cout << "                                          Set audio device to use; if not set, "
             "the default device will be used"
          << endl;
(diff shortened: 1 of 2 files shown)


ntonnaett:
And does `--audiodevice "name\, with comma","name\, with comma"`  work? We should definitely stick to what you expect from the bash and `\\,` is not what you expect.

dyfer:
> Does this also happen if you use single quotes?

yes

> And does `--audiodevice "name\, with comma","name\, with comma"` work?

no

> We should definitely stick to what you expect from the bash and `\\,` is not what you expect.

I'm open to suggestions! 
We just need a delimiter that is not likely to be part of the device name. Or we'd need to figure out how to parse two arguments after long option (IIRC that was not possible for some reason).

dyfer:
After giving this more though, one other solution could be to move name -> device resolution to when the rtaudio backend is initialized and compare passed name to the available devices...

ntonnaett:
Wait. If we get the whole string, we can just don't touch what's inside quotes. I thought we have a problem with what the bash is doing before our parsing.

ntonnaett:
Understand. I already wandered why the bash didn't remove the quotes. The only clean solution I see here is to offer additional flags `--audiooutputdevice' and `--audioinputdevice`.

dyfer:
> The only clean solution I see here is to offer additional flags `--audiooutputdevice` and `--audioinputdevice`.

Got it. It's not a bad option, but of course a new one. Is this the route we want to take? I'd like everyone to be on board with this (@cchafe @psiborg112 ?) before I implement it.

ntonnaett:
I'm not sure how common this is, but if you have one input-output device with a comma in the name, this is broken too. So we might even want to remove the possibility to have two devices with `--audiodevice`.

dyfer:
> I'm not sure how common this is, but if you have one input-output device with a comma in the name, this is broken too. So we might even want to remove the possibility to have two devices with `--audiodevice`.

Just for the record - my PR sort of [handles the situation with a single comma](https://github.com/jacktrip/jacktrip/pull/496/files#diff-4694be0a587e1a04a36effbe84f9d99e36198406f7715778540aa8cbc766a521R888), as long as the device name has comma + space

OConnorStP:
I would mourn losing the ability to use different input and output devices.  it would break the setups of a couple people i've worked with in the last few months.

ntonnaett:
> I would mourn losing the ability to use different input and output devices. it would break the setups of a couple people i've worked with in the last few months.

Nobody wants to remove that.

psiborg112:
Yeah, as annoying as it is to have yet another command line option, I think it really is the cleanest way to handle this.

ntonnaett:
Closed in favor of #858

PR #510, opened 2022-03-17
Docs on running JackTrip with JACK server with custom name

### LINKED ISSUE #491: Support for multiple JACK servers
Currently JackTrip assumes that JACK server that user chooses is named “default”. This is standard behavior when custom name is not specified. It would be nice to be able to _choose_ which server to use: `jacktrip --jack-server-name myfancyname ...`.

My use case is one server to many clients. I’m working on an app on top of JackTrip and JACK. The thing is, that I can’t _isolate_ client from server in tests on one machine.

If you guys decide it is decent feature to add, I’m happy to contribute 👋

vrslev:
This PR adds docs about how one would approach running multiple JackTrip instances on one machine or using JACK server that have non-standard name.

Also I added [`venv` directory](https://packaging.python.org/en/latest/guides/installing-using-pip-and-virtual-environments/#creating-a-virtual-environment) to gitignore. Python is used for docs

Closes #491


docs/MultipleInstances.md
```diff
@@ -0,0 +1,10 @@
+# Running multiple instances
+
+If you want to setup multiple JackTrip instances on single machine, you have to use _different JACK servers_. Set `--name` flag for `jackd` flag and accordingly set `JACK_DEFAULT_SERVER` environment variable for JackTrip instances.
```
ntonnaett, line 3: You can run multiple JackTrip instances with the same JACK server. What was your use case that you tried this? One use case would be to use different audio interfaces for every jacktrip instance.

ntonnaett:
And thanks a lot for your PR! :)

docs/MultipleInstances.md
vrslev: Yep, I should be more clear about the purpose of running different JACK servers. Should I change main agenda to "Using JACK server with non-default name"?

ntonnaett: The idea of starting with real use cases is good. "Using JACK server with non-default name" is more on the solution side.
So starting with your use case of isolating JackTrip enviroments and the use case I mentioned about using different audio interfaces would be nice.
The heading could be **Using multiple JACK servers**. Maybe @dyfer has an opinion here?

dyfer: I think this document should first show running multiple jack instances with a single jack server first and also be explicit that this is only with the Jack backend, as opposed to the RtAudio backend. 
Then a subsection with a heading **Using multiple JACK servers** could explain that specific use case.

Also, I don't think we should be adding `venv` to the `.gitignore` - having python code inside this repository is probably not the best use case.

Finally, the PR should be aimed at `dev`.

Thanks!

docs/CustomJackServerName.md
```diff
@@ -0,0 +1,10 @@
+# Using custom JACK server name
+
+> ❕ This note is applicable only if you use JACK backend.
```
ntonnaett, line 3:
```suggestion
!!! attention ""

    This note is applicable only if you use JACK backend.
```

ntonnaett: Actually this can be removed completely. Right now you only add one part of what Marcin suggested and it clearly states JACK in the heading. I don't see the benefit of adding this note.

docs/CustomJackServerName.md
```diff
@@ -0,0 +1,10 @@
...
+
+In case you want to use JackTrip with JACK server _that has non-default name_ (by default it is `default`), you need to set `JACK_DEFAULT_SERVER` environment variable for JackTrip.
+
+1. Run `jackd` with `--name` flag:`jackd --name myfancyserver ...`,
+2. Run JackTrip with required environment variable set: `JACK_DEFAULT_SERVER=myfancyserver jacktrip ...`.
```
ntonnaett, line 8: The punctiuation isn't wrong (in German) but I think it looks better without.

ntonnaett (commented):
IMO I'm fine with only adding the custom jack server name page. I know the User Guide is lacking a lot but we have to start somewhere. I will approve this with those two little changes. Thanks a lot!

docs/CustomJackServerName.md
dyfer: Okay, you're right, with JACK in the heading that's clear. Sorry for the noise!

ntonnaett: Though it totally makes sense for a full fledged page (that we need…).

vrslev: @ntonnaett If this PR introduces a page with temporary solution, should I just close it?)

ntonnaett: No. I'm very happy somebody's starting this. I don't want to risk not getting anything started just because we want more documentation. We'll always want more documentation. And documentation will never be finished.

ntonnaett (approved):

PR #514, opened 2022-03-22
Plc auto

cchafe:
implements auto mode for bufstrategy 3
CLI syntax as with bufstrategy 1 but meaning is different
the example session below uses auto15 which specifies start with 15ms for bufstrategy 3 tolerance setting
during the first 2 seconds of connection statistics are gathered and 
then the 15ms gets modified up or down, servo'ed by longterm std dev and avg max
interpacket arrival

tested (fedora):

server: (at 64 FPP, stereo)
./jacktrip -S --bufstrategy 3 --udprt -p1 -q auto15 -I 1

client:
PIPEWIRE_LATENCY="64/48000" ./jacktrip -C <server>  --bufstrategy 3 --udprt -I 1 -n 1 -q auto15

### DIFF
--- a/src/Regulator.h
+++ b/src/Regulator.h
@@ -200,6 +200,7 @@ class Regulator : public RingBuffer
     int mFPPratioNumerator;
     int mFPPratioDenominator;
     int mPartialPacketCnt;
+    bool mAuto;
 #ifdef GUIBS3
     HerlperGUI* hg;
     void updateGUI(double msTol, int nSlots, int lostWin);
--- a/src/Regulator.h
+++ b/src/Regulator.h
@@ -95,10 +95,10 @@ class ChanData
 class StdDev
 {
    public:
-    StdDev(int w, int id);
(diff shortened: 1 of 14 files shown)


src/Regulator.h
```diff
@@ -112,12 +112,15 @@ class StdDev
...
+    int lastPlcUnderruns;
     double lastStdDev;
     double longTermStdDev;
     double longTermStdDevAcc;
+    double longTermMax;
+    double longTermMaxAcc;
+    double lastTime;
     int longTermCnt;
-    int plcUnderruns;
-    int lastPlcUnderruns;
```
ntonnaett, line 120: It looks like a lot of these could be private.

I would also like to see that we use mCamelCase for members everywhere. But this can be done automatically with clang-tidy. So don't start to change this manually.

ntonnaett (approved):
This looks okay to me.

PR #518, opened 2022-03-27
Release 1.5.3

ntonnaett:
Don't merge, Just approve!

### DIFF
--- a/.github/workflows/jacktrip.yml
+++ b/.github/workflows/jacktrip.yml
@@ -107,7 +107,7 @@ jobs:
             static-qt-version: 5.12.11
             qt-static-cache-key: 'v05'
             jacktrip-path: jacktrip
-            binary-path: binary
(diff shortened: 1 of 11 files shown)


cchafe (approved):

OConnorStP:
could we hold off on release at least until the call this Thursday?  i'd like to talk over a few ideas about how to expose the release process to a broader group of people.  we had a useful discussion about this several calls ago, but it seems to me there's some work left to go.  these tiny-releases are hard to sell right now, especially the way they're described.

ntonnaett:
The release already happened.

dyfer:
Sorry, haven't noticed, my bad.

ntonnaett:
The change we need for #496 is actually a breaking change. So we even have to discuss what version we call this then.

OConnorStP:
I saw it go by, but was hoping against hope that I'd misunderstood and that it was still in queue.  ;-)

PR #523, opened 2022-03-30
Fix signing before merging 1.5.3

mattahorton:


mattahorton:
Sorry. Chose the wrong repo.

PR #526, opened 2022-04-06
Fix include path to fcntl.h

ntonnaett:
When trying to build jacktrip on Alpine Linux for my Docker container I've got some warnings. One commit fixes the include path for fcntl.h.

The second commit doesn't fix the warning about non initialized struct members but changes the setRealtimeProcessPriority() function to be C on POSIX systems. Designated initializers are part of C since C99 but were introduced in C++ only with C++20. As this change doesn't fix anything I'm open to remove this. Happy to hear your opinion.

### DIFF
--- a/src/UdpDataProtocol.cpp
+++ b/src/UdpDataProtocol.cpp
@@ -52,7 +52,7 @@
 //#include <winsock.h>
 #include <winsock2.h>  //cc need SD_SEND
 #else
-#include <sys/fcntl.h>
+#include <fcntl.h>
 #include <sys/socket.h>  // for POSIX Sockets
 #include <unistd.h>
 #ifndef MANUAL_POLL


cchafe:
I'd be happy to test. I've been using gnome-boxes for virtualization. Please suggest an .iso and a set of packages on top of it that I'd need in order to try this fix out -- looking forward to trying Alpine.

ntonnaett:
I gave up on using alpine linux. My goal was to get a smaller container in comparison to Ubuntu. But I didn't get jacktrip to run and my actual bottleneck right now is ns3. I removed the extern "C" thingy because it didn't fix the warning. So this is just about the path to fcntl.h. This should be a no brainer as long as the CI build doesn't complain.

If you want to get jacktrip working on alpine linux, I would be more than happy to hear from your results. Alpine has a different toolchain to most common Linux distributions. Even ninja is replaced. The work would be to get the toolchain and jacktrip play nice with each other.

cchafe (approved):

PR #531, opened 2022-04-09
Plc asym auto fix

cchafe:
for eventual 1.5.4 release
server and client can have different FPP (tested from FPP 16 to 1024)
stress tested by repeatedly starting & stopping across full range of FPP's
server and client can have different in / out channel count
auto mode -- use -q auto
or for manual setting of initial mMsecTolerance -- use -q auto<msec>
gathers data for 6 sec and then goes full auto


ntonnaett:
var and mSampleRate are unused on macOS: https://github.com/jacktrip/jacktrip/runs/5956915133?check_suite_focus=true#step:17:207
If this is correct, can you mark them as [[maybe_unused]] please?

And we have merge conflicts…

ntonnaett (approved):

PR #532, opened 2022-04-11
fix faust compile

cchafe:
Fixes 
‘cerr’ is not a member of ‘std’ 
found with dev-test-faust-compile
-- does not fix warnings about unused parameters
-- noted that Faust compiler used for the currently checked-in C++ is very old (from 2017)

### DIFF
--- a/src/compressordsp.h
+++ b/src/compressordsp.h
@@ -3,8 +3,9 @@ author: "Julius Smith"
 license: "MIT Style STK-4.2"
 name: "compressor"
 version: "0.0"
-Code generated with Faust 2.28.6 (https://faust.grame.fr)
-Compilation options: -lang cpp -inpl -scal -ftz 0
+Code generated with Faust 2.40.0 (https://faust.grame.fr)
+Compilation options: -a faust2header.cpp -lang cpp -i -inpl -cn compressordsp -es 1 -mcd
+16 -single -ftz 0
 ------------------------------------------------------------ */
 
 #ifndef __compressordsp_H__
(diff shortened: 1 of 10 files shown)


cchafe:
fails to build -- anyone know why?

src/compressordsp.h
```diff
@@ -311,36 +381,40 @@ class dsp_factory
 #ifndef __meta__
 #define __meta__
 
+/**
+ The base class of Meta handler to be used in dsp::metadata(Meta* m) method to retrieve
+ (key, value) metadata.
+ */
 struct Meta {
-    virtual ~Meta(){};
+    virtual ~Meta() {}
```
ntonnaett, line 389: This looks sketchy.

ntonnaett: No. That's right. It's late…

src/makeXcodeproj.sh
```diff
@@ -1 +1 @@
-qmake -spec macx-xcode ../jacktrip.pro
+qmake - spec macx - xcode../ jacktrip.pro
```
ntonnaett, line 1: This doesn't look right. Is it clang-format on a shell script?

josmithiii: It works on my Mac to remove the extra spaces:

```qmake -spec macx-xcode ../jacktrip.pro```

josmithiii: But note that it generates ```jacktrip.xcodeproj``` in the src directory.

cchafe:
starting over

Issue #534, opened 2022-04-12
Overhaul of Faust dsp to c++ generation


cchafe:
First goal is to regenerate the existing features. Second is to add a new feature for live online automatic mastering written in Faust by a colleague.

Maybe someone has already done this successfully and it's just "operator error." Here's my failing incantation:
(failed with FAUST Version 2.40.1 -- latest dev)

git clone --recurse-submodules https://github.com/jacktrip/jacktrip.git
cd jacktrip/
git checkout dev
qmake-qt5
make release
./jacktrip -v
cd faust-src/
make h
make hi
cd ..
make release

I got some help from Julius and he fixed it. But clang-format undoes his fixes. "Immunizing" them by turning off clang-format at those lines works for formatting but then clang-tidy complains when the CI tests run.

ntonnaett:
I looked at how faust generates those header files. We are not able to clean this up. This has to be done in the faust project. So we have to live with what faust generates for us. We could just add all faust generated files to `.clang-format-ignore` and `.clang-tidy-ignore`. But this will introduce compiler warnings. We have to figure out how to silence specific warnings for those files, if the type of warning is okay to be ignored.

ntonnaett:
@josmithiii proposed adding faust as a git submodule (or similar). I think this is a good start tidying up this part of jacktrip and being more up to date with upstream faust. There are several parts to this:

- adding faust as a dependency (system installation of faust should be preferred)
- git submodule fallback for QMake
- FetchContent for CMake?
- Meson subproject (cmake subproject or adding a meson build to faust)
- Integrating faust-src into the build systems

ntonnaett:
@psiborg112, @dyfer, what are your thoughts about a faust dependency?

psiborg112:
I'd be ok with that.

ntonnaett:
We probably don't want to add a CMake dependency to the QMake build. So we must have pregenerated faust code. QMake then only depends on the headers in the faust submodule. If CMake is found on the system we can build faust and regenerate those files. Same of course if faust is found on the system.

cchafe:
Looking forward to a proper overhaul!
Part of the present issue has been fixed by #535 but still have a CI clang-format fail on the <dsp_name>dsp.h files that are generated. A workaround appears to be possible if we add a 
// clang-format off
as the first line of the file. 
That passes the clang-format CI test, at least.
Someone who knows makefile and sed could add a way to do that to faust-src/Makefile.headers (my makefile chops are limited).
[...]

PR #535, opened 2022-04-13
Move #include <iostream> from .cpp to .h files to hopefully keep clan…

josmithiii:
The theory is that clang-tidy insists Foo.cpp have #include "Foo.h" appear before all other includes.  If that's the rub, then this change should both work and be left alone by clang-tidy.

### DIFF
--- a/faust-src/zitarevdsp.dsp
+++ b/faust-src/zitarevdsp.dsp
@@ -2,7 +2,7 @@ import("stdfaust.lib");
 
 // Modified version from Faust Libraries demos.lib
 
-process = zita_rev1; //same as dm.zita_rev1 but for wetness control and some defaults
+process = zita_rev1; // same as dm.zita_rev1 but for wetness control and some defaults
 
 //process = zita_rev1 : _,attach(cout); // Not using this solution yet, but it works
 //cout = ffunction (int cout(), <iostream>, ""); // dummy function to force #include <iostream> in output
--- a/faust-src/zitarevdsp.dsp
+++ b/faust-src/zitarevdsp.dsp
@@ -2,7 +2,10 @@ import("stdfaust.lib");
 
 // Modified version from Faust Libraries demos.lib
 
-process = zita_rev1; // same as dm.zita_rev1 but for wetness control and some defaults
+process = zita_rev1; //same as dm.zita_rev1 but for wetness control and some defaults
+
+//process = zita_rev1 : _,attach(cout); // Not using this solution yet, but it works
+//cout = ffunction (int cout(), <iostream>, ""); // dummy function to force #include <iostream> in output
 
 //----------------------------------`(dm.)zita_rev1`------------------------------
 // Example GUI for `zita_rev1_stereo` (mostly following the Linux `zita-rev1` GUI).
--- a/src/Compressor.cpp
+++ b/src/Compressor.cpp
@@ -37,8 +37,6 @@
 
 #include "Compressor.h"
 
-#include <iostream>
-
 //*******************************************************************************
 void Compressor::compute(int nframes, float** inputs, float** outputs)
 {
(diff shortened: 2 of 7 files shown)


ntonnaett:
Having this order of header files is to ensure that header files are self contained. If the order of includes in your cpp file breaks your code, there must be something wrong with the included header files.

Looking at `Compressor.h`, an iostream include is indeed missing. But those implementations shouldn't be in the header file. So the better solution would be to put those implementations that use std::cout and std::cerr in the cpp file. In Visual Studio Code you get a fix suggestion for that which you can apply with a click.

josmithiii:
Thanks for the info.  I am not a Visual Studio user, but I do have it.  I tried opening jacktrip.pro and that didn't work.  Is there a way to get qmake to generate the VS project file?  (I have generated Xcode project files using qmake.)   I also have AppCode and CLion.  I have never done this kind of refactorization automatically in any app, but I would like to try it.  As for implementations in headers, I normally always put them there when the header will be included in only one compiled unit (saves typing, maintenance, possibly more inlining, etc.).  The only reason there's a Compressor.cpp is that I was following the pattern set by Juan-Pablo's LoopBack.h/cpp.  I don't know why it's written that way, but "when in Rome . . .".

ntonnaett:
I'm not familiar with Vistual Studio, only [Visual Studio Code](https://code.visualstudio.com/) or the free variant [Codium](https://vscodium.com/). I don't think you need to create a project. Just open your jacktrip directory. You might need to build it once. Never thought about it. I'm using Meson. There's a mesonbuild extension for vs code.

josmithiii:
I tried both Xcode and AppCode, and neither seems to offer this refactorization, as far as I can find.  (It would be nice to have for generating pure interface files from complete source.)  It therefore makes sense for someone who has Visual Studio Code to implement the single-click solution you mention, if that's desired.  I don't see any need for it myself, but I have no problem with someone else making that change.  Here's a related question: Is there a better suffix to use other than '.h' when the include-file is intended to contain inlined implementations?  If so, I haven't found it.

ntonnaett:
Some developers use .hpp for headers with templates. I don't see a reason to do so. I understand using .hpp for C++ headers and .h for C headers.

What is this PR supposed to fix? What is the problem with having the own header at the first place?

ntonnaett:
Oh, is it just because clang-tidy complains about newly faust generated C++ code?

josmithiii:
Thanks for the .hpp take, that's what I gleaned as well.  What I want is something like .cpp-include-file, and I'm willing to type all that!

My understanding is that clang-tidy reordered some header includes such that the Faust-generated C++ would no longer compile, so I moved the wrongly positioned header-includes from the .cpp file (which is where they belong, in my opinion, because they are really internal .cpp-include-file s) to the .h file, thereby defeating clang-tidy's broken meddling.  That's all this PR does.

As for the bigger problem, I've been chipping away at those compiler warnings and contributing the changes to the Faust project, but really someone on the Faust project should slog through all that.  I've fixed the ones that bothered me the most, and feel free to send me any fixed files to contribute, or make your own PR, etc.  Note that the Faust compiler last used to generate C++ for JackTrip was version 2.28, while version 2.40 is current.  I think most of my warning fixes happened in between.

ntonnaett (approved):
Are those faust generated files using iostream without including it? Clang-tidy does exactly what we want it to do. Nothing broken there. I approve this. But faust (might be our template file) has to be fixed at some point, if we want this regeneration process to be easy.

josmithiii:
Fair point.  However, we should upgrade to the latest Faust compiler before going off to find and fix that missing-include problem.  It might be fixed already.

Come to think of it, Faust should be a JackTrip submodule (or equivalent).

ntonnaett:
Let's move this general discussion to #534

PR #536, opened 2022-04-17
Clang format ignore dsp

ntonnaett:
Let's ignore faust's formatting.


cchafe (approved):
Fixes the CI problems  from Faust -- compile warnings are still there but that's for later
Thanks!

josmithiii:
We don't need to ignore faust-src/faust2header.cpp since that's us.
Ignoring src/*dsp.h is perfect for ignoring Faust-generated compilation warnings.

PR #543, opened 2022-04-27
Fix OpenSSL in build script (fixes #515)

### LINKED ISSUE #515: JackTrip "Unable to determine external IP address"
I've tried installing and changing settings and I can't get my IP address to show up in this version of Jacktrip, any tips? I'm running Windows 10

mattahorton:
Pulling from vcpkg instead of chocolatey gives us access to the static openssl libraries

### DIFF
--- a/.github/workflows/jacktrip.yml
+++ b/.github/workflows/jacktrip.yml
@@ -275,7 +275,9 @@ jobs:
         shell: bash
         run: |
           choco install jack --version=1.9.17 --no-progress
-          vcpkg install openssl --triplet="${{ matrix.vcpkg-triplet }}"
+          if [[ "${{ matrix.build-system }}" == "qmake" && -n "${{ matrix.static-qt-version }}" && -n "${{ matrix.vcpkg-triplet }}" ]]; then 
+            vcpkg install openssl --triplet="${{ matrix.vcpkg-triplet }}"
+          fi
           if [[ "${{ matrix.system-rtaudio }}" == true && -n "${{ matrix.vcpkg-triplet }}" ]]; then 
             choco install pkgconfiglite --no-progress
             # check out an older version of rtaudio library, see issue #408
--- a/.github/workflows/jacktrip.yml
+++ b/.github/workflows/jacktrip.yml
@@ -343,7 +343,7 @@ jobs:
           echo "installing Qt"
           make install
       - name: install openssl for Windows static builds
-        if: matrix.static-qt-version && steps.cache-static-qt.outputs.cache-hit != 'true' && runner.os == 'Windows'
+        if: matrix.static-qt-version && runner.os == 'Windows'
         shell: bash
         run: |
           if [[ -n "${{ matrix.vcpkg-triplet }}" ]]; then 
--- a/.github/workflows/jacktrip.yml
+++ b/.github/workflows/jacktrip.yml
@@ -144,6 +144,7 @@ jobs:
             bundled-rtaudio: true
             nogui: false
             weakjack: true
+            vcpkg-triplet: x64-mingw-static
             static-qt-version: 5.15.2
             qt-static-cache-key: 'v05'
             jacktrip-path: release/jacktrip.exe
--- a/.github/workflows/jacktrip.yml
+++ b/.github/workflows/jacktrip.yml
@@ -145,7 +145,7 @@ jobs:
             nogui: false
             weakjack: true
             static-qt-version: 5.15.2
-            qt-static-cache-key: 'v04'
+            qt-static-cache-key: 'v05'
             jacktrip-path: release/jacktrip.exe
             binary-path: binary
             installer-path: installer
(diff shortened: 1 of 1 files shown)


mattahorton:
> It seems the TLS situation on windows makes it necessary to think about checking the version number if there are any vulnerabilities.

I'm gonna be real with you that I wouldn't know where to start on that. This PR was a _journey_ for me. Happy to learn if you point me in the right direction.

mattahorton:
@ntonnaett looks like this is openssl 3.0.2 and the [OpenSSL site lists that as the latest patch for vulnerabilities.](https://www.openssl.org/news/vulnerabilities-3.0.html)

dyfer:
Re: vulnerabilities - IIUC this is about making sure that we use the latest openssl as opposed to relying on cache?
IIRC It does seem that release builds do not, after all, use cached qt, so I believe they should always rebuild qt and thus pull the latest ssl from vcpkg

.github/workflows/jacktrip.yml
```diff
@@ -275,7 +276,9 @@ jobs:
         shell: bash
         run: |
           choco install jack --version=1.9.17 --no-progress
-          choco install openssl --no-progress
+          if [[ "${{ matrix.build-system }}" == "qmake" && -n "${{ matrix.static-qt-version }}" && -n "${{ matrix.vcpkg-triplet }}" ]]; then 
+            vcpkg install openssl --triplet="${{ matrix.vcpkg-triplet }}"
+          fi
```
dyfer, line 281: Is this only needed when building Qt (i.e. not needed when cached qt build is used)? If so, we could consider moving this to the "build qt" step, which only runs if qt cache is not available. 
It's not a request, rather a suggestion for consideration. What do you think?

ntonnaett:
I didn't even think about the build process. Good point. My concern was about leaving people with old versions/installations that have known vulnerabilities.

Theoretically users could just copy a new version of openssl.dll to the jacktrip folder if it's dynamically linked. But nobody does this. So the only solution is to remind people in the app to update. We should talk about this in the dev meeting. I don't know what the best approach is for the implementation.

.github/workflows/jacktrip.yml
mattahorton: I am also feeling this just in development on features! I think this is the right move.

.github/workflows/jacktrip.yml
```diff
@@ -342,16 +342,25 @@ jobs:
           # install
           echo "installing Qt"
           make install
+      - name: install openssl for Windows static builds
+        if: matrix.static-qt-version && steps.cache-static-qt.outputs.cache-hit != 'true' && runner.os == 'Windows'
```
mattahorton, line 346: @dyfer I created a new step since I didn't want to rewrite it in cmd

dyfer:
I'll have access to my Windows machine in a few hours and plan to review (test) then.

.github/workflows/jacktrip.yml
```diff
@@ -342,16 +342,25 @@ jobs:
           # install
           echo "installing Qt"
           make install
+      - name: install openssl for Windows static builds
+        if: matrix.static-qt-version && runner.os == 'Windows'
```
dyfer, line 346: I noticed this step still ran when cached qt was used. IIUC this should be the same condition as for `build static Qt on Windows`, right?
```yaml
matrix.static-qt-version && steps.cache-static-qt.outputs.cache-hit != 'true' && runner.os == 'Windows'
```

dyfer (changes requested):

psiborg112:
Yeah, have to agree. If windows had OpenSSL installed somewhere where it could be reliably accessed system wide (as it is on other platforms) then dynamic linking would make sense. But we're far more likely to be able to convince users to run an upgraded installer than to copy things around their file system.

dyfer:
I've run the build on my Windows machine and it seems to be working fine. I can see the external IP address. 

I just have one request regarding the github actions config (see above).

ntonnaett:
> Yeah, have to agree. If windows had OpenSSL installed somewhere where it could be reliably accessed system wide (as it is on other platforms) then dynamic linking would make sense. But we're far more likely to be able to convince users to run an upgraded installer than to copy things around their file system.

There's schannel. But I think it was introduced with Windows 10. https://docs.microsoft.com/en-us/windows/win32/secauthn/transport-layer-security-protocol

psiborg112:
I don't think Qt can be built to use it as a replacement for OpenSSL.

.github/workflows/jacktrip.yml
mattahorton: It still needs to build JackTrip with those library files. The build was failing when I included the cache hit requirement.

dyfer: I see, thank you for clarifying. In that case, does it still make sense to have it as an independent install step?

mattahorton: probably not 😅

mattahorton: Changed back

dyfer (approved):
Looks good to me, thanks!

PR #545, opened 2022-05-05
Virtual Studio

mattahorton:
Hey everyone, this is a big PR so I'm embedding a couple videos here to show exactly what has been done. That said, this adds an optional Virtual Studio GUI to JackTrip. Users get a choice between the original GUI or the new Virtual Studio GUI on their first launch and JackTrip will stick to that choice in the future. 

Most of this work was done by @psiborg112. I mostly helped with build scripts and OpenSSL, but I'm making the PR so we can talk about it at dev meetings.

Something I know we'll want to talk about is that first launch screen, considering every JackTrip user will encounter that screen, and you all should know that what is in the video is not the latest version. I know there was some discussion at the last dev meeting and there should be more today.

Anyway, I'm know we'll be adding more to this PR before merge, but happy to start talking about it!

https://user-images.githubusercontent.com/5567285/166968501-644d3aab-d101-4e4b-a5e2-41377d88a3a0.mov

https://user-images.githubusercontent.com/5567285/166968522-9f1ac975-7123-43c3-ac35-2c6e07965c16.mov


jacktrip.pro
```diff
@@ -22,10 +22,17 @@ equals(QT_EDITION, "OpenSource") {
 
 nogui {
   DEFINES += NO_GUI
-  QT -= gui
```
ntonnaett, line 25: We had a bug where our nogui builds were linked to qtgui. Can you check that we don't regress here?

jacktrip.pro
```diff
@@ -56,7 +63,7 @@ INCLUDEPATH += faust-src-lair/stk
   weakjack {
     message(Building with weak linking of JACK)
     INCLUDEPATH += externals/weakjack
-    DEFINES += USE_WEAK_JACK
+    DEFINES += USE_WEAK_JACK NDEBUG
```
ntonnaett, line 66: This looks like if we build with weakjack we also disable all asserts. I think somebody should open an issue for Windows Terminal Preview instead of this.

meson.build
```diff
@@ -79,7 +79,7 @@ endif
 if get_option('nogui') == true
 	defines += '-DNO_GUI'
 	qt5_dep = dependency('qt5', modules: ['Core', 'Network'], include_type: 'system')
-else
+elif get_option('novs') == true
```
ntonnaett, line 82: The NO_VS macro must be added to defines.

src/gui/qjacktrip.cpp
```diff
@@ -34,7 +34,10 @@
 #include <ctime>
 
 #include "about.h"
-#ifdef NO_JTVS
+#ifndef NO_VS
+#include "virtualstudio.h"
+#endif
+#ifdef PSI
```
ntonnaett, line 40: Where does PSI come from?

src/gui/virtualstudio.cpp
```diff
@@ -0,0 +1,1054 @@
...
+                QDateTime::currentDateTimeUtc().addSecs(60 * 60).toString(Qt::ISODate);
+            QJsonObject json      = {{QLatin1String("enabled"), true},
+                                {QLatin1String("expiresAt"), expiration}};
+            QJsonDocument request = QJsonDocument(json);
+
+            QNetworkReply* reply = m_authenticator->put(
+                QStringLiteral("https://app.jacktrip.org/api/servers/%1")
+                    .arg(studioInfo->id()),
+                request.toJson());
+            connect(reply, &QNetworkReply::finished, this, [=]() {
```
ntonnaett, line 470: Do we need to capture by value here or should we capture by reference?

src/gui/virtualstudio.cpp
```diff
@@ -0,0 +1,1054 @@
...
+
+void VirtualStudio::completeConnection()
+{
+    m_jackTripRunning = true;
+    m_connectionState = QStringLiteral("Connecting...");
+    emit connectionStateChanged();
+    VsServerInfo* studioInfo = static_cast<VsServerInfo*>(m_servers.at(m_currentStudio));
+    try {
+        m_jackTrip.reset(new JackTrip(JackTrip::CLIENTTOPINGSERVER, JackTrip::UDP, 2, 2,
+#ifdef WAIR  // wair
```
ntonnaett, line 505: Do we even support "nowhere"?

src/gui/virtualstudio.cpp
```diff
@@ -0,0 +1,1054 @@
...
+        return;
+    }
+
+    m_jackTripRunning = false;
+    m_connectionState = QStringLiteral("Disconnected");
+    emit connectionStateChanged();
+    emit disconnected();
+    m_onConnectedScreen = false;
+
+    // TODO; Fix up these next three functions.
```
ntonnaett, line 698: What are these next three functions? Done?

src/gui/virtualstudio.cpp
```diff
@@ -0,0 +1,1054 @@
...
+    std::cout << "Received connection" << std::endl;
+    emit connected();
+}
+
+void VirtualStudio::checkForHostname()
+{
+    VsServerInfo* studioInfo = static_cast<VsServerInfo*>(m_servers.at(m_currentStudio));
+    QNetworkReply* reply     = m_authenticator->get(
+            QStringLiteral("https://app.jacktrip.org/api/servers/%1").arg(studioInfo->id()));
+    connect(reply, &QNetworkReply::finished, this, [=]() {
```
ntonnaett, line 734: capture by reference?

src/gui/virtualstudio.cpp
```diff
@@ -0,0 +1,1054 @@
...
+
+    // Get the serverId of the server at the top of our screen if we know it
+    QString topServerId;
+    if (index >= 0 && index < m_servers.count()) {
+        topServerId = static_cast<VsServerInfo*>(m_servers.at(index))->id();
+    }
+
+    QNetworkReply* reply =
+        m_authenticator->get(QStringLiteral("https://app.jacktrip.org/api/servers"));
+    connect(reply, &QNetworkReply::finished, this, [=]() {
```
ntonnaett, line 827: [&]?

src/main.cpp
```diff
@@ -214,7 +225,11 @@ int main(int argc, char* argv[])
     QScopedPointer<JackTrip> jackTrip;
     QScopedPointer<UdpHubListener> udpHub;
 #ifndef NO_GUI
-    QScopedPointer<QJackTrip> window;
+    QSharedPointer<QJackTrip> window;
```
ntonnaett, line 228: Why is this a shared pointer?

ntonnaett (commented):
That's a big chunk of work. Looks nice!

docs/Build/Linux.md
```diff
@@ -16,7 +16,7 @@ Optional:
 
 ### Fedora
 ```sh
-dnf install qt5-devel
+dnf install qt5-devel qt5-qtnetworkauth-devel qt5-qtquickcontrols2-devel qt5-qtsvg-devel
```
ntonnaett, line 19: Funny 'qt5-devel' doesn't exist. Probably my mistake. Could you change that to 'qt5-qtbase-devel'?

.github/workflows/jacktrip.yml
```diff
@@ -61,13 +61,13 @@ jobs:
...
             release-name: Linux-x64-nogui
-            runs-on: ubuntu-18.04
+            runs-on: ubuntu-20.04
             system-rtaudio: false
             bundled-rtaudio: false
             nogui: true
             weakjack: false
-            static-qt-version: 5.12.11
-            qt-static-cache-key: 'v22' # we use the same key as the main Linux build
+            static-qt-version: 5.15.2
```
dyfer, line 69: The reason for using Qt 5.12 initially was to make the build compatible with Ubuntu 20.04 (the latest LTS version at the time of this addition). Static qt 5.15 did not run on this version when I last tested it. If possible, I think it would be preferable to keep building with 5.12 to support older OSs...

ntonnaett:
IMO there should be a back button instead of the cancel button on the Settings page.

If I drag and hold the window the window decoration looks like it loses focus.

ntonnaett:
With the meson build I get these:
```
qrc:/vs/Connected.qml:36: TypeError: Cannot read property 'canConnect' of undefined
qrc:/vs/Connected.qml:35: TypeError: Cannot read property 'isManageable' of undefined
qrc:/vs/Connected.qml:34: TypeError: Cannot read property 'isPublic' of undefined
qrc:/vs/Connected.qml:33: TypeError: Cannot read property 'name' of undefined
qrc:/vs/Connected.qml:32: TypeError: Cannot read property 'flag' of undefined
qrc:/vs/Connected.qml:31: TypeError: Cannot read property 'location' of undefined
```

The ./build build doesn't quit cleanly.

CMakeLists.txt
```diff
@@ -51,6 +57,7 @@ if (${CMAKE_SYSTEM_NAME} MATCHES "Linux" OR ${CMAKE_SYSTEM_NAME} MATCHES "Darwin
   endif ()
 endif ()
 
+set_property(SOURCE src/Regulator.h PROPERTY SKIP_AUTOGEN ON)
```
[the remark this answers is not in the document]
mattahorton: Is this some weird git shenanigans from me merging dev into the branch? I know @cchafe just worked on faust code.

src/gui/virtualstudio.cpp
ntonnaett: I've got following with scan-build. Maybe they are related.
```
[68/71] Compiling C++ object jacktrip.p/src_gui_virtualstudio.cpp.o
../../../src/gui/virtualstudio.cpp:840:33: warning: Access to field 'm_refreshInProgress' results in a dereference of a null pointer [core.NullDereference]
            m_refreshInProgress = false;
            ~~~~~~~~~~~~~~~~~~~~^~~~~~~
../../../src/gui/virtualstudio.cpp:920:14: warning: Access to field 'm_allowRefresh' results in a dereference of a null pointer [core.NullDereference]
        if (!m_allowRefresh) {
             ^~~~~~~~~~~~~~
2 warnings generated.
```

CMakeLists.txt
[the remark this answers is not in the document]
ntonnaett: IIRC the fix by @cchafe is already in dev. But I don't see this line on dev.

src/gui/virtualstudio.cpp
ntonnaett: That's probably a copy and paste thing. We can remove those later everywair.

.github/workflows/jacktrip.yml
mattahorton: I thiiiiink this change was a mistake. A `nogui` build shouldn't need any Qt updating. Going to double check.

docs/Build/Linux.md
mattahorton: For sure. Looks like it's live there right now, but we can fix that in this PR.

jacktrip.pro
mattahorton: Yep. Will double check.

.github/workflows/jacktrip.yml
mattahorton: I will say there are Qml features being used (required properties, as an example) that require 5.15 for VS builds. 

I know you're asking specifically about Linux `nogui` builds, but here's some added context.

5.15 still supports Windows 7, but it drops support for macOS Sierra, which when I looked into it was estimated to be used by about 2% of the existing active macOS install base. That data didn't include Monterey which means it is probably less than that.

mattahorton:
> IMO there should be a back button instead of the cancel button on the Settings page.
> 
> If I drag and hold the window the window decoration looks like it loses focus. Edit: It looks like an issue about updating the decoration. The close button doesn't animate or is stuck. That's only in virtual studio mode. When switching to the old interface the window decoration behaves correctly.

Which platform do you mean here, @ntonnaett?

.github/workflows/jacktrip.yml
ntonnaett: Would be great to switch to 5.15. #320 was the PR that downgraded to 5.12. @dyfer Do you remember what the compatibility problem was? Maybe we can fix it?

ntonnaett:
That's Fedora 35 with Gnome on wayland.

.github/workflows/jacktrip.yml
ntonnaett: [This](https://github.com/jacktrip/jacktrip/pull/320#issuecomment-857121580) looks like the the static xcb plugin was missing.

mattahorton:
Just made a mistake push to the wrong remote. Sorry

.github/workflows/jacktrip.yml
ntonnaett: @dyfer Do you remember if you tried [adding qxcb or similar to QTPLUGIN_platforms](https://doc.qt.io/qt-5/plugins-howto.html#static-plugins)?

jacktrip.pro
mattahorton: Added this line back with no issues

.github/workflows/jacktrip.yml
mattahorton: Looks like installing `libxcb-util-dev` fixed it for some people. [Seems like a hidden requirement](https://forum.qt.io/topic/125582/qt-5-15-2-do-not-build-libqxcb-so-but-at-the-same-time-qt-5-14-0-successfully-build-libqxcb-so). This is 18.04, but worth a shot.

mattahorton: This is _extremely_ fresh 20.04 (booted off usb) attempting to run the static shared build with rtaudio. I had to install `libxcb-xinerama0-dev` and it still loads to a blank screen for 5.12/5.15 incompatibility reasons (see last screenshot).

Screenshot from 2022-05-05 22-29-53
Screenshot from 2022-05-05 22-29-21
Screenshot from 2022-05-05 22-35-03

cchafe (approved):

CMakeLists.txt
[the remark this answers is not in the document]
psiborg112: It was a workaround for when it was being moced for my builds. Might not be needed anymore, but doesn't really affect things anyway, since I don't think anyone other than me is using CMake. (It's certainly not supported in any official capacity.)

src/main.cpp
psiborg112: Yep

mattahorton:
Did we mean to merge this yet?

ntonnaett:
That was a little bit early. But maybe it's nice having the remaining comments as separate issues.

PR #554, opened 2022-05-06
qmake: remove wrong include paths

ntonnaett:
They were annoying me for a long time…

### DIFF
--- a/jacktrip.pro
+++ b/jacktrip.pro
@@ -52,9 +52,6 @@ nojack {
   DEFINES += NO_JACK
 }
 
-# for plugins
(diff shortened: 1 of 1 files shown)


dyfer (approved):
Thanks!

dyfer:
I think the current issue building Linux with static qt on dev is due to outdated cached qt (as we changed the build OS version). It'll either go away with another merge, or we need to trigger rebuilding Qt by changing [static cache key](https://github.com/jacktrip/jacktrip/blob/dev/.github/workflows/jacktrip.yml#L43).

PR #557, opened 2022-05-08
WIP/Experimental: Qt6 port

ntonnaett:
- [ ] HiDPI changes
- [ ] fix blank screen when window size changes
- [ ] fix non draggable scrollbar

### DIFF
--- a/.github/workflows/flatpak-devel.yml
+++ b/.github/workflows/flatpak-devel.yml
@@ -8,7 +8,7 @@ jobs:
     name: "Flatpak"
     runs-on: ubuntu-latest
     container:
-      image: bilelmoussaoui/flatpak-github-actions:kde-5.15
+      image: bilelmoussaoui/flatpak-github-actions:kde-6.2
       options: --privileged
     steps:
     - uses: actions/checkout@v2
--- a/.github/workflows/jacktrip.yml
+++ b/.github/workflows/jacktrip.yml
@@ -237,7 +237,7 @@ jobs:
           else
             sudo add-apt-repository universe
             sudo apt-get update
-            sudo apt-get install --yes qt6-default qt6-qmake qttools6-dev libqt6networkauth6-dev qtdeclarative6-dev libqt6svg6-dev
+            sudo apt-get install --yes qt6-base-dev qt6-base-dev-tools qt6-qmake qt6-tools-dev libqt6networkauth6-dev qt6-declarative-dev libqt6svg6-dev
           fi
           if [[ "${{ matrix.system-rtaudio }}" == true ]]; then 
             sudo apt-get install --yes librtaudio-dev
--- a/.github/workflows/jacktrip.yml
+++ b/.github/workflows/jacktrip.yml
@@ -237,7 +237,7 @@ jobs:
           else
             sudo add-apt-repository universe
             sudo apt-get update
-            sudo apt-get install --yes qt6-base-dev qt6-base-dev-tools qt6-qmake qt6-tools-dev libqt6networkauth6-dev qt6-declarative-dev libqt6svg6-dev
+            sudo apt-get install --yes qt6-base-dev qt6-base-dev-tools qmake6 qt6-tools-dev qt6-declarative-dev
           fi
           if [[ "${{ matrix.system-rtaudio }}" == true ]]; then 
             sudo apt-get install --yes librtaudio-dev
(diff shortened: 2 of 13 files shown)


dyfer:
Do you consider having preprocessor switches so that it still builds with Qt5? 
I would like that, but 1) we can add switches in a separate PR if you don't want to do it yourself and/or 2) if everyone is happy to "live in the future and forget the past" (and drop Qt5), I can also accept that.

psiborg112:
Given how long it might take qt6 to make it into some stable releases (thinking of Debian here), it would probably make sense to use preprocessor switches.

psiborg112:
That's going to be a problem for the qml files though. What we might have to do is have different versions of the qml files for qt5 and qt6. Then all we have to do is have a preprocessor directive in virtualstudio.cpp that loads either vs.qml or vs6.qml depending on what we're building for.

ntonnaett:
I'm not sure if it is worth the effort. Qt5 support would only be needed for old Linux distributions. So we would need to decide what we want to support.

dyfer:
> Qt5 support would only be needed for old Linux distributions.

FYI Qt5 support is also needed for macOS 10.13.

ntonnaett:
> > Qt5 support would only be needed for old Linux distributions.
> 
> FYI Qt5 support is also needed for macOS 10.13.

macOS 10.13 got its last security updates in December 2020. People should update. I'm not a mac user but how old must your hardware be so you can't update? Honest question. I don't want to obsolete working hardware.

ntonnaett:
If you use High Sierra 10.13, you can probably update to Catalina 10.15:
https://support.apple.com/en-gb/HT210222
Or Big Sur:
https://support.apple.com/en-gb/HT211238

I hoped to see older models…

dyfer:
> I'm not a mac user but how old must your hardware be so you can't update?

The cutout varies between various models (macbook vs macbook pro vs imac etc) but it's roughly 2011 (ie most machines released 2012 and later can run 10.14)... This is quite old, but it seems some people still use them, as it came up in the other thread.

ntonnaett:
It might be easy to have a qt version switch. Right now I have no time to work on this. So if anybody wants to steel this PR feel free.

OConnorStP:
Here's the wikipedia article that I used to look up these dates - https://en.wikipedia.org/wiki/MacOS

High Sierra (10.13) doesn't support hardware older than 2009

Mojave (10.14) doesn't support hardware older than 2012

I have a handful of High Sierra Jacktrip users.  I'm sure there are lots of machines that are that old in educational institutions...

Could we just keep building a "special case" version of JT for those oldsters and park it off to the side somewhere?

dyfer:
> Could we just keep building a "special case" version of JT for those oldsters and park it off to the side somewhere?

Part of the discussion here was whether to maintain the ability to build with Qt5 (needed to make High Sierra compatible builds) or to enforce building with Qt6. Keeping the ability to build with both versions requires some extra work in comparison with just moving on to Qt6. 
From the discussion so far it seems that we might consider maintaining compatibility with both versions once work on this is resumed.

mattahorton:
> > Could we just keep building a "special case" version of JT for those oldsters and park it off to the side somewhere?
> 
> Part of the discussion here was whether to maintain the ability to build with Qt5 (needed to make High Sierra compatible builds) or to enforce building with Qt6. Keeping the ability to build with both versions requires some extra work in comparison with just moving on to Qt6. From the discussion so far it seems that we might consider maintaining compatibility with both versions once work on this is resumed.

Could we not point those users to older versions? This is a common tactic for this problem, and it's not like we're saying you can't download them.

dyfer:
I would like to reiterate my preference to keep compatibility with both versions when porting to Qt6... I'm happy to put ifdefs wherever necessary to make that happen :)

psiborg112:
Just wanted to second what @dyfer said about ifdefs, especially since I don't think we're actually going to end up needing too many of them. As a test, I just built a NO_VS build on Debian with Qt6 and it seemed to work without any modification. (Except for the exit button - would need to look into that.) So I think there are large parts of the code (particularly the core components of it) that won't really need to be modified.

dyfer:
In line with what Aaron observed, IIUC it should be possible to use a single codebase to target Qt 5.15 and 6 (save for build system particularities). So we'd only need some ifdefs to keep supporting older Qt (currently we use 5.12 in our Linux build to maintain compatibility with older systems).

PR #558, opened 2022-05-09
Move static Linux builds to 5.15

mattahorton:
Added missing dependencies and updated flags for the 5.15 world. Creating the PR before testing so we can rely on Windows and Mac static qt caches for build times.

I can tell you from early testing that 5.15 worked with minimal installs necessary on 20.04. If these flags truly do bundle those libraries with the binary, then this should work for us.

### DIFF
--- a/.github/workflows/jacktrip.yml
+++ b/.github/workflows/jacktrip.yml
@@ -40,7 +40,7 @@ jobs:
             # vcpkg-triplet: x64-mingw-static # used when installing rtaudio with vcpkg (Windows)
             # qt-arch: 'win64_msvc2019_64' # specify when using shared (official) qt libraries (Windows)
             static-qt-version: 5.15.2 # if set, it will enable static Qt build
-            qt-static-cache-key: 'v26' # required for static qt; update this to force rebuilding static Qt
+            qt-static-cache-key: 'v27' # required for static qt; update this to force rebuilding static Qt
             jacktrip-path: jacktrip # needed for binary upload
             binary-path: binary # directory relative to build path; triggers artifact upload for jacktrip binary
             # bundle-path: bundle # directory relative to build path; triggers application bundle creation and upload (macOS)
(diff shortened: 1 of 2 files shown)


psiborg112:
Thanks for taking this on Matt. Was going to look at it but haven't had any time recently.

mattahorton:
> Thanks for taking this on Matt. Was going to look at it but haven't had any time recently.

No problem. I'm thinking of building with the `-no-xinerama` flag. We have xrandr which is built in to modern distros. That _should_ cover us for multi-monitor.

mattahorton:
It looks like the core problem here is `libxcb-xinerama` which is 1. outdated and not necessary, 2. only a requirement for 5.15 (not 5.12 or 6.0+), and 3. broken in 5.15. Depending on how some of these builds go, I'll try the `-no-xinerama` flag, but reading up on it, that is also broken, apparently.

ntonnaett:
Do you really have to statically link all those libxcb* libraries? I thought they are available on all distributions by default. But I might be wrong.

mattahorton:
> Do you really have to statically link all those libxcb* libraries? I thought they are available on all distributions by default. But I might be wrong.

Going off of the discussion in the older issue, that was my original approach yes.

However I discovered that Xinerama is broken in Qt 5.15 and will likely be trying to just build without it with the other libraries dynamically linked. Because you're right and I've learned that from working on this.

ntonnaett:
I think we don't want the wayland plugin right now with the static build. With flatpak I default to xcb until we switch to qt6.

mattahorton:
> I think we don't want the wayland plugin right now with the static build. With flatpak I default to xcb until we switch to qt6.

I agree. Just following error logs at the moment. Rolling a bunch of this back anyway.

mattahorton:
This is _infuriating_. 

I managed to build without the single xcb package that isn't in 20.04 and then 3 new not found libraries popped up:

> libicui18n.so.60 => not found 
libicuuc.so.60 => not found 
libpcre2-16.so.0 => not found

People are saying that reinstalling qt5 on 20.04 solves this, but that's worse than just having people install xinerama!

mattahorton:
@psiborg112 Have any thoughts, since you were thinking about looking at this? My only thought right now is to set a default value on the required property for section headers and move back to 5.12.11 until we can move to Qt6. Qt6 removed the xinerama requirement, so that shouldn't be an issue going forward.

ntonnaett:
I never looked into it but maybe [the opensuse build service](https://openbuildservice.org/) is the right thing for us. It can generate packages for all major distributions from the same source. Then we could ditch the static build for Linux.

PR #559, opened 2022-05-12
Adapt QML for 5.12 on Linux

mattahorton:
This seems to be the one issue in the code that is keeping things from running on 5.12. I have tried to move our builds to 5.15 unsuccessfully and am going to wait until we move to qt6 to continue down that path. 

For now, here is a fix. 

Going to test this before upgrade to a full PR.

### DIFF
--- a/.github/workflows/jacktrip.yml
+++ b/.github/workflows/jacktrip.yml
@@ -40,7 +40,7 @@ jobs:
             # vcpkg-triplet: x64-mingw-static # used when installing rtaudio with vcpkg (Windows)
             # qt-arch: 'win64_msvc2019_64' # specify when using shared (official) qt libraries (Windows)
             static-qt-version: 5.12.11 # if set, it will enable static Qt build
-            qt-static-cache-key: 'v23' # required for static qt; update this to force rebuilding static Qt
+            qt-static-cache-key: 'v24' # required for static qt; update this to force rebuilding static Qt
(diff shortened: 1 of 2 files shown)


mattahorton:
Ubuntu 20.04 fresh install. Works just fine. 

Screenshot from 2022-05-12 13-45-40

mattahorton:
I did test the nogui static build as well and it also ran fine.

dyfer:
Thanks! This works well. I'm glad it still builds with 5.12

So is the addition of `libxcb-xinerama0-dev` necessary? was that missing before?

mattahorton:
> Thanks! This works for me on my Ubuntu 20.04. I'm glad JT still builds with 5.12.
> 
> So is the addition of `libxcb-xinerama0-dev` necessary? was that missing before?

I never quite figured this out, but something about the way our builds changed when we merged the Virtual Studio branch in caused `libxcb-xinerama0-dev` to be dynamically linked in a way it wasn't before. It's probably that Quick/QML requires either it or Xrandr to be present. Most distros come use Xrandr for multiscreen support and not Xinerama these days, so all I needed to do for 5.12 was make sure we _didn't_ have `libxcb-xinerama0-dev` installed on the build system so that the configure script didn't autodetect and link it.

Well, that and adapt the QML to work on 5.12.

dyfer (approved):
Thanks!

PR #560, opened 2022-05-12
VS: New First Launch screen

mattahorton:
This is the first launch screen we talked about in the last dev meeting.

Screen Shot 2022-05-12 at 1 22 59 PM


ntonnaett:
What I would like is an association of those buttons to Virtual Studio Mode and Classic Mode. Did we decide for "Classic"?

mattahorton:
> What I would like is an association of those buttons to Virtual Studio Mode and Classic Mode. Did we decide for "Classic"?

We did, but I thought it after that discussion that this screen would be a yes or no question and that buttons in the interfaces would reference Classic mode and Virtual Studio mode.

ntonnaett:
I'm totally for the yes or no buttons. I thought of something like smaller labels below Yes and No.

mattahorton:
> I'm totally for the yes or no buttons. I thought of something like smaller labels below Yes and No. Or just emphasize Virtual Studio and Classic in the list below.

The context of the discussion was that we wanted to avoid branding either experience on this page, I thought. We knew we couldn't really get away from it when someone needs to switch. Otherwise we would have gone with one of the ones that had those names more prominent in the UI.

mikedickey:
Looks like "Virtual Studios" is already in the bullet list for "Yes". How about we just change the text on the "No" side to "The Classic JackTrip experience"

mikedickey (approved):

PR #566, opened 2022-05-14
Renaming buttons on settings page

nwang92:
Screen Shot 2022-05-14 at 2 05 50 PM

### DIFF
--- a/src/gui/Settings.qml
+++ b/src/gui/Settings.qml
@@ -231,38 +231,38 @@ Item {
         border.color: "#33979797"
 
         Button {
-            id: backButton
+            id: cancelButton
             background: Rectangle {
                 radius: 6 * virtualstudio.uiScale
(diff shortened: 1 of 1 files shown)


ntonnaett:
Could you please add an explanation why this should be changed (again)! IMO there should only be a back button and changes should get applied immediately. I'm get used to that paradigm more and more.

mikedickey (approved):

nwang92:
> Could you please add an explanation why this should be changed (again)! IMO there should only be a back button and changes should get applied immediately. I'm getting used to that paradigm more and more.

Sorry, I should've added some more info the PR! I believe this was a request from some early customer feedback where "ok" and "back" were confusing as to whether the settings got applied or not. I don't think I'm skilled enough yet to remove the save button but this seemed like a good intermediate step :)

PR #567, opened 2022-05-16
More user friendly confirmation of Virtual Studio login success

mikedickey:
After logging into Virtual Studio, the default message displayed in your browser window is cryptic and looks bad. It's just raw text:

`Callback received. Feel free to close this page.`

This overrides the default with a more user friendly confirmation.

Screenshot from 2022-05-16 11-47-14

### DIFF
--- a/src/gui/virtualstudio.cpp
+++ b/src/gui/virtualstudio.cpp
@@ -808,7 +808,14 @@ void VirtualStudio::setupAuthenticator()
             }
         });
 
-        m_authenticator->setReplyHandler(new QOAuthHttpServerReplyHandler(port, this));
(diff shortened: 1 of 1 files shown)


mattahorton:
Love it. I can fix the clang stuff in the morning. 

We should be able to get Poppins and scripts in there to do fancy stuff in the next iteration!

psiborg112 (approved):

mikedickey:
@codedoser 's original design included Poppins and black background. Sadly, Qt seems to hard-code everything through and including the body tag so changing those things does not seem to be possible. But someone more CSS savvy than us may be able to figure out a way to do it all from within the body tag? https://code.qt.io/cgit/qt/qtnetworkauth.git/tree/src/oauth/qoauthhttpserverreplyhandler.cpp?h=dev#n124

PR #568, opened 2022-05-16
Adding studio filters to list view

nwang92:
This adds a button + menu that allows you to apply different studio filtering options. 

Screenshots:
Screen Shot 2022-05-16 at 2 29 21 PM
Screen Shot 2022-05-16 at 2 29 35 PM

### DIFF
--- a/src/gui/virtualstudio.cpp
+++ b/src/gui/virtualstudio.cpp
@@ -65,9 +65,11 @@ VirtualStudio::VirtualStudio(bool firstRun, QObject* parent)
 {
     QSettings settings;
     settings.beginGroup(QStringLiteral("VirtualStudio"));
-    m_refreshToken = settings.value(QStringLiteral("RefreshToken"), "").toString();
-    m_userId       = settings.value(QStringLiteral("UserId"), "").toString();
-    m_uiScale      = settings.value(QStringLiteral("UiScale"), 1).toFloat();
+    m_refreshToken   = settings.value(QStringLiteral("RefreshToken"), "").toString();
+    m_userId         = settings.value(QStringLiteral("UserId"), "").toString();
+    m_uiScale        = settings.value(QStringLiteral("UiScale"), 1).toFloat();
+    m_showInactive   = settings.value(QStringLiteral("ShowInactive"), false).toBool();
+    m_showSelfHosted = settings.value(QStringLiteral("ShowSelfHosted"), false).toBool();
     settings.endGroup();
     m_previousUiScale = m_uiScale;
 
@@ -295,9 +297,13 @@ bool VirtualStudio::showInactive()
     return m_showInactive;
 }
 
-void VirtualStudio::setShowInactive(bool val)
+void VirtualStudio::setShowInactive(bool inactive)
 {
-    m_showInactive = val;
+    m_showInactive = inactive;
+    QSettings settings;
+    settings.beginGroup(QStringLiteral("VirtualStudio"));
+    settings.setValue(QStringLiteral("ShowInactive"), m_showInactive);
+    settings.endGroup();
 }
 
 bool VirtualStudio::showSelfHosted()
@@ -305,9 +311,13 @@ bool VirtualStudio::showSelfHosted()
     return m_showSelfHosted;
 }
 
-void VirtualStudio::setShowSelfHosted(bool val)
+void VirtualStudio::setShowSelfHosted(bool selfHosted)
 {
-    m_showSelfHosted = val;
+    m_showSelfHosted = selfHosted;
+    QSettings settings;
+    settings.beginGroup(QStringLiteral("VirtualStudio"));
+    settings.setValue(QStringLiteral("ShowSelfHosted"), m_showSelfHosted);
+    settings.endGroup();
 }
 
 float VirtualStudio::fontScale()
@@ -658,13 +668,11 @@ void VirtualStudio::manageStudio(int studioIndex)
 void VirtualStudio::toggleInactiveFilter()
 {
     setShowInactive(!m_showInactive);
-    getServerList(false);
 }
 
 void VirtualStudio::toggleSelfHostedFilter()
 {
     setShowSelfHosted(!m_showSelfHosted);
-    getServerList(false);
 }
 
 void VirtualStudio::createStudio()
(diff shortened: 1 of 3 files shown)


src/gui/virtualstudio.cpp
```diff
@@ -866,9 +898,17 @@ void VirtualStudio::getServerList(bool firstLoad, int index)
...
+                bool activeStudio  = status == QLatin1String("Ready");
+                bool managedStudio = servers.at(i)[QStringLiteral("managed")].toBool();
+                // Only iterate through servers that we want to show
+                if (!m_showSelfHosted && !managedStudio) {
+                    continue;
+                }
+                if (!m_showInactive && !activeStudio) {
+                    continue;
+                }
+                if (m_showInactive || activeStudio || serverInfo->isManageable()) {
```
mattahorton, line 911: The behavior I expect is `if ( activeStudio || (serverInfo->isManageable() && m_showInactive))`. Extra parentheses for emphasis. I think we only want to show inactive studios you are the admin of? That's how it was before the filters.

mattahorton (changes requested):
I love it visually, but I think "show my inactive studios" has slightly incorrect behavior. Make it only show inactive studios I'm admin of, and I think this is gold.

src/gui/virtualstudio.cpp
nwang92: Done

src/gui/virtualstudio.cpp
```diff
@@ -290,6 +290,26 @@ QString VirtualStudio::connectionState()
...
 }
 
+bool VirtualStudio::showInactive()
+{
+    return m_showInactive;
+}
+
+void VirtualStudio::setShowInactive(bool val)
+{
+    m_showInactive = val;
```
psiborg112, line 300: For consistency with the rest of the code, can we use inactive rather than val here? (Same below.)

src/gui/virtualstudio.cpp
```diff
@@ -866,9 +898,17 @@ void VirtualStudio::getServerList(bool firstLoad, int index)
                 VsServerInfo* serverInfo = new VsServerInfo(this);
                 serverInfo->setIsManageable(
                     servers.at(i)[QStringLiteral("admin")].toBool());
-                QString status = servers.at(i)[QStringLiteral("status")].toString();
-                // Only include active servers, or servers that we can manage.
-                if (status == QLatin1String("Ready") || serverInfo->isManageable()) {
+                QString status     = servers.at(i)[QStringLiteral("status")].toString();
+                bool activeStudio  = status == QLatin1String("Ready");
+                bool managedStudio = servers.at(i)[QStringLiteral("managed")].toBool();
```
psiborg112, line 903: Why create a managedStudio variable here? Why not just use serverInfo->isManageable()?

mikedickey:
> I love it visually, but I think "show my inactive studios" has slightly incorrect behavior. Make it only show inactive studios I'm admin of, and I think this is gold.

What does "my" mean in this context? This may not be clear to users. This filter naturally becomes more complicated by it representing two concepts versus one. What is the intended behavior for inactive studios which are not "mine?"

psiborg112:
A few observations. The state of these check boxes should probably be saved using QSettings under the VirtualStudio group. (I'd suggest just going with ShowInactive and ShowSelfHosted as the key names.) Also, changing the state of these buttons triggers a refresh which means that the action won't be instantaneous, and the user make think that nothing has happened while awaiting the reply from the VS server. To that end we should either show the blocking refresh layer (which gives some visual indication that something is happening which requires the user to wait, and is the easier solution) or we should store multiple server lists - a full list that we can then use to generate filtered lists. (If status isn't already being stored in vsServerInfo, we might need to add that.) Then we can just use the appropriate array to populate the qml model.

psiborg112:
Also, I know it hasn't merged yet, and there are still a couple of things to fix in it, but it would be good if any additional interface changes could be based on the vs-dark-mode branch. (Because it will save a lot of messing around later.)

src/gui/virtualstudio.cpp
nwang92: The meaning is different between these two, it looks like `serverInfo->isManageable()` is based on the `admin` key of the server object whereas this `managedStudio` variable is based on the `managed` key.

The `admin` field is a boolean indicating whether your current user has been given permission to control all the knobs that VS supports. The `managed` field is true if the server is managed by our cloud infra, and false if it's a self-hosted server (see step 2 on https://app.jacktrip.org/studios/create). It's admittedly confusing but `managed: false` would mean the server is self-hosted which is what the toggle intends to control.

nwang92: I think I'll change the variable name to `hostedStudio` to avoid any confusion

psiborg112: I misread that. I thought it was using the same key but just saw that the manageable is taken from the admin setting. Yes changing to hostedStudio would make more sense.

nwang92:
> A few observations. The state of these check boxes should probably be saved using QSettings under the VirtualStudio group. (I'd suggest just going with ShowInactive and ShowSelfHosted as the key names.) Also, changing the state of these buttons triggers a refresh which means that the action won't be instantaneous, and the user may think that nothing has happened while awaiting the reply from the VS server. To that end we should either show the blocking refresh layer (which gives some visual indication that something is happening which requires the user to wait, and is the easier solution) or we should store multiple server lists - a full list that we can then use to generate filtered lists. (If status isn't already being stored in vsServerInfo, we might need to add that.) Then we can just use the appropriate array to populate the qml model.

I opted for the easier approach :) Not sure I fully understand how to use QSettings effectively but I moved some of the get/set logic into updating that object. But let me know if you think it looks incorrect, happy to take another stab at it.

mattahorton:
> > I love it visually, but I think "show my inactive studios" has slightly incorrect behavior. Make it only show inactive studios I'm admin of, and I think this is gold.
> 
> ~What does "my" mean in this context? This may not be clear to users. This filter naturally becomes more complicated by it representing two concepts versus one. What is the intended behavior for inactive studios which are not "mine?"~
> 
> ~The original purpose of this task was to not show _any_ studios which I am unable to join, by default.~
> 
> Nevermind, after some more testing I can see that it never shows inactive studios which I am not an admin of. So although I still think "My" makes the toggle more confusing, I think this meets the original goal.

Glad you understand now.

I agree it's not the best wording, but I put "my" in the quick design to 1. refer to the existing behavior before this ticket which was showing inactive studios I was admin of and 2. to explicitly not show a bunch of public inactive studios like is an option on web.

mikedickey (approved):

PR #569, opened 2022-05-16
VS: Add device setup screen after login

mattahorton:
Screen Shot 2022-05-16 at 5 38 36 PM

This comes up after login. That means it loads every launch if you're using Virtual Studio mode. It's basically the top portion of the settings menu. 

I wanted to do device selection in the Connected screen but found managing the necessary disconnect/reconnect work to be too cumbersome for the time I have at the moment.

### DIFF
--- a/src/gui/Setup.qml
+++ b/src/gui/Setup.qml
@@ -33,7 +33,7 @@ Item {
         model: backendComboModel
         currentIndex: virtualstudio.audioBackend == "JACK" ? 0 : 1
         onActivated: { virtualstudio.audioBackend = currentText }
-        x: 234 * virtualstudio.uiScale; y: 200 * virtualstudio.uiScale
+        x: 234 * virtualstudio.uiScale; y: 150 * virtualstudio.uiScale
         width: parent.width - x - (16 * virtualstudio.uiScale); height: 36 * virtualstudio.uiScale
         visible: virtualstudio.selectableBackend
     }
@@ -49,7 +49,7 @@ Item {
     
     Text {
         id: jackLabel
-        x: leftMargin * virtualstudio.uiScale; y: 200 * virtualstudio.uiScale
+        x: leftMargin * virtualstudio.uiScale; y: 150 * virtualstudio.uiScale
         width: parent.width - x - (16 * virtualstudio.uiScale)
         text: "Using JACK for audio input and output. Use QjackCtl to adjust your sample rate, buffer, and device settings."
         font { family: "Poppins"; pixelSize: fontMedium * virtualstudio.fontScale * virtualstudio.uiScale }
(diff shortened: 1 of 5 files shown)


psiborg112:
Can we add a 'don't ask me again option'? This would drive me nuts having to choose every time.

src/gui/virtualstudio.cpp
```diff
@@ -431,6 +431,10 @@ void VirtualStudio::applySettings()
     m_previousBuffer     = m_bufferSize;
     m_previousInput      = m_inputDevice;
     m_previousOutput     = m_outputDevice;
+
+    // make sure device change signals are emitting
```
psiborg112, line 435: Don't really need the comment here. These just weren't included before because all they do is make sure the interface shows the right devices. When the applySettings function could only be called from one place in the UI there was no need to update the UI since that was driving the change and already in sync.

src/gui/vs.qml
```diff
@@ -41,6 +54,7 @@ Rectangle {
             name: "settings"
             PropertyChanges { target: loginScreen; x: -loginScreen.width }
             PropertyChanges { target: startScreen; x: -startScreen.width }
+            PropertyChanges { target: setupScreen; x: -2*setupScreen.width }
```
psiborg112, line 57: Why the multiplier here?

src/gui/Setup.qml
```diff
@@ -0,0 +1,147 @@
...
+        visible: virtualstudio.audioBackend != "JACK"
+    }
+
+    Button {
+        id: refreshButton
+        background: Rectangle {
+            radius: 6 * virtualstudio.uiScale
+            color: refreshButton.down ? "#DEE0E0" : (refreshButton.hovered ? "#D3D4D4" : "#EAECEC")
+            border.width: 1
+            border.color: refreshButton.down || refreshButton.hovered ? "#BABCBC" : "#34979797"
```
psiborg112, line 102: These shorts of sections should really be taken from the dark mode branch so that they don't need reworking later.

psiborg112:
In addition to my previous comment, this would be particularly annoying for people who habitually use JACK. (Because in that case, there's really nothing worth setting here.)

mikedickey:
> Can we add a 'don't ask me again option'? This would drive me nuts having to choose every time.

+1

mattahorton:
> In addition to my previous comment, this would be particularly annoying for people who habitually use JACK. (Because in that case, there's really nothing worth setting here.)

You got it. Adding that and some warnings today. And I'll grab some of the dark mode branch work if I can.

mattahorton:
Showing the "Ask me next time" check box functioning

https://user-images.githubusercontent.com/5567285/168897099-6dd11bcd-76f7-4390-a472-95c9fa2fb531.mov

mikedickey (approved):

psiborg112:
Noooooo! This merged too soon. The way to fix those fly through issues is not to add arbitrary multipliers to the position, but to ensure that the zorder is correct. The Setup window needs to be created under the login and browse windows.

mattahorton:
@psiborg112 I did put them in the right zorder with some content hiding. I think that multiplier may have been from a previous attempt at fixing it. I've created an issue to remove it.

PR #574, opened 2022-05-19
Add dblsqd for auto-updates

nwang92:
This is a rather large PR but essentially this enables auto-updates via the [dblsqd SDK](https://bitbucket.org/pentacent/dblsqd-sdk-qt/src/master/dblsqd/). More information on dblsqd can be found [here](https://www.dblsqd.com/).

While not perfect (compared to the other frameworks mentioned in the [discussions](https://github.com/jacktrip/jacktrip/discussions/544)) it does facilitate an easy manner of cross-platform upgrades that would be useful now and give us time to research other solutions in the future:
Screen Shot 2022-05-18 at 5 14 06 PM

Here's a list of changes:
* Per [installation guide](https://www.dblsqd.com/docs/integration/qt), added dblsqd SDK directly in `src/dblsqd`. There are some divergences which are: change relative include structure to support both qmake + meson builds, and reformatting all files to comply with our clang rules
* Added `feed` + `updateDialog` in main.cpp for only mac + windows, and only when GUI is enabled
* Set QT application version to enable version comparison

The update information is stored in two different channels (stable + edge), and will ideally support both mac + windows:
```
# stable - empty right now
curl https://files.jacktrip.org/app-releases/stable/mac-manifests.json
curl https://files.jacktrip.org/app-releases/stable/win-manifests.json

# edge - I added one entry for mac just for testing
curl https://files.jacktrip.org/app-releases/edge/mac-manifests.json
curl https://files.jacktrip.org/app-releases/edge/win-manifests.json
```

The contents of these files match the JSON schema defined here: https://www.dblsqd.com/docs/guide/feeds

I've only enabled one "upgrade" on Mac so if you're using that you can change this line and rebuild to test. This could be another GUI setting, just not sure how to do that yet.
```c++
-        QString channel    = "stable";
+        QString channel    = "edge";
```

Some notes:
* These `*-manifests.json` files are in a GCS bucket which only the JackTrip Labs team has write access to
* It would be great to make stable/edge a setting within the app but that could also be a future enhancement
* It would be nice to enable automatic publishing of updates to the edge channel (merge to dev -> update `app-releases/edge/*-manifests.json`) but that's probably a completely separate task. I'm happy to take that on in the future though, I just don't really see it as an immediate blocker for this
* Similar to the above, but for stable as well (merge to main -> update `app-releases/stable/*-manifests.json`)
* In the interim, I'll happily volunteer to handle any changes to stable/edge and put together a user-facing changelog


nwang92:
Looks like I broke meson builds, I'll fix those

nwang92:
Per feedback, I'm going to add a build flag to control when we bundle in dblsqd

nwang92:
I added a `noupdates` build flag which only matters when `nogui` is not used. I wasn't sure if [CMakeLists.txt](https://github.com/jacktrip/jacktrip/blob/main/CMakeLists.txt) is used at all? I tried to update that as well but for the life of me cannot figure out how to test it. It doesn't seem like it has impacted the GHA builds at all though.

I should note that I'm 0% confident in my ability to figure out how these builds work, it's very likely I missed something or did something wrong so no offense taken for anything called out.

mikedickey (approved):

PR #586, opened 2022-05-25
Moving update feed to github

nwang92:


dyfer:
I wanted to point out that we have a [changelog](https://github.com/jacktrip/jacktrip/blob/dev/docs/changelog.yml) file, which should be probably used for the changelog entry here.

Regarding this PR: should creation of these files be automated? (not exactly _with_ the release, but still somehow automated)

dyfer:
> I'm not sure there's an easy way to reference that from the json though - maybe a link would suffice?

I'm guessing that is we used a script that builds the manifests file, it could take the json as the partial source, similarly what's not happening with flatpak packages thanks to @ntonnaett 

> Ideally we can find a way to automate this via some Actions hooks but I don't think that's necessary at the moment. I think we're still trying to figure out if putting this here works.

Got it.

My feeling is that _maybe_ a separate repo in the JT org would be best for this. An action there would build the json file, using appropriate sources etc and expose it under some known location (either as a release artifact, or as a repository "blob")

cchafe (approved):

PR #593, opened 2022-05-27
Vs dark mode

psiborg112:


mattahorton:
> It seems this doesn't obey the global setting on linux. It would be great if this was supported.

Agreed. I'm betting there is a global check for all platforms in Qt and that is better than a button in the app, imo.

psiborg112:
Then feel free to change how dark mode is being set.

psiborg112:
(Although I do have to say that for development, being able to manually change it is extremely useful.)

nwang92:
FWIW, the pattern I see everywhere is a "choose one of light, dark, or match system" which still preserves the development aspect. One example:
Screen Shot 2022-05-28 at 9 45 56 AM

nwang92:
I'm seeing an empty screen on mac with these errors:
Screen Shot 2022-05-28 at 9 49 58 AM

nwang92:
> I'm seeing an empty screen on mac with these errors: Screen Shot 2022-05-28 at 9 49 58 AM

I think I fixed it here? https://github.com/jacktrip/jacktrip/pull/602

cchafe:
I've built the branch in qt and meson on fedora35 but both hang after launching GUI with ./jacktrip cmd
///////////////////////////////////////////////////////////////////////////////////////////////////////////////
qrc:/vs/vs.qml:85:5: Type Setup unavailable 
         Setup { 
         ^
qrc:/vs/Setup.qml:391:13: id is not unique 
                 id: backendCombo 
                 ^
SSL Library: OpenSSL 1.1.1n  FIPS 15 Mar 2022

nwang92 (approved):

cchafe:
all good

PR #606, opened 2022-05-30
Store listview y position after refresh

psiborg112:
Make sure we update our saved scroller y value in the browse window when returning from settings.


nwang92:
Should this be `src/gui/Browse.qml`?

psiborg112:
Ugggh, it should. This is what happens when I try to do a quick web upload.

psiborg112:
And that's better. Thanks.

nwang92 (approved):

PR #622, opened 2022-06-13
VS: Register JackTrip as Virtual Studio device

mattahorton:
This is work in progress.

It registers JackTrip as a Virtual Studio device when in the Virtual Studio GUI, and deletes that device from the server on logout.


mattahorton:
Would appreciate a good look at this one. I had to learn a lot to get this one done and probably did something weirdly for Qt.

src/gui/virtualstudio.cpp
```diff
@@ -933,6 +1000,189 @@ void VirtualStudio::setupAuthenticator()
...
+        {QLatin1String("queueBuffer"), 0},
+        {QLatin1String("devicePort"), 4464},
+        {QLatin1String("reverb"), 0},
+        {QLatin1String("limiter"), false},
+        {QLatin1String("compressor"), false},
+        {QLatin1String("quality"), 2},
+        {QLatin1String("captureBoost"), false},
+        {QLatin1String("captureMute"), false},
+        {QLatin1String("captureVolume"), 100},
+        {QLatin1String("playbackBoost"), false},
```
nwang92, line 1063: Can we remove `captureBoost` and `playbackBoost`? They are valid options, but we've recently started ignoring/retiring them

src/gui/virtualstudio.cpp
```diff
@@ -933,6 +1000,189 @@ void VirtualStudio::setupAuthenticator()
...
+        alsaName - "jacktripapp"
+        overlay - "jacktrip_app"
+        mac - UUID tied to app session
+        version - app version - will need to update in heartbeat
+        apiPrefix - random 7 character string tied to app session
+        apiSecret - random 22 character string tied to app session
+    */
+
+    QJsonObject json = {
+        {QLatin1String("period"), 128},
```
nwang92, line 1053: I think this should be the same value as `m_bufferSize`? At least, it's possible for studio buffer size != device buffer size

nwang92:
One thing I did that seems to have messed something up:
* I opened the desktop app + signed in
* I saw the new device register to my account
* I closed the desktop app and the device remained on my account
* I deleted the device from Virtual Studio web
* I re-opened the desktop app and the device didn't re-appear

Not sure what the ideal result should be, but I would expect a device to always show up even if I manually/accidentally deleted it on the Virtual Studio side.

mattahorton:
> One thing I did that seems to have messed something up:
> 
> * I opened the desktop app + signed in
> * I saw the new device register to my account
> * I closed the desktop app and the device remained on my account
> * I deleted the device from Virtual Studio web
> * I re-opened the desktop app and the device didn't re-appear
> 
> Not sure what the ideal result should be, but I would expect a device to always show up even if I manually/accidentally deleted it on the Virtual Studio side.

That is such a hard thing to work around. Like...do we _always_ do an extra check for existence of the device on startup? Or always do a create? What happens at the api level if we do a post on /devices and the device exists already?

Anyway, just a bunch of thoughts on that.

Also:
1. Why would someone delete it on the web?
2. The way to make it work is to log out using button in settings and then log back in.

src/gui/virtualstudio.cpp
mattahorton: good call

nwang92:
I don't know why someone would delete it on web but it's currently possible so it seems like a case we should account for. The exclusion of a valid desktop-app-device also doesn't prevent usage but I imagine it will probably break other things we want to do down the road so it'd be nice if we never let people get into this scenario. 

Some possible solutions:
* Prevent DELETE on `jacktrip_app` devices. Pretty easy to do, one thing to consider is if someone logs into the desktop app on n machines they will always have n devices in their account.
* On startup, run a GET with the current `m_appUUID` to see if it exists. It's one additional call, but it might be safe to assume if we get a 404 that we should re-create a new device
* Always POST /devices on startup. This should be fine provided the ownerID, apiPrefix, and apiHash always match

src/gui/vsWebSocket.cpp
```diff
@@ -0,0 +1,108 @@
...
+}
+
+bool VsWebSocket::isConnected()
+{
+    return m_connected;
+}
+
+bool VsWebSocket::isValid()
+{
+    return !m_error && m_connected;
```
nwang92, line 107: Similarly, `m_error` never gets changed anywhere

src/gui/virtualstudio.cpp
```diff
@@ -933,6 +992,244 @@ void VirtualStudio::setupAuthenticator()
...
+
+    if (m_heartbeatWebSocket->isValid()) {
+        // Send heartbeat via websocket
+        m_heartbeatWebSocket->sendMessage(request.toJson());
+    } else {
+        // Attempt to open socket for next time
+        if (!m_heartbeatWebSocket->isConnected()) {
+            m_heartbeatWebSocket->openSocket();
+        } else {
+            // Recreate websocket as there has been an error
```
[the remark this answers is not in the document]
nwang92: I don't have much to offer in C++ knowledge. If I had to guess, I would think this does not clean up the old vswebsocket but I can't say for certain.

nwang92 (approved):

nwang92 (approved):

mattahorton:
I would be fine merging this if we don't plan on pushing 1.6.2 soon. Otherwise I want to wait for #641.

nwang92:
Since there's a boatload of changes here, I think we should probably cut an RC and do extensive testing with that and see what happens.

dhing1024 (approved):

mattahorton:
> Since there's a boatload of changes here, I think we should probably cut an RC and do extensive testing with that and see what happens.

Very much agree.

PR #624, opened 2022-06-14
Set ToS IP header to use DSCP Expedited Forwarding for audio packets

mikedickey:
I believe this is the latest standard for quality of service in routers. Expedited Forwarding (EF) is recommended for real time audio streams.


nwang92 (approved):

mikedickey:
Sadly, this seems to only work on Linux..

mikedickey:
For Windows, I believe we would need to use qos2 or qWAVE https://docs.microsoft.com/en-us/windows/win32/api/qos2/nf-qos2-qossetflow

Issue #615, opened 2022-06-07
Issue with login to virtual studio account


forsythac:
We are seeing an issue where the app hangs on "Logging in..." on some Linux platforms.

**We also have one reported issue with this behavior on Mac**. User is reporting that no browser tab is opened when the app tries to log in, and the app hangs on "Logging in..."

screen shot from user
![login issue](https://user-images.githubusercontent.com/3578658/172493223-dfa1eb29-2fee-4ef5-846b-bcf809fb1361.jpg)

quick reproduction of issue on VM:
https://user-images.githubusercontent.com/3578658/172493780-dc0596d5-a086-4e6c-a83e-c67ae335665e.mp4

Note that in the video above Russ had to launch firefox to get it to show "Virtual Studio Login Successful!" and that the app did not get the memo about that.

Rjapeck:
To add on to this, I took a much longer video. If I delete the config file created when you first launch the app you are returned to the first status the app is in. However, it just goes back to the same screen that it got hung on the first time

https://user-images.githubusercontent.com/105318997/172500044-2ddb22c6-6f0c-48ce-9358-ad1de00b1b61.mp4

.

mattahorton:
I repro'd myself on Ubuntu. Not on Mac though.

psiborg112:
Can't reproduce here on any platform, so I'm not going to be much help debugging this. Is there any terminal output associated with is? Is the lambda function even being called? (Can someone try adding some debugging cout lines to it?)

psiborg112:
Which linux version is being used for this? (Built locally? Binary? Flatpack? Is there someone facing these issues who is capable of creating a local build?)

forsythac:
I'll chime in that I was able to get the Mac user sorted by cleaning out the temp files with https://freemacsoft.net/appcleaner/ and installing the latest version (they had been on rc1 before). So, there are currently no known outstanding issues on Mac.

psiborg112:
My suspicion is that there's potentially some system setup that is causing QDesktopServices::openUrl to fail. So when the authorizeWithBrowser signal fires on our m_authenticator object, Qt isn't finding a way to open the URL. When called directly, QDesktopServices::openUrl has a boolean return state. It would be worth manually trying to open a URL (can be any arbitrary URL) at the start of the login function and checking the return value of this function to see if this is the actual issue or whether we need to look elsewhere.

psiborg112:
[...]

PR #626, opened 2022-06-16
GHA: update static Qt version on Linux

dyfer:
It seems that there is a a newer version of 5.12 available, let's use that on Linux.
~~This _might_ help with #615, but it's not only a guess.~~ EDIT: it does not fix that issue.

EDIT 1: Windows and Flatpak build failures are unrelated to this change
EDIT 2: Before merging, please download and run the Linux build. ~~For the record, please check if it by any chance changes the behavior observed in #615~~

### DIFF
--- a/.github/workflows/jacktrip.yml
+++ b/.github/workflows/jacktrip.yml
@@ -39,8 +39,8 @@ jobs:
             weakjack: false
             # vcpkg-triplet: x64-mingw-static # used when installing rtaudio with vcpkg (Windows)
             # qt-arch: 'win64_msvc2019_64' # specify when using shared (official) qt libraries (Windows)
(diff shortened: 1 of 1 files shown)


dyfer:
I was able to run the binary in Ubuntu 20.04.

ntonnaett:
Login doesn't work.

```
./jacktrip
Warning: Ignoring XDG_SESSION_TYPE=wayland on Gnome. Use QT_QPA_PLATFORM=wayland to run on Wayland anyway.
SSL Library: OpenSSL 1.1.1n  FIPS 15 Mar 2022
qt.networkauth.replyhandler: Host requires authentication
```

ntonnaett (approved):

PR #632, opened 2022-06-20
Fix perpetual "Logging In" screen

nwang92:
Deep Google searches led me to these two issues which seem to fix the auth flow behavior on Ubuntu 20.04:
* https://bugreports.qt.io/browse/QTBUG-84866
* https://stackoverflow.com/questions/69585272/c-signal-is-not-received-in-qml-slot

I've also enabled the "Back" button on the "Logging in" screen so you can return to the original start state if it ever does hang again.

The more appropriate fix would be to use QT 5.15.x for Linux (which we're already using for Mac + Windows) but that's currently not an option for reasons beyond my knowledge.


cchafe (approved):

mattahorton:
Thanks Nelson!!!

PR #639, opened 2022-06-24
Feature/exit on leave button

nwang92:
This wires up the "leave" button functionality from the web <-> desktop client. There's also a huge refactoring to move the device-specific code into `vsDevice.cpp` which I'm not sure if it was a good idea to attempt to do.

### DIFF
--- a/src/gui/virtualstudio.cpp
+++ b/src/gui/virtualstudio.cpp
@@ -82,6 +82,7 @@ VirtualStudio::VirtualStudio(bool firstRun, QObject* parent)
     m_appID           = settings.value(QStringLiteral("AppID"), "").toString();
     settings.endGroup();
     m_previousUiScale = m_uiScale;
+    m_device = new VsDevice(m_appID, m_authenticator->token(), m_apiPrefix, m_apiSecret);
 
     // Load our font for our qml interface
     QFontDatabase::addApplicationFont(QStringLiteral(":/vs/Poppins-Regular.ttf"));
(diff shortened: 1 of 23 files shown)


src/gui/virtualstudio.cpp
```diff
@@ -643,56 +635,27 @@ void VirtualStudio::completeConnection()
...
+        if (m_inputDevice == QLatin1String("(default)")) {
+            input = "";
         }
-#endif
-        m_jackTrip->setBufferStrategy(1);
-        m_jackTrip->setBufferQueueLength(-500);
-        m_jackTrip->setPeerAddress(studioInfo->host());
-        m_jackTrip->setPeerPorts(studioInfo->port());
-        m_jackTrip->setPeerHandshakePort(studioInfo->port());
+        std::string output = m_outputDevice.toStdString();
```
mattahorton, line 642: You need to wrap any reference to `m_outputDevice` and `m_inputDevice` in an `#ifdef RT_AUDIO`

mattahorton (changes requested):

src/gui/virtualstudio.cpp
mattahorton: basically all of these from `virtualstudio.h`

`QStringList m_inputDeviceList;
    QStringList m_outputDeviceList;
    QString m_inputDevice;
    QString m_outputDevice;
    quint16 m_bufferSize;
    QString m_previousInput;
    QString m_previousOutput;
    quint16 m_previousBuffer;
    bool m_previousUseRtAudio = false;`

PR #641, opened 2022-06-27
VS: Join Studio via URL scheme

mattahorton:


mattahorton:
We are going to go ahead and ship this feature for macOS and Windows. I will make a new PR for Linux.

mikedickey (approved):

nwang92 (approved):

src/gui/qjacktrip.cpp
```diff
@@ -281,7 +286,7 @@ QJackTrip::QJackTrip(int argc, QWidget* parent)
 
     // One of our arguments will always be --gui, so if that's the only one
     // then we don't need to show the warning message.
-    if ((!gVerboseFlag && m_argc > 2) || m_argc > 3) {
+    if (((!gVerboseFlag && m_argc > 2) || m_argc > 3) && !launchedByDeeplink) {
```
psiborg112, line 289: Stuff like this is why you need to be constantly testing NOVS builds as well. (No need to fix this one, I've already done it, and it'll ultimately land in a PR to clean up main and create better separation between core and vs code.)

ntonnaett: But the URL scheme should be generic. Isn't it?

psiborg112: As far as I'm aware, it's not. But that's not the issue here. It's the reliance on a variable that does not exist when NO_VS is set. (And the same issue happens a number of times in main, which also needs a clean up in general. I'm hoping to encapsulate off some of the new stuff when I get a chance, but I at least now have a version which will build properly - will upload later in the week when I'm not on the road and have time to test properly.)

PR #647, opened 2022-07-12
Exposing buffer strategy option

nwang92:
I don't think we finalized the UX for this, but this PR should contain all the necessary backend work + a basic selector to control buffer strategy. If there's any follow up on visual changes we can do that iteratively?


mattahorton:
Are we still doing this giving the issues with loss concealment?

mikedickey:
@nwang92 @mattahorton I'd like to resurrect this for the next release. Now that the interface has multiple settings panes, I think this would best belong under a new "Advanced" one. Other things that should probably move there are "Update Channel" and possibly "Display Mode."

nwang92:
Updated so it looks like this now:
Screen Shot 2022-08-31 at 10 17 02 AM

mikedickey (approved):

Issue #648, opened 2022-07-13
Windows 10 client gets stuck at "step 5" / datagram available step during connection to server


forsythac:
We're seeing some really unpredictable behavior when connecting to servers from a JT client on windows 10

Client JT version = 1.6.1
Server JT versions tested = 1.4.0 and 1.6.1

**Sometimes connection works (maybe 50% of the time), but sometimes only input is connected, and there is no output (user can be heard in the server but can't hear the other participants).** It seems strangely network dependent- some cloud regions seem to have a better success rate than others. This client basically always fails when connecting to SF or Boston region, but almost always works when connecting to the Portland region. Both are AWS regions.

Here's the output from the client when the problem presents itself
![image](https://user-images.githubusercontent.com/3578658/178815293-418a7cf8-203a-4340-b596-097e44c21e66.png)

Here's the output from the server (server thinks the connection is made)
```
Jul 13 18:50:04 ip-10-2-20-79 jacktrip[1106835]: JackTrip HUB SERVER: Client Connection Received!
Jul 13 18:50:04 ip-10-2-20-79 jacktrip[1106835]: JackTrip HUB SERVER: Client Connect Received from Address : ::ffff:70.33.15.77
Jul 13 18:50:04 ip-10-2-20-79 jacktrip[1106835]: JackTrip HUB SERVER: Reading UDP port from Client...
Jul 13 18:50:04 ip-10-2-20-79 jacktrip[1106835]: JackTrip HUB SERVER: Client UDP Port is = 4464
Jul 13 18:50:04 ip-10-2-20-79 jacktrip[1106835]: JackTrip HUB SERVER: Sending Final UDP Port to Client: 61003
Jul 13 18:50:04 ip-10-2-20-79 jacktrip[1106835]: Writing port: 61003
Jul 13 18:50:04 ip-10-2-20-79 jacktrip[1106835]: JackTrip HUB SERVER: Client TCP Connection Closed!
Jul 13 18:50:04 ip-10-2-20-79 jacktrip[1106835]: JackTrip HUB SERVER: Starting JackTripWorker...
Jul 13 18:50:04 ip-10-2-20-79 jacktrip[1106835]: Setting JACK Process Callback...
Jul 13 18:50:04 ip-10-2-20-79 jacktrip[1106835]: SUCCESS
Jul 13 18:50:04 ip-10-2-20-79 jacktrip[1106835]: ---------------------------------------------------------
[...]

PR #656, opened 2022-07-15
Bugfix/win datagram available

nwang92:
This seems to fix https://github.com/jacktrip/jacktrip/issues/648 and the user confirmed the fix. Apparently this is a windows socket bug, problem + fix described here: https://stackoverflow.com/a/42409198

While debugging where `datagramAvailable()` is used, I found that `getPeerAddressFromFirstPacket()` was an unused function so I took the liberty of removing it.

### DIFF
--- a/src/DataProtocol.h
+++ b/src/DataProtocol.h
@@ -159,9 +159,6 @@ class DataProtocol : public QThread
      */
     virtual void setPeerPort(int port) = 0;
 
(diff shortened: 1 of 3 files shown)


psiborg112:
Even though it's unused, I'd tend to err on the side of leaving it in there (albeit maybe commented out) in case it's useful in future.

psiborg112:
(But otherwise it looks good.)

PR #659, opened 2022-07-18
Enabling join + connected view experience

nwang92:
When clicking that semi-hidden testUrlScheme button, this joins the SF public studio and changes the view.

TODO: Error states (studio not found, bad path params and identifiers, etc.)

### DIFF
--- a/src/gui/virtualstudio.cpp
+++ b/src/gui/virtualstudio.cpp
@@ -441,6 +441,24 @@ void VirtualStudio::setDebugText(QString text)
     emit debugTextChanged();
 }
 
+void VirtualStudio::joinStudio(const QUrl& url)
+{
+    if (url.scheme() != "jacktrip" || url.path().length() <= 1) {
+        return;
+    }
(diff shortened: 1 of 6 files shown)


nwang92:
From very limited testing, this seems to work on mac and windows

src/gui/virtualstudio.cpp
```diff
@@ -441,6 +441,24 @@ void VirtualStudio::setDebugText(QString text)
     emit debugTextChanged();
 }
 
+void VirtualStudio::joinStudio(const QUrl& url)
+{
+    if (url.scheme() != "jacktrip" || url.path().length() <= 1) {
+        return;
+    }
+    QString targetId = url.path().remove(0, 1);
```
nwang92, line 449: For what it's worth, the `path()` function turns
```
jacktrip://join/745646a0-704b-4ba9-895d-e5f6752dad08"
```
into
```
/745646a0-704b-4ba9-895d-e5f6752dad08
```
which is why there's the remove call following it

PR #664, opened 2022-07-21
Feature/jt as vs device with stats

dhing1024:
Adds latency statistics to heartbeat messages sent to `vs-server` as part of the Virtual Studio device emulation. Adds two classes: `VsPinger` and `VsPing`. `VsPinger` opens up a websocket connection to the currently connected studio and measures round-trip latency times, similar to how it is done in [jacktrip-agent](https://github.com/jacktrip/jacktrip-agent/blob/develop/cmd/devicePinger.go). `VsPinger` keeps track of the statistics, and the statistics are included with every heartbeat message.

This PR does not include UI components to display network statistics within the desktop app GUI. However, the statistics are visible to users in the JackTrip web app.


src/gui/vsDevice.cpp
```diff
@@ -37,6 +37,7 @@
 
 #include "vsDevice.h"
 
+#include <QCryptographicHash>
```
nwang92, line 40: I don't believe this is used anymore

src/gui/vsDevice.cpp
```diff
@@ -296,6 +315,25 @@ void VsDevice::reconcileAgentConfig(QJsonDocument newState)
     }
 }
 
+// initPinger intializes the pinger used to generate network latency statistics for
+// Virtual Studio
+VsPinger* VsDevice::startPinger(VsServerInfo* studioInfo)
+{
+    QString id        = studioInfo->id();
+    QString sessionId = studioInfo->sessionId();
```
nwang92, line 323: This `QString sessionId` seems to be unused?

src/gui/vsDevice.cpp
```diff
@@ -296,6 +315,25 @@ void VsDevice::reconcileAgentConfig(QJsonDocument newState)
...
+    QString host      = studioInfo->sessionId();
+    host.append(QString::fromStdString(".jacktrip.cloud"));
+
+    m_pinger = new VsPinger(QString::fromStdString("wss"), host,
+                            QString::fromStdString("/ping"));
+}
+
+// stopPinger stops the Virtual Studio pinger
+void VsDevice::stopPinger()
```
nwang92, line 332: Where is `stopPinger()` called? Ideally we only setup this pinger when connected to a studio and teardown once disconnected

src/gui/vsDevice.cpp
```diff
@@ -296,6 +315,25 @@ void VsDevice::reconcileAgentConfig(QJsonDocument newState)
...
+VsPinger* VsDevice::startPinger(VsServerInfo* studioInfo)
+{
+    QString id        = studioInfo->id();
+    QString sessionId = studioInfo->sessionId();
+    QString host      = studioInfo->sessionId();
+    host.append(QString::fromStdString(".jacktrip.cloud"));
+
+    m_pinger = new VsPinger(QString::fromStdString("wss"), host,
+                            QString::fromStdString("/ping"));
+}
```
nwang92, line 329: Probably noob language syntax question here, but do we need a `return m_pinger` here to match the method signature? I know some language automatically return the last time but I'm not quite sure if that's true in this case

src/gui/vsDevice.cpp
dhing1024: added `stopPinger` right before `stopJackTrip` is called - left it out by mistake.

src/gui/vsDevice.cpp
dhing1024: hmm interesting question. I don't think it threw any compiler warnings or errors and I'm not actually sure why. Maybe because the function signature returns a pointer it fills in nullptr if there's no return statement?

meson.build
```diff
@@ -30,6 +30,8 @@ src = [	'src/DataProtocol.cpp',
 	'src/Regulator.cpp',
 	'src/Settings.cpp',
 	'src/UdpDataProtocol.cpp',
+	'src/vsPinger.cpp',
```
mattahorton, line 33: These includes should only be for !novs builds.

Look around line 101-105. Somewhere in there. Do not include them at this level.

meson.build
```diff
@@ -45,6 +47,8 @@ moc_h = ['src/DataProtocol.h',
 	'src/PacketHeader.h',
 	'src/Settings.h',
 	'src/UdpDataProtocol.h',
+	'src/vsPinger.h',
```
mattahorton, line 50: Same feedback about !novs

mattahorton (changes requested):

src/gui/vsDevice.cpp
```diff
@@ -296,6 +314,26 @@ void VsDevice::reconcileAgentConfig(QJsonDocument newState)
...
+    m_pinger = new VsPinger(QString::fromStdString("wss"), host,
+                            QString::fromStdString("/ping"));
+
+    return m_pinger;
+}
+
+// stopPinger stops the Virtual Studio pinger
+void VsDevice::stopPinger()
+{
+    m_pinger->stop();
```
nwang92, line 334: You might want nullptr checks here in case `m_pinger` has not be initialized?

src/gui/vsDevice.cpp
```diff
@@ -309,6 +347,15 @@ void VsDevice::terminateJackTrip()
 void VsDevice::onTextMessageReceived(const QString& message)
 {
     QJsonDocument newState = QJsonDocument::fromJson(message.toUtf8());
+
+    // We have a heartbeat from which we can read the studio auth token
+    // Use it to set up and start the pinger connection
+    QString token = newState["authToken"].toString();
```
nwang92, line 353: This seems to handle the case of going from empty authToken to non-empty authToken, but there's also the inverse we should tackle as well.

It's possible to go from non-empty authToken to empty authToken when you disconnect from a server. I think there's other code that probably handles stopping m_pinger when you disconnect but I think we'll want some guards around this section to make sure we don't restart it? At the very least, `m_pinger->start()` should fail if there's an empty token supplied

src/gui/vsDevice.cpp
psiborg112: It generally won't throw an error unless you turn additional warnings on, but it is usually undefined behaviour. It probably won't return null, but instead the last thing that was in the relevant register. Which is definitely not a good thing for a pointer.

psiborg112: (What IDE are you using? There are usually settings in most that will allow the syntax analyser to warn you about these sorts of things as you code.)

src/gui/vsDevice.cpp
dhing1024: added some additional validation on the `vsPinger` class. Now the caller cannot set/unset the token while the pinger is active, nor can it start the pinger before the token has been set.

nwang92 (approved):

mattahorton (approved):

src/gui/vsDevice.cpp
dhing1024: I'm using VSCode, though I haven't yet really figured out the C/C++ extensions and settings that would help with these sorts of things

PR #665, opened 2022-07-26
Fixing fail view when opening the app from a link

nwang92:


nwang92:
I think this works for mac now - I also cross checked with Andrew on how he wants the warning/device setup behavior to interact with the join link, and the things in Setup.qml will always take precedence.

This seems to work on Mac, I'll test Windows in like 3 hours.

mattahorton:
Did you get to test Windows @nwang92 ?

nwang92:
> Did you get to test Windows @nwang92 ?

Yeah, sorry forgot to update. It works ok on Windows, the only case that doesn't seem to work is opening a join link when the app is already open.

mattahorton:
@nwang92 that's a known issue but is more related to the 2nd instance issue with deeplinks in general. There's a thing I'm trying to solve with that right now.

PR #674, opened 2022-08-05
VU Meters

dhing1024:
Opening this branch up for feedback on a working prototype of VU meters. The volume measurement system is a Faust-compiled C++ plugin, generated from `faust-src/vumeterdsp.dsp` and compiled with `make headers`. These VU meters range from -70dB to 0dB, an an instance is created for each input or output channel. The VU meter plugin and code is structured similarly to the compressor, limiter, and reverb Faust plugins.

To-Do's:
- Add clip indicator functionality [DONE]
- Figure out how to silence build warnings (I suspect that the other Faust plugins do something like this) [DONE]
- Figure out why build actions don't complete successfully [DONE]
- General cleanup [DONE]


ntonnaett:
I think we can add something like `src/*dsp.h` to .clang-tidy-ignore to silence clang-tidy warnings.

dhing1024:
> I think we can add something like `src/*dsp.h` to .clang-tidy-ignore to silence clang-tidy warnings.

@ntonnaett It was already in there before this PR, but the bot still decided to call it out anyways. Interesting...

ntonnaett:
> > I think we can add something like `src/*dsp.h` to .clang-tidy-ignore to silence clang-tidy warnings.
> 
> @ntonnaett It was already in there before this PR, but the bot still decided to call it out anyways. Interesting...

I opened #678 to add a .clang-tidy-ignore file. I must have forgotten to add this when adding the .clang-format-ignore file.

src/JackTrip.cpp
```diff
@@ -588,9 +588,23 @@ void JackTrip::completeConnection()
     for (auto& i : mProcessPluginsFromNetwork) {
         mAudioInterface->appendProcessPluginFromNetwork(i);
     }
+
+    VuMeter* outputMeterPlugin = new VuMeter(mNumAudioChansOut);
+    mAudioInterface->appendProcessPluginFromNetwork(outputMeterPlugin);
```
mattahorton, line 593: I realize this work is still in progress, but why is this not done along with the other plugins in line 589? Same question for input.

dhing1024: The other plugins are instantiated in `Effects.h` (see [here](https://github.com/jacktrip/jacktrip/blob/85fb914e2418fcfe86d23e2e26bdec92852c29b4/src/Effects.h#L77)), which I think is only used when parsing command-line options without a GUI. Or if there is a GUI, they're instantiated in `qjacktrip.cpp` because they are added based on whether certain boxes are checked (see [here](https://github.com/jacktrip/jacktrip/blob/85fb914e2418fcfe86d23e2e26bdec92852c29b4/src/gui/qjacktrip.cpp#L1333)). I think this plugin is a bit different from the others because it doesn't modify the audio signal (it's not really an "effect") and we don't include it based on whether certain boxes in classic mode are checked. 

Adding the VU plugin at `QJackTrip::appendPlugins` seems a little weird to me, but maybe it makes more sense to you?

mattahorton: I guess. I just would like the separation to exist, I think. So that every JackTrip process does not necessarily have this plugin in the chain that it may or may not be using. Is there a way to instantiate it within the VS side of things? 

Maybe not. You'd know better.

src/JackTrip.cpp
```diff
@@ -589,21 +589,29 @@ void JackTrip::completeConnection()
...
     }
 
-    VuMeter* inputMeterPlugin = new VuMeter(mNumAudioChansIn);
-    mAudioInterface->appendProcessPluginToNetwork(inputMeterPlugin);
-    connect(inputMeterPlugin, &VuMeter::onComputedVolumeMeasurements, this,
-            &JackTrip::receivedInputVolumeMeasurements);
-    mVuMeterValuesIn.resize(mNumAudioChansIn);
+#ifndef NO_GUI
+    if (mJackTripMode == CLIENT || mJackTripMode == CLIENTTOPINGSERVER) {
+        VuMeter* inputMeterPlugin = new VuMeter(mNumAudioChansIn);
```
mattahorton, line 608: Sorry to ask again, but why is this line specifically in JackTrip.cpp? It seems like something the GUI should do. _Or_ a flag should be set by the GUI and Effects.h can handle it.

If you look in qjacktrip.cpp, all the other effects are instantiated there. That's for Classic mode, obviously, but we could do something similar in virtualstudio.cpp and maybe later provide controls and a view for classic mode.

mattahorton: Realizing today that I repeated what you said to me in the other comment. 🤦 

I will go back there

src/JackTrip.cpp
mattahorton: I do think having it there where they're instantiated with the GUI makes the most sense. We'll have to adapt it for VS mode because right now we don't use any plugins in that mode, but in this case, we just don't have any check to include the VU meter like we would for the others.

dhing1024: I agree that it makes sense to move it out of the main JackTrip thread. The instantiation is now in `virtualstudio.cpp`. It took a bit of time because there was an issue related to how many channels are actually being metered (the issue specifically affected mono microphone inputs with RTAudio), but I found a workaround for that.

With this update, we can still implement VU meters with classic mode, though we would need to have separate instantiations that happen in `qjacktrip.cpp`.

dhing1024:
@ntonnaett @dyfer any chance one of you could review and approve if this looks good? This branch should be good to go now

src/VuMeter.cpp
```diff
@@ -0,0 +1,155 @@
...
+
+#include <QVector>
+
+#include "jacktrip_types.h"
+
+//*******************************************************************************
+void VuMeter::init(int samplingRate)
+{
+    ProcessPlugin::init(samplingRate);
+    if (samplingRate != fSamplingFreq) {
```
mikedickey, line 49: I don't believe this is possible, since you are calling `ProcessPlugin::init` immediately before this, and it sets `fSamplingFreq`

src/VuMeter.cpp
```diff
@@ -0,0 +1,155 @@
...
+  NONINFRINGEMENT. IN NO EVENT SHALL THE AUTHORS OR COPYRIGHT
+  HOLDERS BE LIABLE FOR ANY CLAIM, DAMAGES OR OTHER LIABILITY,
+  WHETHER IN AN ACTION OF CONTRACT, TORT OR OTHERWISE, ARISING
+  FROM, OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE OR
+  OTHER DEALINGS IN THE SOFTWARE.
+*/
+//*****************************************************************
+
+/**
+ * \file Limiter.cpp
```
mikedickey, line 33: is it? :)

src/VuMeter.cpp
```diff
@@ -0,0 +1,155 @@
...
+    // require the same function signature for the compute() function and is normally used
+    // for the faust plugin output. However, this plugin is not supposed to modify the
+    // signal itself like the other plugins (e.g. Limiter) do, so we don't want to write
+    // to this buffer. We just need to report the VU meter output
+
+    if (not inited) {
+        std::cerr << "*** VuMeter " << this << ": init never called! Doing it now.\n";
+        if (fSamplingFreq <= 0) {
+            fSamplingFreq = 48000;
+            std::cout << "Limiter " << this
```
mikedickey, line 92: not Limiter

src/VuMeter.cpp
```diff
@@ -0,0 +1,155 @@
...
+                      << ": *** HAD TO GUESS the sampling rate (chose 48000 Hz) ***\n";
+        }
+        init(fSamplingFreq);
+    }
+
+    for (int i = 0; i < mNumChannels; i++) {
+        /* Run the signal through Faust to  */
+        QVector<float> meterBuf(nframes);
```
mikedickey, line 100: given that `nframes` does not change, this (and the next two `float` statements) can be moved outside of the `for` loop.

src/VuMeter.cpp
```diff
@@ -0,0 +1,155 @@
...
+        mNumChannels = nChansOut;
+    }
+
+    mValues.resize(mNumChannels);
+    QVector<float>::iterator it;
+    for (it = mValues.begin(); it != mValues.end(); ++it) {
+        *it = -80.0;
+    }
+
+    return;
```
mikedickey, line 139: unnecessary `return`

src/VuMeter.cpp
```diff
@@ -0,0 +1,155 @@
...
+    if (outgoingPluginToNetwork) {
+        mNumChannels = nChansIn;
+    } else {
+        mNumChannels = nChansOut;
+    }
+
+    mValues.resize(mNumChannels);
+    QVector<float>::iterator it;
+    for (it = mValues.begin(); it != mValues.end(); ++it) {
+        *it = -80.0;
```
mikedickey, line 136: this should probably be a constant

src/VuMeter.cpp
```diff
@@ -0,0 +1,155 @@
...
+        }
+        init(fSamplingFreq);
+    }
+
+    for (int i = 0; i < mNumChannels; i++) {
+        /* Run the signal through Faust to  */
+        QVector<float> meterBuf(nframes);
+        float* meterBufPtr = meterBuf.data();
+        float** output     = &meterBufPtr;
+        vumeterP[i]->compute(nframes, &inputs[i], output);
```
mikedickey, line 103: I'm not sure why the Faust processing is necessary. If we're just taking max (peak), why not just use the `float` samples directly?

mikedickey (changes requested):

dyfer:
> @ntonnaett @dyfer any chance one of you could review and approve if this looks good?

I have very limited experience with this part of the codebase, so I can't offer feedback off the top of my head. I believe others are better suited to review this, sorry I can't help!

src/VuMeter.cpp
dhing1024: whoops!

src/VuMeter.cpp
dhing1024: done

src/VuMeter.cpp
dhing1024: done

src/VuMeter.cpp
dhing1024: I think we should leave this in, but really mostly because the other ProcessPlugin subclasses also do it within their `init()` functions. I don't actually know why they did that, maybe it was just for debugging purposes?

src/VuMeter.cpp
dhing1024: I think it leaves us with some additional flexibility, and I was using it primarily because the original metering was RMS. If we ever wanted to customize how the metering works, we would need to use Faust. It's also nice to have some separation between the audio processing modules and the rest of the codebase. The Faust stuff can be experimented with and tested on separately.

There's also not really a performance cost. Perf test results: with bufsize 1024 and 2 channels, normal C++ would take about 30 microseconds to process the buffers, and Faust C++ takes about 45 microseconds. By comparison, including the freeverb plugin would add about 400 microseconds of compute time.

dhing1024: I agree that it doesn't provide immediate benefits - it just would leave us with more flexible options later down the road if we wanted to exercise those options. But I can take it out if you feel strongly about it.

dhing1024:
Updated branch based on @mikedickey's feedback. Also moved `VuMeter.cpp` and `VuMeter.h` to just `Meter.cpp` and `Meter.h`

nwang92 (approved):

nwang92:
Did some functional testing on Mac and Windows and it seems pretty solid. There's some discrepancy between the volume reported here and the studio card, but I think it's best to tackle that in a separate PR once we figure out the root cause.

mikedickey (approved):

PR #675, opened 2022-08-06
Fixes for NO_VS builds

psiborg112:
Fix build issues when NO_VS is set

Update CMakeLists.txt

Move all vs files into src/gui since they're not used by the command line version of the app


ntonnaett:
We should have a GHA check for the NO_VS build. Doesn't need to be in this PR. Maybe @dyfer wants to help with that?

psiborg112:
That would be useful to have down the line. Would like to clean up main a bit as well and encapsulate some of the VS specific stuff that seems to be happening there, but just wanted to get a working build to start with.

mattahorton:
Good catch. This is not a build issue then? Because those tests pass.

psiborg112:
We don't currently have any NO_VS builds in github actions that still include the rest of the GUI. So it wouldn't have been caught by the regular tests, just by me updating my build.

mattahorton:
> We don't currently have any NO_VS builds in github actions that still include the rest of the GUI. So it wouldn't have been caught by the regular tests, just by me updating my build.

Gotcha. That makes total sense.

src/main.cpp
```diff
@@ -314,9 +314,11 @@ int main(int argc, char* argv[])
         // Check if we need to show our first run window.
         QSettings settings;
         int uiMode = settings.value(QStringLiteral("UiMode"), QJackTrip::UNSET).toInt();
+#ifndef __unix__
```
mattahorton, line 317: is `__unix__` true for macOS?

src/gui/vsDevice.cpp
```diff
@@ -474,13 +472,13 @@ QString VsDevice::randomString(int stringLength)
         "abcdefghijklmnopqrstuvwxyzABCDEFGHIJKLMNOPQRSTUVWXYZ0123456789");
 
     if (!seeded) {
-        qsrand(QTime::currentTime().msec());
+        m_randomizer.seed((QTime::currentTime().msec()));
```
mattahorton, line 475: Thanks for this. I wasn't sure how to update this one.

src/main.cpp
psiborg112: No. But it is true on BSD which is why we're using it here instead of just using __linux__.

mattahorton: Thanks for the clarification!

mattahorton (approved):

dyfer:
A follow up question: should we add no_vs build option to qmake and meson?

src/gui/vsDevice.cpp
psiborg112: No problem. There are more secure ways to do this without needing to seed things, but didn't know if you needed predictable results from the generator.

psiborg112:
Yes, we would need a NO_VS build option on at least one of those. I'm 
happy to put one together for qmake when I get a chance.

ntonnaett:
> Yes, we would need a NO_VS build option on at least one of those. I'm happy to put one together for qmake when I get a chance.

Meson and qmake already have a novs option.

psiborg112:
CMake definitely would, because I did the initial VS development and that's what I use to build. I just forgot that it had been added to the other build systems. (Even though the commit history shows that I'm the one who added it to qmake.)

PR #678, opened 2022-08-09
Add .clang-tidy-ignore file

ntonnaett:
Adds .clang-tidy-ignore file. This should silence clang-tidy GHA for all dsp headers.

### DIFF
--- a/.clang-tidy-ignore
+++ b/.clang-tidy-ignore
@@ -1,0 +1,4 @@
+subprojects
+externals
+documentation
+src/*dsp.h


mattahorton:
I don't understand enough about this part of the codebase, but is that something we want to do?

dhing1024:
> I don't understand enough about this part of the codebase, but is that something we want to do?
> 
> EDIT: I mean, do we specifically want to silence the warnings for those headers?

The `src/*dsp.h` files are the auto-generated header files from Faust. They have all the Faust classes defined within them, and can get quite long, around 2,000 lines. I think it's fine to silence warnings for them.

ntonnaett:
> I don't understand enough about this part of the codebase, but is that something we want to do?
> 
> EDIT: I mean, do we specifically want to silence the warnings for those headers?

Those *dsp.h files are generated by faust. So every regeneration means fixing those clang-tidy warnings. But I don't recall if we decided to do this. It's not that often that we regenerate those files.

dhing1024 (approved):

PR #681, opened 2022-08-09
Use bundled RtAudio on macOS - fix compatibility with macOS 10.13

### LINKED ISSUE #671: JackTrip crashes on High Sierra on first launch
Followup from high sierra issue reported on https://github.com/jacktrip/jacktrip/issues/552 another user has encountered this crash when launching JackTrip GUI

MacOS v 10.13.6 High Sierra

image

this is the output of "Report"

[high_sierra_report.txt](https://github.com/jacktrip/jacktrip/files/9263384/high_sierra_report.txt)

any ideas on this one? I also can't find any firm recommendations about what the oldest supported MacOS version is.

dyfer:
Fixes #671.

This ~~is an attempt to~~ fixes compatibility with macOS 10.13. RtAudio from homebrew is built for macOS 10.14 and later AFAIU so building RtAudio ourselves ~~might fix this~~ is necessary to support an earlier OS.

Changes in this PR:
- JT GHA: main macOS build uses bundled RtAudio
- JTLabs GHA: macOS build uses bundled RtAudio
- qmake file for RtAudio now specifies the same deployment target as JackTrip's qmake file
  - deployment target is set very low - 10.9 - but the current build practically supports only 10.13 and up. This is due to Qt's compatibility - Qt 5.15, which we use on macOS for the release builds, supports 10.13 at the earliest.
- JT GHA: shared macOS build (non-release) uses system RtAudio (for testing build configuration)
- affected actions are renamed to reflect the use of bundled/system RtAudio

### DIFF
--- a/.github/workflows/jacktrip.yml
+++ b/.github/workflows/jacktrip.yml
@@ -101,11 +101,11 @@ jobs:
           #   build-system: meson
           #   meson-library-type: static
 
-          - name: macOS-x64-qmake-clang-static
+          - name: macOS-x64-qmake-clang-static-bundled_rtaudio
             release-name: macOS-x64
             runs-on: macos-11
-            system-rtaudio: true
-            bundled-rtaudio: false
+            system-rtaudio: false
+            bundled-rtaudio: true
             nogui: false
             novs: false
             weakjack: true
--- a/.github/workflows/jacktrip.yml
+++ b/.github/workflows/jacktrip.yml
@@ -117,10 +117,10 @@ jobs:
             installer-path: installer
             build-system: qmake
 
-          - name: macOS-x64-qmake-clang-shared-bundled_rtaudio
+          - name: macOS-x64-qmake-clang-shared
             runs-on: macos-11
-            system-rtaudio: false
-            bundled-rtaudio: true
+            system-rtaudio: true
+            bundled-rtaudio: false
             nogui: false
             novs: false
             weakjack: true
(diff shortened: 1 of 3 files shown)


mattahorton (approved):
The work looks great! I do want someone to test on High Sierra, but it hasn't broken on the two versions of macOS I can test on so I feel comfortable merging it.

ntonnaett:
Shouldn't we set the minimum version to 10.13 then, if we practically only support that?

dyfer:
> Shouldn't we set the minimum version to 10.13 then, if we practically only support that?

One can still build with an earlier version of Qt (at least 5.12, but maybe even 5.9, which would bring the support down to macOS 10.10). IMO it doesn't hurt to leave that low so that those who build themselves don't have to edit the file on older systems.

dyfer:
> I do want someone to test on High Sierra, but it hasn't broken on the two versions of macOS I can test on so I feel comfortable merging it.

Yes, it would be good to have an additional person to test on High Sierra (I don't have access to one currently). FWIW Mike O'Connor reported an earlier build from this PR to work on 10.13.

OConnorStP:
I'm happy to test, but I need to be led by the hand to the installer.  :-)

dyfer:
> I'm happy to test, but I need to be led by the hand to the installer. :-)

Here it is: https://github.com/jacktrip/jacktrip/actions/runs/2833922486

For future, you can find builds for a given PR by going to "Checks" tab on top (see: Conversation | Commits | Checks | Files changed). Then, click on the appropriate Github Action name, in our case `build`. This will lead you to a page, where build artifacts are listed.

OConnorStP:
Ah!  Thanks for the roadmap.  :-)

Works!  Here's the output from the Terminal run:

https://www.dropbox.com/s/aiqqxcg4dlkc17p/jacktrip.1.6.2.-rc.2_OSX10.13_test_2.txt

Issue #660, opened 2022-07-18
Jacktrip crashes upon devices that do not match hub server sample rate


nwang92:
I was able to repro a customer issue in which using bluetooth headsets crash the jacktrip app. The problem isn't bluetooth-related, but rather device compatibility. 

Upon loading the app, rtaudio seems to select both bluetooth mic + headset as my input + output devices respectively. When I join a hub server that's running at 48k sample rate, the app crashes. Using a dev build I was able to find the following:
```
step 1
  JackTrip:startProcess before checkIfPortIsBinded(mReceiverBindPort)
  JackTrip:startProcess before checkIfPortIsBinded(mSenderBindPort)
  JackTrip:startProcess before setupAudio
Setting Up RtAudio Interface
---------------------------------------------------------
Cannot connect to server socket err = No such file or directory
Cannot connect to server request channel
jack server is not running or cannot be started
JackShmReadWritePtr::~JackShmReadWritePtr - Init not done for -1, skipping unlock
JackShmReadWritePtr::~JackShmReadWritePtr - Init not done for -1, skipping unlock
Cannot connect to server socket err = No such file or directory
Cannot connect to server request channel
jack server is not running or cannot be started
JackShmReadWritePtr::~JackShmReadWritePtr - Init not done for -1, skipping unlock
JackShmReadWritePtr::~JackShmReadWritePtr - Init not done for -1, skipping unlock
Selecting default INPUT device
Cannot connect to server socket err = No such file or directory
Cannot connect to server request channel
jack server is not running or cannot be started
JackShmReadWritePtr::~JackShmReadWritePtr - Init not done for -1, skipping unlock
JackShmReadWritePtr::~JackShmReadWritePtr - Init not done for -1, skipping unlock
Selecting default OUTPUT device
INPUT DEVICE:
Cannot connect to server socket err = No such file or directory
Cannot connect to server request channel
jack server is not running or cannot be started
JackShmReadWritePtr::~JackShmReadWritePtr - Init not done for -1, skipping unlock
JackShmReadWritePtr::~JackShmReadWritePtr - Init not done for -1, skipping unlock
Audio Device  [1] : Apple Inc.: BOSE
  Output Channels : 0
  Input Channels  : 1
  Supported Sampling Rates: 16000
  --Default Input Device--
  --Probed Successful--
---------------------------------------------------------
OUTPUT DEVICE:
Cannot connect to server socket err = No such file or directory
Cannot connect to server request channel
[...]

PR #686, opened 2022-08-12
Don't crash on unsupported RtAudio sample rate

mattahorton:
What this change _actually_ does is pass RtAudio errors up to Qt for JackTrip to handle. If a GUI is present, the GUI handles it gracefully. 

In classic mode, it just pops up a message box and refuses to connect.

In Virtual Studio mode, it pops up a message box and goes back to the browse screen.

### DIFF
--- a/src/JackTrip.cpp
+++ b/src/JackTrip.cpp
@@ -1050,14 +1050,14 @@ void JackTrip::stop(const QString& errorMessage)
     mHasShutdown = true;
     std::cout << "Stopping JackTrip..." << std::endl;
 
-    if (mDataProtocolSender != nullptr){
+    if (mDataProtocolSender != nullptr) {
         // Stop The Sender
         mDataProtocolSender->stop();
         mDataProtocolSender->wait();
     }
 
-    if (mDataProtocolReceiver != nullptr){
-         // Stop The Receiver
+    if (mDataProtocolReceiver != nullptr) {
+        // Stop The Receiver
         mDataProtocolReceiver->stop();
         mDataProtocolReceiver->wait();
     }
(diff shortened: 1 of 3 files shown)


mattahorton:
This should close #660

nwang92:
Tested with my bluetooth headphones and it seems to work for me

nwang92 (approved):

mattahorton:
Thanks @nwang92. I do want to have at least 1 other person look at this, as this is the first time I'm getting into the lower-level parts of the code.

src/JackTrip.cpp
```diff
@@ -1050,13 +1050,17 @@ void JackTrip::stop(const QString& errorMessage)
     mHasShutdown = true;
     std::cout << "Stopping JackTrip..." << std::endl;
 
-    // Stop The Sender
-    mDataProtocolSender->stop();
-    mDataProtocolSender->wait();
+    if (mDataProtocolSender != nullptr) {
```
dhing1024, line 1053: Is there a specific reason you're removing these checks?

dhing1024 (commented):
I don't think I'm able to test this out for myself, but it looks good I think apart from the question I have about the null pointer check.

dyfer (approved):
I tested this on macOS 12 and Windows 10 (ASIO) by attempting to use 32kHz sampling rate (unsupported by my audio hardware on both platforms). This works as expected, displaying an error message instead of crashing JT.

src/JackTrip.cpp
mattahorton: I'm not. I moved them under the protection of that if. If those dataprotocol objects are null (like when we fail out of setting up audio for sample rate reasons), we get a segfault trying to shut down the jacktrip process.

PR #689, opened 2022-08-13
UI: fix fractional UI scaling on Windows

### LINKED ISSUE #688: HiDPI scaling on Windows
It seems that JT 1.6.x rounds up scaling on Windows. I.e. when screen is set to 150% scaling, JT renders at 200%, in my case making it too large for the display ("Connect" button is below the screen bounds).

I think I know where to look for the solution, I'll try to submit a PR later.

dyfer:
Fixes #688

This applies primarily to Windows, when the display scaling is set to 150%, 125%, 175% etc. Previously, the interface was rendered at 100% or 200% (most noticeable for 150%, which was rendered at 200%). With this change, the interface is scaled accordingly to to the fractional scaling.

This might also apply to Linux, although not for our static build, which uses Qt 5.12 (the fix is only valid for Qt 5.14+).
macOS was already behaving correctly.

Reference: https://doc.qt.io/qt-5/highdpi.html#high-dpi-support-in-qt

### DIFF
--- a/src/main.cpp
+++ b/src/main.cpp
@@ -140,10 +140,20 @@ QCoreApplication* createApplication(int& argc, char* argv[])
 #if defined(Q_OS_MACOS) && !defined(NO_VS)
         // Turn on high DPI support.
         JTApplication::setAttribute(Qt::AA_EnableHighDpiScaling);
+        // Fix for display scaling like 125% or 150% on Windows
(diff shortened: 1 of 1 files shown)


mattahorton:
Tested on Windows 11 and works great!

mattahorton (approved):

PR #698, opened 2022-08-17
Fix cache for real

dyfer:
Okay, this time for real... (?) 
It turns out, that my last fix for cache [did not fully work](https://github.com/jacktrip/jacktrip/runs/7887245774?check_suite_focus=true#step:50:2).

cache action@v1 [supports](https://github.com/jacktrip/jacktrip/runs/7866917357?check_suite_focus=true#step:50:3) relative paths (which we used for the static qt build), but does not work on scheduled jobs (that's why I updated it for 1.6.2).
cache action@v2/v3 works on scheduled jobs, but does not support relative paths... which this PR fixes.

This PR changes the location of the static qt build, which fixes caching with the new version of the action. IUUC, cache should also now work for the scheduled jobs (but we'll get the confirmation once the scheduled job gets triggered...)

The caching of the static build [did work](https://github.com/jacktrip/jacktrip/runs/7890961982?check_suite_focus=true#step:10:20) for the second build in this PR (i.e. at the very least caching is confirmed to work on the same branch).

Additionally, I've noticed that the "install qt" action (for shared qt library) now provides built-in caching, so I've switched to use that.

EDIT: I've also added caching for vcpkg on windows, which should save quite a bit of time. Compare vcpkg in builds [2603](https://github.com/jacktrip/jacktrip/runs/7891098371?check_suite_focus=true#step:8:276) and [2604](https://github.com/jacktrip/jacktrip/runs/7892034279?check_suite_focus=true#step:8:78).

Finally, I've made analogous changes to the JTLabs workflow, but I didn't (couldn't) test that.

---
Since we need this change on `main` in order to be applied to scheduled jobs, I think we should pull this into `main`, then pull `main` into `dev`. Is that okay?

### DIFF
--- a/.github/workflows/jacktrip.yml
+++ b/.github/workflows/jacktrip.yml
@@ -234,10 +234,18 @@ jobs:
           fi
           echo "::set-output name=version::$VERSION"
           echo "::set-output name=name::$NAME"
+          echo "::set-output name=stamp::$(date '+%Y-%m-%d')" # set timestamp for cache
       - name: setup python
         uses: actions/setup-python@v2
         with:
           python-version: '3.x'
+      - name: cache vcpkg on Windows
+        if: runner.os == 'Windows' && matrix.vcpkg-triplet
+        uses: actions/cache@v3
+        with:
+          path: ~/AppData/Local/vcpkg/archives
+          key: ${{ matrix.runs-on }}-${{ matrix.vcpkg-triplet }}-${{ matrix.system-rtaudio }}-${{ steps.set-version.outputs.stamp }}
+          restore-keys: ${{ matrix.runs-on }}-${{ matrix.vcpkg-triplet }}-${{ matrix.system-rtaudio }}
       - name: install dependencies for Linux
         if: runner.os == 'Linux'
         run: |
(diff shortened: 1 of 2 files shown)


mattahorton:
> I've also added caching for vcpkg on windows, which should save quite a bit of time

You're a dream. I was looking into this myself and didn't know where to start. 😭

mattahorton:
> Finally, I've made analogous changes to the JTLabs workflow, but I didn't (couldn't) test that.

My only request is that we copy the vcpkg caching to the JTL workflow before merging. Going to main and then dev is fine by me. Idk about everyone else.

dyfer:
> My only request is that we copy the vcpkg caching to the JTL workflow before merging.

Done

mattahorton (approved):

dyfer:
Yay, it looks that the cache is (finally) working correctly - the scheduled build [did retrieve cache](https://github.com/jacktrip/jacktrip/runs/7935186457?check_suite_focus=true#step:11:20), so did the [release build](https://github.com/jacktrip/jacktrip/runs/7983646287?check_suite_focus=true#step:11:20).

PR #699, opened 2022-08-18
Bump global version to 1.6.3

mattahorton:


dyfer:
I can't remember whether there were previous discussions on this - but since this is pre- 1.6.3-rc, would we consider setting the name to 1.6.3-dev or something like that? That way it would be easier to differentiate dev builds from release builds.

I don't think we've done this before (can't remember) so maybe there should be a larger discussion on this.

mattahorton:
> I can't remember whether there were previous discussions on this - but since this is pre- 1.6.3-rc, would we consider setting the name to 1.6.3-dev or something like that? That way it would be easier to differentiate dev builds from release builds.
> 
> I don't think we've done this before (can't remember) so maybe there should be a larger discussion on this.

I thought the same thing when i was creating the PR. Happy to make it `-dev`

nwang92 (approved):

Issue #616, opened 2022-06-09
Unplugging input/output device while connected causes app to hang


nwang92:
To reproduce:
* Open the app in Virtual Studio mode
* Use some non-default device as either your input or output device
* Connect to any studio
* Unplug the physical device from your computer
* Wait ~1m and the app will hang

Once the app is in this state, you lose the ability leave the studio or close the app (basically no interaction whatsoever). I have to force quit or kill the process manually.

cchafe:
RtAudio or Jack (or both)?

cchafe:
Fails the same with Fedora 35, RtAudio, meson build, USB audio interface
...after unplugging the interface, continuous messages to terminal...
RtApiAlsa::callbackEvent: audio read error, No such device.
RtApiAlsa::callbackEvent: audio read error, No such device.
RtApiAlsa::callbackEvent: audio read error, No such device.

Closing the window doesn't close it but the OS responds with a force quit of 
unresponsive application dialog which is enough to kill it.

cchafe:
Same hang with Jack (except no msgs printing).

JanSto1995:
Thx for posting this here, Nelson. It's be great if that audio read error could be monitored by the program and, once occurring:
- Show that error/pop-up message (along the lines of "It seems like your selected audio device is no longer connected. Please choose a new device!") and/or
- automatically open the setup window again (and automatically refresh devices)

JanSto1995:
After some more testing, here are my findings (on Mac running Catalina 10.15.7 with a Steinberg824 audio interface):
- App crashes as described above when selecting my audio interface for both Input and Output and disconnecting it
- App also crashes when selecting another device (internal MacBook mic) as Input, and my audio interface as Output
- App does NOT crash when selecting my audio interface only as Input, and other device (internal MacBook speakers) as Output

It thus seems like the app crashes when the Output device is disconnected and continues to run when only the input device is changed.. hope this helps

ntonnaett:
I think we didn't implement any callbacks for this. In RtAudio there is [RtAudioErrorCallback](http://www.music.mcgill.ca/~gary/rtaudio/RtAudio_8h.html#aee31aa6f8714fc17b8f99e8ac2ca07fa). For JACK it's probably [JackInfoShutdownCallback](https://jackaudio.org/api/group__ClientCallbacks.html#gaf54d9f133267170ef1e6ce5219d24dd4).

psiborg112:
[...]

PR #701, opened 2022-08-19
Fix device disconnect crashes

mattahorton:
This should address part of #616 

It, in my testing, completely fixes JackTrip crashing when the device is unplugged for those using JACK. It should just throw an error to be handled by JackTrip.

Fixing it for RtAudio is not included, as RtAudio does not provide a way, currently, to detect when a device with an open audio stream is disconnected. This changes in RtAudio 6, but that is not release. We will have to wait for that, unfortunately.

I'm going ahead and opening this PR, but we should leave #616 open until RtAudio 6 is released and we can make the necessary updates to our codebase.

**Testing**

I tested this using VS and Classic mode both with RtAudio and JACK. Can someone do the same with the command line and potentially propose any peer to peer or hub server tests we should run?


dyfer:
On Windows (and macOS) we use bundled_rtaudio, i.e. we build it from source, so the actual upgrade should be done to the rtaudio git submodule. 
I think the proposed change to vcpkg should still be included for consistency, as it is right now.

src/JackAudioInterface.cpp
```diff
@@ -277,6 +279,14 @@ void JackAudioInterface::jackShutdown(void*)
...
 }
 
+//*******************************************************************************
+void JackAudioInterface::jackInfoShutdown(jack_status_t code, const char* reason,
+                                          void* arg)
+{
+    std::cout << reason << std::endl;
+    JackAudioInterface::jackShutdown(arg);
+}
+
```
psiborg112, line 289: If using the jack_on_info_shutdown() callback fixes the issue then we should replace the old jack_on_shutdown() callback rather than adding a second callback. Per the JACK API documentation:

NOTE: if a client calls this AND jack_on_shutdown, then in the event of a client thread shutdown, the callback passed to this function will be called, and the one passed to jack_on_shutdown() will not.

psiborg112:
JACK stuff looks good. Afraid I don't know enough about RtAudio to comment on any changes there. Happy with this going in if someone with more knowledge on that front can check that. It's weird that jack_on_shutdown doesn't get called when the device is unplugged. (Maybe this is a JACK bug?) Given that QjackCtl didn't react either when I unplugged my USB device on linux, maybe it's also relying on that callback? (I should delve into the code when I get a chance.)

nwang92 (approved):

PR #703, opened 2022-08-21
Don't print every argument

psiborg112:
Not sure why this was added, but it definitely shouldn't be in a release version.


psiborg112:
Superseded by #704

PR #704, opened 2022-08-21
Fix qDebug segfault

psiborg112:
Don't install the custom message handler when we haven't initialised the text stream.


mattahorton (approved):
The logging was something I accidentally left in and didn't notice.

Thanks for fixing the message handler. You're right we should release the fix asap.

PR #712, opened 2022-08-25
Bugfix: Join Button from VS-Web Skips Setup

dhing1024:
This should fix the bug causing Virtual Studio to immediately go to the Connected screen upon clicking a Join link from Virtual Studio web app.

The cause is that `joinStudio()` was being called too early, immediately after the call to `setStudioToJoin()`. This PR resolves the bug by removing the signal. 
 
Simulating a deep link from the command line (additionally also works for first time users with sign in):
https://user-images.githubusercontent.com/26987971/186782714-cfe78cc4-1bf4-4e2a-9d1f-ef47efb68c96.mov

### DIFF
--- a/src/gui/virtualstudio.cpp
+++ b/src/gui/virtualstudio.cpp
@@ -1121,150 +1121,154 @@ void VirtualStudio::getServerList(bool firstLoad, bool signalRefresh, int index)
 
     QNetworkReply* reply =
         m_authenticator->get(QStringLiteral("https://app.jacktrip.org/api/servers"));
-    connect(reply, &QNetworkReply::finished, this, [&, reply, topServerId, firstLoad, signalRefresh]() {
-        if (reply->error() != QNetworkReply::NoError) {
-            std::cout << "Error: " << reply->errorString().toStdString() << std::endl;
-            emit authFailed();
-            reply->deleteLater();
-            return;
-        }
+    connect(
+        reply, &QNetworkReply::finished, this,
+        [&, reply, topServerId, firstLoad, signalRefresh]() {
+            if (reply->error() != QNetworkReply::NoError) {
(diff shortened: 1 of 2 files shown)


src/gui/virtualstudio.cpp
```diff
@@ -1124,149 +1121,154 @@ void VirtualStudio::getServerList(bool firstLoad, int index)
...
-        if (firstLoad) {
-            emit authSucceeded();
-            emit refreshFinished(index);
-            m_refreshTimer.setInterval(10000);
-            m_refreshTimer.start();
-            m_heartbeatTimer.setInterval(5000);
-            m_heartbeatTimer.start();
-        } else {
-            emit refreshFinished(index);
-        }
```
nwang92, line 1264: I believe the video and that this fixes the bug, but one think I see is that both branches of this conditional call `emit refreshFinished(index)` (lines 1257 and 1263). 

Would this mean we'd always want `signalRefresh` to be true? Especially in the new `refreshStudios()` implementation?
```
void VirtualStudio::refreshStudios(int index)
{
    getServerList(false, false, index);
}
```

dhing1024: Just did some extra testing here. If we set `signalRefresh` to true here, that would introduce a new bug, where waiting on the "Select Audio Devices" screen would connect on it's own if you were to wait a while.

https://user-images.githubusercontent.com/26987971/186789950-70ee2a5d-b6cb-448b-bc8d-63195a715057.mov

nwang92 (approved):

src/gui/virtualstudio.cpp
mattahorton: So this has broken the refresh list button. `signalRefresh` needs to be true in that instance. I'm working on a fix that includes the knowledge from this discussion.

PR #713, opened 2022-08-26
PLC avoid processing same packet -- zero instead when extreme underrun 

cchafe:
if a client stalls out, a worker using PLC was uselessly processing the last good packet and driving up load


mikedickey (approved):

cchafe:
I may have one more tweak -- converting to draft

PR #714, opened 2022-08-26
PLC distinguish stuck clients

cchafe:
'good underruns' = those under a 30ms threshold (same as used for the UPD waiting too long msg)
'bad underruns' = stuck clients beyond that threshold

the former are likely jitter and should get processed, 
the latter should'nt due to the high DSP load incurred (output zeroes instead)

### DIFF
--- a/src/Regulator.cpp
+++ b/src/Regulator.cpp
@@ -382,11 +382,19 @@ void Regulator::pullPacket(int8_t* buf)
                 goto PACKETOK;
             }
         }
-        if (mLastSeqNumOut == mLastSeqNumIn) {
-            //            std::cout << "mLastSeqNumIn: " << mLastSeqNumIn <<
-            //            "\tmLastSeqNumOut: " << mLastSeqNumOut << std::endl;
+        // make this a global value? -- same threshold as
+        // UdpDataProtocol::printUdpWaitedTooLong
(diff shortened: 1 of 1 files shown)


mikedickey:
How does PLC distinguish a packet that is delayed by > 30ms versus one that never arrives (dropped)? I've encountered a lot of connections that have consistent drop rates (bad home routers?) and am hoping that PLC will be able to handle these.

I'm also a bit confused by how it would know that it's been > 30ms. If that much time has passed, wouldn't it have already triggered PLC and moved on by then?

cchafe:
Good question and good reason for this draft of a theory of operation.
Chris
//////////////////////////////////////
[regulatorTheoryOfOperation.txt](https://github.com/jacktrip/jacktrip/files/9473614/regulatorTheoryOfOperation.txt)

mikedickey:
> Good question and good reason for this draft of a theory of operation.
> Chris
> //////////////////////////////////////
> [regulatorTheoryOfOperation.txt](https://github.com/jacktrip/jacktrip/files/9473614/regulatorTheoryOfOperation.txt)

Thanks, I feel like I understand it about 40% better. :) Look forward to reading the full paper!

mikedickey (approved):

PR #715, opened 2022-08-30
Fix Windows connect hang

mattahorton:
Fixed by introducing a "delay" that processes UI events for the duration of the animation, and then connects to the server. The hang is caused because RtAudio::openStream (via VsDevice::startJacktrip and then JackTrip::startProcess and on down) is blocking the UI.

The "real" fix would to be getting RtAudio opened on a separate thread, but I didn't think that approach was 1. an appropriate level of work or 2. right for how much of the codebase this bug affects. It doesn't make sense to change the structure JackTrip.cpp or RtAudioInterface.cpp because of this level of UI bug.

### DIFF
--- a/src/gui/virtualstudio.cpp
+++ b/src/gui/virtualstudio.cpp
@@ -608,9 +608,9 @@ void VirtualStudio::logout()
     emit hasRefreshTokenChanged();
 }
 
-void VirtualStudio::refreshStudios(int index)
+void VirtualStudio::refreshStudios(int index, bool signalRefresh)
 {
-    getServerList(false, false, index);
+    getServerList(false, signalRefresh, index);
 }
 
 void VirtualStudio::refreshDevices()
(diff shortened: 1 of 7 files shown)


src/gui/virtualstudio.cpp
```diff
@@ -756,7 +756,7 @@ void VirtualStudio::completeConnection()
     }
 
     m_jackTripRunning = true;
-    m_connectionState = QStringLiteral("Connecting...");
+    m_connectionState = QStringLiteral("Preparing audio...");
```
nwang92, line 759: If we want to preserve the functionality of `sendHeartbeat()` not being invoked after this line, we might need to change [this check using m_connectionState](https://github.com/jacktrip/jacktrip/blob/main/src/gui/virtualstudio.cpp#L1103-L1105) to
```
if (m_device != nullptr && m_connectionState != "Connecting..." && m_connectionState != "Preparing audio...") {
```

nwang92 (approved):

PR #716, opened 2022-08-31
docs: update Fedora dependencies for Meson build

ntonnaett:


mattahorton (approved):
Is the jinja requirement something meson-specific?

ntonnaett:
> Is the jinja requirement something meson-specific?

Jinja is used by `add_changelog_to_metainfo.py`. Also yaml… I should add this.

PR #722, opened 2022-09-01
Sign bundled_rtaudio build

mattahorton:
We didn't fully update `jacktriplabs.yml` when we bundled rtaudio with the macOS release to fix High Sierra.


dyfer (approved):
Sorry I missed this, thanks!

PR #727, opened 2022-09-06
VS: Make UI usable when the studio list is empty

mattahorton:
**When all available Studios are filtered out**
CleanShot 2022-09-06 at 18 07 30

**When there are no Studios at all**
CleanShot 2022-09-08 at 11 30 39@2x

### DIFF
--- a/src/gui/virtualstudio.cpp
+++ b/src/gui/virtualstudio.cpp
@@ -420,6 +420,17 @@ void VirtualStudio::setShowSelfHosted(bool selfHosted)
     settings.endGroup();
 }
 
+bool VirtualStudio::showCreateStudio()
+{
+    return m_showCreateStudio;
+}
+
+void VirtualStudio::setShowCreateStudio(bool createStudio)
+{
+    m_showCreateStudio = createStudio;
+    emit showCreateStudioChanged();
+}
+
 bool VirtualStudio::showDeviceSetup()
 {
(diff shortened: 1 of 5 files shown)


src/gui/virtualstudio.cpp
```diff
@@ -1363,6 +1377,15 @@ void VirtualStudio::getServerList(bool firstLoad, bool signalRefresh, int index)
             if (yourServers.isEmpty()) {
                 if (subServers.isEmpty()) {
                     m_logoSection = QStringLiteral("Public Studios");
+
+                    if (pubServers.isEmpty() && skippedStudios == 0) {
+                        // This is a new user
+                        setShowCreateStudio(true);
```
dhing1024, line 1383: If user A just happens to have one of their public studios running and user B is a new user logging in for the first time, then user B won't see the new user screen. Do I have that right?

Just want to make sure I understand that my understanding is correct. (Also recognizing that this is largely intended to be serve as a patch so that we can shut down JackTrip Public Studios if a more robust solution isn't achievable right away)

mattahorton: Yes. this is catching an edge case, not creating a FTUX.

dhing1024: cool, thanks for the clarification

dhing1024 (approved):

PR #741, opened 2022-09-26
VS: Digital level testing and adjustment

mattahorton:
This is ready. It adds simple multiplier-based volume levels and input muting to VS mode in both the setup screen and the connected screen.

This is done through a new plugin called Volume that has a "mute" and a "volume" parameter. "Mute" is a boolean and "volume" is a float from 0.0 to 1.0. 

There is a lot of work here regarding access to the audio interface _without_ starting the JackTrip process. I believe I was careful to not do anything that would break existing functionality, but do let me know if we need to find another way to do some of these things. As it was, the AudioInterface classes were very tightly coupled with the JackTrip class, and we needed a way to get audio information in VS mode without starting the JackTrip process. That's the goal there.

We'll definitely want to smoke test on Windows and Linux. I'll be trying on Windows when the tests finish up.

| Setup | Connected |
| ------ | ----------- |
| CleanShot 2022-09-29 at 17 11 36@2x | CleanShot 2022-09-29 at 16 49 00@2x |


mattahorton:
One strange thing you might notice is that the faust plugin headers all got updated. When I generated the header for Volume in the latest Faust, it didn't compile alongside the existing headers. When I moved back to the last version, it just didn't work. I updated them all to the latest version and addressed some clang issues _by hand_. 

I messed around with the clang files in the project but couldn't get my local clang linter to automatically fix the dsp.h files for me. It was all "unused parameter" errors.

Anyway. It shouldn't cause an issue. If you think it will, please test the plugins in classic mode.

ntonnaett:
The .clang-tidy-ignore file doesn't work for our CI setup. We need another ignore solution for *dsp.h. Sorry for this unpleasant work.

Because of ODR we have to regenerate all dsp files most of the time when we (re)generate one of them with a new faust version. We should put the common parts from the generated files in their own headers. But somebody has to look into the faust generation process for that.

src/gui/virtualstudio.h
```diff
@@ -297,12 +319,22 @@ class VirtualStudio : public QObject
...
+    Meter* m_inputTestMeter;
+    Volume* m_inputVolumePlugin;
+    Volume* m_outputVolumePlugin;
     QTimer m_inputClipTimer;
     QTimer m_outputClipTimer;
 
     float m_meterMax = 0.0;
     float m_meterMin = -64.0;
 
+    float m_inMultiplier  = 0.7;
```
dyfer, line 331: is there a reason this is not `1.0`?

dyfer (commented):
Is the scaler linear or exponential? (sorry, I can't test right now and I can't find it in the code)

Also, was there a consideration to allow amplifying the signal beyond `1.0` (0dB)? I'm not sure if this is a good idea, but might be helpful in some cases. On the other hand, it can easily cause clipping, so maybe not?

src/gui/virtualstudio.h
mattahorton: Not any real reason other than I was testing if it was working when I changed it. I'll change it back

mattahorton:
> Is the scaler linear or exponential? (sorry, I can't test right now and I can't find it in the code)
> 
> Also, was there a consideration to allow amplifying the signal beyond `1.0` (0dB)? I'm not sure if this is a good idea, but might be helpful in some cases. On the other hand, it can easily cause clipping, so maybe not?

It's definitely linear. The focus was on getting the function working, but I think your implication (that we should be using a logarithmic volume) is correct. I can work on that.

I didn't add any digital gain for that exact reason: more likely to clip. But if we did logarithmic, we could do something like +3 or +6 dB and have the default be 0.

mattahorton:
I'm seeing an issue on Windows with JACK. It seems to be attempting to start the Jack client at the start for some unknown reason and failing. This might be because JACK isn't configured well on my PC, but it shouldn't start JACK at all if that's not the previous setting.

dhing1024 (commented):
I couldn't get the sliders to work on my Mac, though the mute button worked fine. 

https://user-images.githubusercontent.com/26987971/193917138-483d94d0-a125-407c-96da-c3af0ee8a858.mov

I also think there might be some bugs still related to the number of channels on the audio device setup screen that present when switching between 1-channel and 2-channel devices. If the previously saved device was 1-channel (the Macbook external microphone in this screen recording), then switching to a 2-channel device (my Pyle sound card) wouldn't display both meters. If the previously saved device was 2-channel, then switching to a 1-channel device would cause the second channel to freeze up

https://user-images.githubusercontent.com/26987971/193917194-57b3ae4b-08fd-48d2-aba9-e0f3028208cd.mov

mattahorton:
> I couldn't get the sliders to work on my Mac, though the mute button worked fine.
Strange. I wonder if the plugin is getting deleted...

> I also think there might be some bugs still related to the number of channels 
good call.

dhing1024:
One thing I noticed about the sliders is that sliding them all the way to the left isn't equivalent to muting. I think this might confuse some users. One suggestion that might be worth doing is doing an edge case "soft mute" if the slider just happens to be all the way to the left which effectively mutes the audio but leaves the slider still enabled.

Screen Shot 2022-10-05 at 4 18 22 PM

Apart from that, the volume slider functionality and bug patches look good to me.

mattahorton:
@dhing1024 Yeah. I tried to write that in Faust and really struggled. I can get back to it. I think I know where I was messed up. 

If it doesn't work, I can trigger mute based on the slider input.

mattahorton:
I think this is good to go. Got that 0 cutoff working now.

faust-src/volumedsp.dsp
```diff
@@ -7,6 +7,8 @@ declare description "Volume Control Faust Plugin for JackTrip, based on Faust ex
 
 import("stdfaust.lib");
 mute = checkbox("[1] Mute");
-gain = hslider("[0] Volume", 0, -40, 0, 0.1) : ba.db2linear : si.smoo : _;
-gainVMute = _ * gain, 0 : select2(mute) : _;
-process = _ <: vgroup("Volume Control", _ : gainVMute);
+gain(v) = v : ba.db2linear : si.smoo : _;
+gainVMute(v) = _ * gain(v), 0 : select2(mute) : _;
+zeroCutoff(v) = _ * gain(v), 0 : select2(v == -40) : _;
```
dhing1024, line 12: I think `gain()` is being called in twice here in the chain It so it would still work, though the scaling would be different (i.e. setting the slider 50% has the effect of using 25% volume).

Screen Shot 2022-10-06 at 11 35 08 AM
Screen Shot 2022-10-06 at 11 35 14 AM

I think this line should be:
`zeroCutoff(v) = _ , 0 : select2(v == -40) : _;`

mattahorton: Oh yeah! This explains that thing I asked you about in slack.

dhing1024 (approved):

dyfer (changes requested):
I tested this and it works, thanks!
~~However, the slider feels like it's still a linear attenuator. I really think it should be exponential.~~

EDIT: I see it's not linear, but the scaling seems off (the other way than linear pot, probably).

Also, it would be great to have this also in the "classic" GUI - though that might be for another PR...

dyfer:
What is the middle point on the slider? Is that -20dB? 
I thought that the volume is too low in the middle point, but maybe it's fine. There's more than one way of scaling an audio slider...
FWIW, here's what I can see in macOS Audio MIDI Setup for built-in soundcard slider values:
100% -> 0 dB
75% -> -8.5 dB
50% -> -18.6 dB
25% -> -31.73 dB
The lowest value before it gets muted is around -63.5 dB.

mattahorton:
@dyfer yes it's -20dB. When I was testing with -60 - 0, the -60 - -40 range was not even registering on the meters.

I didn't add any digital gain here either. Both are things we can change if we want.

dyfer:
> When I was testing with -60 - 0, the -60 - -40 range was not even registering on the meters.

Yeah... linear dB range -60 to 0 would put the middle even lower (-30) and that's not good. I think ideally this should be on some kind of a curve, but I don't have math for it handy. 
FWIW - for the actual numbers in Audio MIDI settings, -40 dB is mapped to around 14%.
Another option would be to map 0.5 to 1 to something like -18 dB to 0 dB and map 0 to 0.5 to -60 dB to -18 dB. I agree this can be adjusted later - it's fine as it is.

mattahorton:
> > When I was testing with -60 - 0, the -60 - -40 range was not even registering on the meters.
> 
> Yeah... linear dB range -60 to 0 would put the middle even lower (-30) and that's not good. I think ideally this should be on some kind of a curve, but I don't have math for it handy. FWIW - for the actual numbers in Audio MIDI settings, -40 dB is mapped to around 14%. Another option would be to map 0.5 to 1 to something like -18 dB to 0 dB and map 0 to 0.5 to -60 dB to -18 dB, but the transition between the two ranges might be weird that way.
> 
> I agree this can be adjusted later - it's fine as it is.

I'm feeling a bit confused 😅

Only the inputs to the plugin are linear. The outputs are logarithmic/exponential (same difference, different direction).

As I understand it, Faust's dB2linear does the magic. It turns a dB value into a multiplier value we can work with. The Qt slider goes from 0 to 1. The headless, code-only Faust slider goes from -40 to 0. That -40 to 0 is the thing that matters. 

[I think this makes it clear the best.](https://www.dr-lex.be/info-stuff/volumecontrols.html)

> A volume control should not be based on percentages, because this implies linearity. **A percentage is only acceptable if it maps to a range of dB values, for instance 0% = -60 dB and 100% = 0 dB**.

^ that's exactly what we're doing.

dyfer:
Sorry if I wasn't clear. Percentage in my previous posts refer to the visual position of the slider (50% - slider in the middle, 100% slider all the way up).

If you look at the sliders in audio software, they typically go from -60/-70/-80/-90 dB to 0/+6/+12 dB. But for a slider that goes from -90 dB to 0 dB it doesn't make sense to put -45 dB half way through (too low). So the dB range shouldn't be linearly mapped to the visual position. 

However, I think that the linear dB range from -40dB to 0dB is good enough (with -20 dB being in the middle).

nwang92 (approved):

mikedickey (approved):

PR #745, opened 2022-10-05
Feature/adding test mode

nwang92:
Adding a button to flip between tenants:
Screen Shot 2022-10-05 at 4 05 28 PM

This triggers a logout because of how we store device IDs in the app. 

Marking this as a draft to prioritize https://github.com/jacktrip/jacktrip/pull/741 - I can resolve merge conflicts after that goes in


dhing1024:
Is there any particular reason why a logout is needed after switching modes? Something to do with the auth token?

nwang92:
The access token works across tenants actually, but `vsDevice` stores its device.id in both `m_appID` and `settings.value(QStringLiteral("AppID")`. Upon switching tenants, this device.id (+ accompanying apiPrefix, apiSecret, etc.) is no longer valid and needs to be reloaded.

I suppose we could get fancy and add a method in vsDevice to do this, but we already have all the cleanup logic in our logout routine so it's easier to piggyback off that.

mattahorton:
One issue with this is that I currently have to quit the app and reload for it to refresh my user metadata. So if I'm logged in with jacktrip.org, then I log out, log in with non-jacktrip.org, I still see my jt.org profile and can go to test mode.

I'm not sure what's going on there. Profile info sticking around until quit has been an issue for a while.

mattahorton:
Need to update the URL for invite links

nwang92:
> One issue with this is that I currently have to quit the app and reload for it to refresh my user metadata. So if I'm logged in with jacktrip.org, then I log out, log in with non-jacktrip.org, I still see my jt.org profile and can go to test mode.
> 
> I'm not sure what's going on there. Profile info sticking around until quit has been an issue for a while.

I think I fixed that by nulling `m_userMetadata` on logout as follows:
```
m_userMetadata = QJsonObject();
```

mikedickey:
idk why but for me, it takes a REALLY long time (about 15 seconds) for the app to respond to a connect to studio request, triggered by joining from web app, while running in test cluster mode.

nwang92:
Is that occurring when you join with the app not running, or are you joining with the app running in the background? If you're joining with the app running in the background, that's a known bug we have in our backlog.

mikedickey (approved):
@nwang92 you are right the delay in connecting is from another bug that also exists currently in production release. it connects right away for me if the app is closed when I trigger join via web app dialog.

PR #746, opened 2022-10-06
VS: Hook up digital levels to web levels

mattahorton:


nwang92:
I'm getting this authentication error when I run this branch:
Screen Shot 2022-10-17 at 2 30 32 PM

psiborg112:
Maybe do a packet capture on the initial TCP handshake to find out what's being sent? Also how is your hub server configured?

psiborg112:
Basically you're getting an error because your server is returning values it shouldn't be. (But there's nothing in these code changes that jump out as a cause of this because they're far removed from where this is happening.) When the client connects to the hub server via TCP it sends the UDP port number it intends to use. But this was sent by JackTrip as a 32 bit value for a 16 bit port. So I was able to hijack the upper 16 bits to send authentication related messages. (Thus adding the ability to request an SSL connection for credential exchange without breaking backwards compatibility with older clients/servers.) The server uses these same upper 16 bits to send through auth responses. The likely reason for this error is that your server is somehow sending through something in these upper 16 bits that it shouldn't be when returning the server UDP port to the client. And the client is noping out because it's not expecting anything authentication related to be returned. (Branch starting at line 856 in JackTrip.cpp.)

psiborg112:
As an example of what you should see this is the exchange between my computer and the jackloop256.stanford.edu server for a connection that doesn't use authorization. This is the TCP data (minus headers) sent by the client.
```
             70 11 00 00 4d 4c 42 41 61 72 6f 6e 00 00   j.p...MLBAaron..
0050   00 00 00 00 00 00 00 00 00 00 00 00 00 00 00 00   ................
0060   00 00 00 00 00 00 00 00 00 00 00 00 00 00 00 00   ................
0070   00 00 00 00 00 00 00 00 00 00 00 00 00 00 00 00   ................
0080   00 00 00 00 00 00                                 ......
```
The first four bytes are our port in 32 bit little endian format 0x00001170. (Most networking protocols default to big endian, but we never explicitly set it and most jacktrip installs are on little endian architectures, so in the interest of not breaking the existing install base I've fixed it to always be little endian.) This gives us the port number in decimal as 4464. Exactly what we expect with zeros in bytes 3 and 4. The remaining 64 bytes are our remote client name as a null terminated string. (This is currently set so that it gets patched as a loopback client on our jacktrip server.)

The server response is simply:
```
0042   4a ee 00 00                                 ..J...
```
This gives us our port 0x0000ee4a, or 61002. Notice once again that bytes 3 and 4 are all zeros. In your case, there may be something in these two bytes that would be triggering the error you're seeing. Where on earth that could be coming from is the mystery here.

mattahorton:
@nwang92 To be clear, the server ignores the app setting the volumes to 100 on registration?

nwang92:
Yeah but I can fix that

mattahorton:
Alright @nwang92. Since we let users set the volume at app open, I didn't want to override that if the server was different when they connected. So I set the server value to the local value when VsDevice is created (after login) as a way to keep the app and server in sync.

If you change it in the setup interface and then connect later, it's OK that the web UI will take a second to adjust to the new value.

src/gui/vsDevice.cpp
```diff
@@ -51,12 +51,67 @@ VsDevice::VsDevice(QOAuth2AuthorizationCodeFlow* authenticator, bool testMode,
...
 
     // Determine which API host to use
     m_apiHost = PROD_API_HOST;
     if (testMode) {
         m_apiHost = TEST_API_HOST;
     }
+
+    // Set server levels to stored versions
+    QJsonObject json = {
+        {QLatin1String("captureVolume"), m_captureVolume},
```
nwang92, line 72: This area needs some x100 conversion

src/gui/vsDevice.cpp
```diff
@@ -51,12 +51,67 @@ VsDevice::VsDevice(QOAuth2AuthorizationCodeFlow* authenticator, bool testMode,
...
+                // TODO: Fix me
+                // emit authFailed();
+                reply->deleteLater();
+                return;
+            }
+        } else {
+            QByteArray response       = reply->readAll();
+            QJsonDocument deviceState = QJsonDocument::fromJson(response);
+            float deviceCaptureVol =
+                (float)deviceState.object()[QStringLiteral("captureVolume")].toDouble();
```
nwang92, line 96: And /100 here

nwang92 (approved):

mikedickey (approved):

PR #747, opened 2022-10-07
Use sane buffer length for loss concealment in virtual studio mode

mikedickey:
"-q auto" normally sets buffer length to -500, but this isn't a sane value when using loss concealment. "-q auto3" is the recommend value, which is equivalent to -3 or 3ms of headroom.


mikedickey:
See also https://github.com/jacktrip/jacktrip-agent/pull/86

nwang92 (approved):

Issue #725, opened 2022-09-02
Build JackTrip for i686


forsythac:
### Discussed in https://github.com/jacktrip/jacktrip/discussions/721

<div type='discussions-op-text'>

<sup>Originally posted by **forsythac** September  1, 2022</sup>
Does JackTrip support 32bit Windows machines? Do we need different executables to support them?</div>

Seems that we need to clarify whether JackTrip is supported on 32bit OS. This issue was reported from some students trying to use JackTrip with a small Acer laptop running Windows 10 32bit

ntonnaett:
x86 machines that only support 32bit are very old. amd64 was introduced around twenty years ago. Fedora even removed their 32bit compatibility packages. Nevertheless building 32bit versions of JackTrip shouldn't be a problem. I'm just wondering why people haven't switched to a 64bit Windows version yet. What hardware are they using?

You mentioned in the discussion that JackTrip is installed in the 32bit location on Windows. Is this just your assumption because there's x86 in the name?

forsythac:
I'm a Windows user myself and it currently installs in Program Files (x86) by default
<img width="744" alt="image" src="https://user-images.githubusercontent.com/3578658/188764661-9b1b16f3-1248-47cf-856c-6e1658d0847c.png">

forsythac:
Hardware is an Acer laptop being used for schools. details from the user:

Acer Computer
Installed RAM: 2.00 GB
System type: 32-bit operating system, x64 based processor

Windows specifications
Edition: Windows 10 Pro
Version: 21H2
Installed on: 5/10/22
OS build: 19044.1766

ntonnaett:
> I'm a Windows user myself and it currently installs in Program Files (x86) by default <img alt="image" width="744" src="https://user-images.githubusercontent.com/3578658/188764661-9b1b16f3-1248-47cf-856c-6e1658d0847c.png">

x86 doesn't mean it is 32bit. AMD64/x86-64 and IA-32 are both extensions to the original 16bit ISA.

If they have a 64bit processor, they should install a 64bit Windows or Linux.

ntonnaett:
There seems to be no official documentation about install locations, but after a short search I found [this](https://stackoverflow.com/questions/63281973/wix-msi-default-installation-location-is-32bit-instead-of-64bit-program-files). Seems we have to tell WiX that the platform is x64.

dyfer:
[...]

PR #750, opened 2022-10-10
CMake build customisations

psiborg112:
These changes only impact my CMake builds. Incorporating them will remove the need for me to maintain my own custom fork, which means I can spend more of my limited dev time on more useful and important things.


dyfer:
Thanks @psiborg112 
I believe this PR also includes commits from #748 and #749 (?) If that's the case, should they stay here or be removed from this PR?

Also, would you be okay if I added cmake build to our CI or do you still consider it to be "internal use only"? Ideally it would be great to have at least rudimentary instructions for building with CMake, either on the [docs page](https://jacktrip.github.io/jacktrip) or even included in a comment here...

psiborg112:
Given they've just been merged it probably doesn't matter too much. And yeah, happy for you to do that although it might be of limited use. I think it should work on most linux distros, and I'm about to make some changes that should allow it to be a bit more flexible on windows. It would be interesting to see if it works for others on Mac. (I use a custom patched macports install of Qt to allow me to make universal binaries. Might need to make some changes for homebrew.)

ntonnaett:
Would you mind rebasing and fixing the conflicts? I think it's better to merge this ASAP if nothing is missing in your point of view.

ntonnaett (approved):

dyfer:
I wanted to do a full review, testing cmake etc... but if we want to merge it first and then look at it more closely that's okay too, as nobody else is using cmake. I was just thinking I'd test it out and make rudimentary docs along the way, just haven't gotten around to it yet.

ntonnaett:
No pressure. I just thought that it might be good for avoiding more merge conflicts and stop Aaron patching this on all his other branches. Horrifying that he did this all this time. Aaaand we should have more PRs. Nobody celebrated our 500th merged PR. I'm quite disappointed.

dyfer:
> Nobody celebrated our 500th merged PR. I'm quite disappointed.

😀

psiborg112:
Just to check though, the one thing in this that doesn't just affect me is the change to the install directory in windows - placing JackTrip in the 64 bit Program Files directory. Is that ok with everyone? (Otherwise I can move it out of this PR.)

ntonnaett:
> Just to check though, the one thing in this that doesn't just affect me is the change to the install directory in windows - placing JackTrip in the 64 bit Program Files directory. Is that ok with everyone? (Otherwise I can move it out of this PR.)

Could this impact upgrades? Haven't we changed that in qmake yet?

psiborg112:
It's changed in the windows installer script. It shouldn't affect upgrades, beyond possibly leaving an old version of the file lying around in the former directory. (But this won't be pointed to any more in the start menu, so the user wouldn't see an impact.)

psiborg112:
(It fixes something that was noted tangentially to #725 - that we're incorrectly installing in the 32 bit program files folder.)

ntonnaett:
Sounds fine to me.

PR #752, opened 2022-10-11
Round scalefactor in windows NO_VS builds

psiborg112:
Not rounding it causes widget border issues for non-qml widgets on windows. (At least on windows 10 with Qt5 and Qt6.)


dyfer:
Well, I've added the pass-through scaling factor, otherwise JT looks too small/too large for scaling between 125% and 175%. Is there really no way to keep using fractional scaling and prevent the issues with borders?

psiborg112:
@dyfer, which platform was this needed on, and was this for the standard or virtual studio interface?

dyfer:
@psiborg112 this was for Windows. (MacOS works without that. Linux - I don't know, haven't tested fractional scaling).
Can't member which interface (probably both).

psiborg112:
In my testing I found that setting it or not made no difference on windows 10 to the font size. (At least for 125% scaling.) The font looked small either way. Seems to match the system size with or without the setting in Qt6. (This is just for the standard GUI. No idea about that.) Also, just to confirm that this will only affect NO_VS builds. (So it won't affect the standard packages.)

dyfer:
Not sure if I understand correctly - does that mean that the fractional scaling looks correct _in Qt6_ even if pass-through scaling is not set? Can you confirm that the interface still scales well at 150% and 175%? (I don't have a Qt6 build to try). Is `Qt::HighDpiScaleFactorRoundingPolicy::RoundPreferFloor` needed at all in Qt6 for the fractional scaling to work properly?

If fractional scaling indeed works without this in Qt6, then wouldn't it be better to change the version check so that it is active for Qt >=5.14 and Qt <=6.0 ?

psiborg112:
The issue happens with both Qt5 and Qt6. And I don't see a font size difference happening in the standard UI if I change the setting between the two. It looks too small at 125% either way.

psiborg112:
And to clarify, the widget border issue happens in Qt6 if the rounding policy isn't set.

dyfer:
> The issue happens with both Qt5 and Qt6. And I don't see a font size difference happening in the standard UI if I change the setting between the two. It looks too small at 125% either way.

Have you restarted JT after changing the scaling? It should work...

psiborg112:
Yes, jacktrip restarted, but this is without changing the scaling - i'm just leaving things at 125% the entire time. Changing the Qt::HighDpiScaleFactorRoundingPolicy makes no change to the appearance of fonts in the standard UI. It just alters the widgets. (At least in windows 10.)

psiborg112:
The only difference is that there ends up being more space between widgets, and their borders flash in and out at times depending on mouse and keyboard interaction, and scaling of the window by the user.

dyfer:
Just leaving a note - I need to sign off now but I'd like to play with these settings more (and checking why the fonts do not scale) before merging this.

src/main.cpp
```diff
@@ -155,8 +155,13 @@ QCoreApplication* createApplication(int& argc, char* argv[])
 #endif
         // Fix for display scaling like 125% or 150% on Windows
 #if (QT_VERSION >= QT_VERSION_CHECK(5, 14, 0))
+#if !(defined(NO_VS) && defined(_WIN32))
         QGuiApplication::setHighDpiScaleFactorRoundingPolicy(
             Qt::HighDpiScaleFactorRoundingPolicy::PassThrough);
+#else
```
ntonnaett, line 161: Shouldn't this be `#elif !(defined(NO_VS))` ?

psiborg112: PassThrough works fine in linux (and in fact is needed in some cases). It's really only when we're compiling NO_VS for windows that we want it set otherwise. (The pull request title is a little out of date.)

ntonnaett: Oh sorry, I confused NO_VS with NO_GUI.

dyfer:
On my Windows 10 system I can't see any issues with the current build in the classic mode. Moreover, it seems that the whole interface (widgets, text, spacing) is responding properly to changes in scaling between 125%, 150%, and 175%.

> widget border issues for non-qml widgets

Could you provide a screenshot of the problem?

> And I don't see a font size difference happening in the standard UI if I change the setting between the two. It looks too small at 125% either way.

Can you compare this behavior between your build and the GHA build? As I said, on my Windows box it seems to be working fine...

Which Qt version(s) are you building with?

psiborg112:
Qt 5.15.2 and Qt 6.4.0 (direct from the Qt installer). Can't test at the moment because I don't have my windows machine on me. Widgets and spacings scale, but text doesn't and the borders sometimes disappear. (And this seems to be a known issue. From the Qt docs: The two principal options are whether fractional scale factors should be rounded to an integer or not. Keeping the scale factor as-is will make the user interface size match the OS setting exactly, but may cause painting errors, for example with the Windows style.) Windows 10 has all updates installed.

psiborg112:
If you're certain it's not being seen elsewhere, then I'm happy to change the NO_VS check to a PSI one so it's just my builds.

psiborg112:
(Given that no one else is providing NO_VS builds at the moment, it would effectively be just my builds anyway, but that would just cement it.)

dyfer:
Yes, I'm aware there are warnings about this in Qt docs. However, I couldn't observe these issues myself.

> but text doesn't and the borders sometimes disappear.

This is inconsistent between your observations and mine. I would like to try to find the cause of this before (instead of?) turning off fractional scaling - which seems to work fine on my end.

Next I'll try to build with Qt 5.15.2 (both regular and NO_VS versions) and will compare the results.

Still, it would be good to get a screenshot of the problem once you have access to your windows machine.

> If you're certain it's not being seen elsewhere, then I'm happy to change the NO_VS check to a PSI one so it's just my builds.

I'd rather find out why the fractional scaling behaves differently between the two builds - I might be wrong but my objective would be to make fractional scaling behave properly in the NO_VS build.

psiborg112:
In lieu of a screenshot, just imagine that instead of having black borders on each side, some of the text and combo boxes only have 3 or 2, and the appearance of this is inconsistent. Which borders are painted changes depending on window size and whether the boxes are selected.

psiborg112:
(And hopefully I'll actually be able to send a shot of this later today.)

psiborg112:
Screenshot 2022-10-14 161449
From the official branding, you can see that this is with the github 1.6.4 release, not my own build. You can see the missing borders. The font is this size whether the policy is set to PassThrough (as it is in the official build) or whether it's set to be rounded.

dyfer:
The locally built JT (`dev` branch, qmake, with virtual studio, using Qt 5.15.2 and mingw 8.1) still scales properly on my system.

I couldn't get CMake build to work... it configures and builds, but it doesn't create the actual executable. 
I've run
```
mkdir build2 && cd build2
cmake ..
cmake --build . --config Release --target install
```
I might have paths configured wrong etc. Could you provide a minimal set of commands you use when building with cmake so I can see what I'm doing wrong?

QMake build with `novs` option ends up erroring out with `../src/main.cpp:501:66 error: 'updateChannel' was not declared in this scope` so I couldn't create novs build in the end...

psiborg112:
It's not really a NO_VS build specific issue as can be seen in the above screenshot. I just didn't want to interfere with how things were done in the official builds in case scaling was vital for the VS stuff, so was restricting the change to NO_VS builds.

psiborg112:
And I just use cmake within kdevelop (installed via chocolatey) so couldn't give you a list of commands offhand. But kdevelop seems to run the following for configuring:
`C:/Users/micro/Documents/QJackTrip/builddir> "C:/Program Files/CMake/bin/cmake.exe" "-DCMAKE_EXPORT_COMPILE_COMMANDS=ON" "-DCMAKE_INSTALL_PREFIX=C:/Program Files (x86)/QJackTrip" "-DCMAKE_BUILD_TYPE=Release" C:/Users/micro/Documents/QJackTrip`

And then for building:
`C:/Users/micro/Documents/QJackTrip/builddir> make -j6`

psiborg112:
Yeah, I see the error. I've never built without NO_UPDATER set. (The cmake builds aren't configured to deal with any of the dblsqd stuff so I've never bothered with it.)

updateChannel is created far, far earlier than it needs to be from what I can tell. It should be much closer to the error that's being thrown at line 506.

dyfer:
This is how GHA build (dev branch) looks on my screen:
<details>

100%
jt100

125%
jt125%

150%
jt150%

</details>

I also noticed the disappearing borders this time.
However, I'd prefer disappearing borders with properly scaled interface, vs text/window that is too small/too large... provided that it actually works, of course. I also understand if other might prefer it the other way around...

psiborg112:
Yeah, I definitely prefer an application to not look like it's interface is broken. Especially since I'm not seeing the scaling on my machine. Can't tell you why that's the case though.

psiborg112:
Happy to edit this to make it just a feature of my builds if others prioritise text scaling over UI consistency.

dyfer:
Ideally, I'd want to have both fractional scaling and non-broken look. Seems like this is a tall order for Qt though :/

(It probably doesn't change anybody's mind, but the borders seem to come back once the window is resized/refreshed.)

Given that there doesn't seem to be a reasonable way to achieve this and the behavior is not even consistent for the current official build, I'd be okay to accept this PR as is - to make this change for NO_VS build. Win32 check is superfluous I believe, but can stay.

The one thing I want to test whether the NO_VS build will resize _down_ when system setting is at 150% (I understand this is what should happen with the current setting in this PR, but I'd like to check that). One reason I am so adamant about scaling is that when the window is too large and appears partially off screen (as it does if it rounds 150% up), it's hard to resize it...

psiborg112:
Because of the prefer floor setting, 150% will be rounded down. And it's the flickering of the borders in and out that really contributes to the broken feel.

psiborg112:
Also, fractional scaling does exist on linux, and seems to work fine in both Qt5 and Qt6. So the windows check is definitely needed there.

dyfer:
> Also, fractional scaling does exist on linux, and seems to work fine in both Qt5 and Qt6. So the windows check is definitely needed there.

So are there no artifacts when using pass-through scaling on Linux? In that case sure, let's do this on Windows only.

dyfer:
> Because of the prefer floor setting, 150% will be rounded down. And it's the flickering of the borders in and out that really contributes to the broken feel.

I understand that's what it should do. I just want to see it do it, considering how reliable Qt is about scaling...

psiborg112:
Seems to do it on my system, but yes, it would be good to have a second opinion on that.

dyfer:
> And then for building:
> `C:/Users/micro/Documents/QJackTrip/builddir> make -j6`

So do I need to have `make` installed? or is this mingw32-make? When I run `cmake ..` it seems to configure it with MSVC.

>  I've never built without NO_UPDATER set. 

Okay, building with `noupdater` now.

psiborg112:
#760 should fix that issue, so you can build with it. And as far as I'm aware it is mingw make.

dyfer:
Okay, using https://github.com/jacktrip/jacktrip/pull/760 I was able to build and compare the rounding policies.

TL;DR - I'm sorry for the noise. It looks like when JT is restarted after scaling change, `RoundPreferFloor` actually looks better than PassThrough. There's a difference in text size, but not as much.

For the classic interface, I'd be leaning towards changing the rounding policy to `RoundPreferFloor` for _all_ builds.

However, the VS interface does look better with pass through policy and there don't seem to be any visual artifacts. I don't know what to think about this change anymore... (since I don't think we can change policy when switching interfaces)

psiborg112:
Pretty sure we can't change policy with the interface change: it's annoyingly an all or nothing thing. I think when we're including VS stuff we have to err on the side of what looks good for that. The people who are likely to be using classic mode are probably the ones who would be more comfortable downloading a custom build to improve their experience.

dyfer (approved):
Given all the discussion and testing, I think this change is fine.

Please squash on merge

PR #755, opened 2022-10-13
Use Overlapped I/O for Windows Networking

psiborg112:
This change finally brings windows networking into line with what we have for linux and mac: rather than waking the receiving thread every 0.1ms to check if a packet has arrived, the thread is woken up when network traffic is received. (This old behaviour can be re-enabled by defining MANUAL_POLL in UdpDataProtocol.cpp.)


psiborg112:
This is a significant change and requires a lot of testing. In my initial tests, it seems to perform on par with the old system. (Not noticing any appreciable change in audio quality.) Auto queue settings remain the same, but I am occasionally getting a few more UDP waiting too long messages. This may be a result of how we're measuring this though: by incrementing the elapsed time each time the thread is woken. With such a high resolution, our manual polling mechanism might not be waking up every 0.1ms and could be significantly under-reporting how much time has passed. (While our lower res, 10ms timeout on our kernel notification mechanisms is unlikely to suffer from this.)

psiborg112:
And funnily enough the UDP waiting too long message frequency is back on par with what's expected when I remember to reset the wait time on packet arrival.

nwang92:
@Rjapeck we should see if these builds do anything to address https://github.com/jacktrip/jacktrip/issues/648

psiborg112:
We can hope, but I'm not holding my breath. Having said that though, it will give us a better place to start from in addressing it, and it does make the Windows networking code more modern. @mikedickey measured a 3-4x decrease in CPU usage on other platforms when we made the shift, and hopefully we'll see a similar improvement here.

mattahorton:
What do we need to do to test this? Just use it on Windows?

psiborg112:
Yes, and ideally this should have gone into the latest rc so that it could have seen real world testing before release.

psiborg112:
(I imagine it'll have to wait until the next release cycle now.)

ntonnaett:
That's not okay. What is the reason this wasn't considered before the release?

mattahorton:
It wasn't an intentional decision. It just got missed.

forsythac:
I'm sure it was not intentional- we have lots of Windows users out there and care very much about improving the networking, especially given that we have a long standing critical bug in the area.

I'm sure that we just have not spent the time required, since
`This is a significant change and requires a lot of testing.` 
is a scary message

How do we test this to get it included in the next release? We have windows testers including myself and multiple devs who are ready to test the change if we can just make a build with the change and understand how to test it.
What we won't do is just include this in the next release by default and test it in Live after a dev calls for significant testing.

dyfer:
> What we won't do is just include this in the next release by default and test it in Live after a dev calls for significant testing.

OTOH one could argue that this is what `rc` releases are for.

psiborg112:
What @dyfer said. A release candidate would have allowed us to test this across a wide variety of machines to make sure this worked before it went out to the broader public. And as pointed out in the initial commit, the old behaviour is still in the code and is a preprocessor switch away. If we needed more time to refine it based on what we learnt in the release candidate we could have shipped the next release with the code disabled.

psiborg112:
(The exact same thing happened for the changes on mac and linux. We had the option to revert if things went pear shaped.)

mattahorton:
Yeah this should have gotten in the RC we did cut. It should go in the next one.

mattahorton (approved):
Approving so we can [test in the first RC for 1.6.7](https://github.com/jacktrip/jacktrip/discussions/784#discussioncomment-4029512)

PR #757, opened 2022-10-14
Attempting to fix detached QVector crash

nwang92:
There's something in VS gui mode that's periodically causing jacktrip to crash. It happens very rarely for me, but the last log line I get is:
```
ASSERT: "isDetached()" in file /opt/homebrew/opt/qt@5/include/QtCore/qvector.h, line 407
```
Some initial research suggests there's a thread-safety problem when using QVectors:
* https://forum.qt.io/topic/90180/assert-isdetached-with-multithreaded-qvectors
* https://interest.qt-project.narkive.com/aVUB82lb/assert-isdetached-with-multithreaded-qvectors
* https://blog.actorsfit.com/a?ID=01150-35eca5ca-9371-43ac-bbcd-5e3757c9940d

I'm not 100% certain it's coming from here, but seems like a possibility? This probably requires more testing

### DIFF
--- a/src/Meter.cpp
+++ b/src/Meter.cpp
@@ -112,7 +112,9 @@ void Meter::compute(int nframes, float** inputs, float** /*_*/)
         }
 
         /* Update mValues */
+        mValuesMutex.lock();
         mValues[i] = max;
+        mValuesMutex.unlock();
     }
 
     /* Set processed audio flag */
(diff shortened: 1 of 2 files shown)


src/Meter.cpp
```diff
@@ -143,9 +147,11 @@ void Meter::onTick()
         emit onComputedVolumeMeasurements(mValues);
```
nwang92, line 147: Another thought/area of concern was how `mValues` gets passed here but it looks like QT has a copy-on-write paradigm built in? https://doc.qt.io/qt-5/implicit-sharing.html#implicit-sharing-in-detail

nwang92:
Closing in favor of https://github.com/jacktrip/jacktrip/pull/763

PR #760, opened 2022-10-14
Allow NO_VS to build without NO_UPDATER set

psiborg112:
Even though it's probably not the most practical idea without a separate update stream.


psiborg112:
One thing doing this made me notice is that at 375 in main there's a check for linux within a block that will only be compiled if we're building for windows. Looks like this is unnecessary and can go.

dyfer (approved):
Works as advertised, as far as I can tell.

Issue #537, opened 2022-04-18
GUI tweak for small displays


cchafe:
I'm running JackTrip on a PinePhone and have the problem that the "connect" button and other bottom buttons are off the bottom of the display. My workaround is to rescale the phone's display but maybe it's possible to shrink the size of the GUI without losing any functionality? Seems there's a bit of vertical territory that's unused at this point.

psiborg112:
Already on my todo list, but might be a while before I have a chance to get around to it. Will keep you posted.

ntonnaett:
This prevents me switching to vs mode on my laptop.

psiborg112:
What resolution is your laptop display?

psiborg112:
I keep forgetting that I'm not as affected by this as others because my builds have the "Use Authentication" settings under the basic options tab, which keeps the advanced options tab a little more under control in hub client mode. Now that there's a dedicated virtualstudio interface, moving this back to the basic options tab could help to mitigate this issue.

dyfer:
It would be good to be able to squish down the main window, even if the contents need to be scrolled, for the connect button to be available. I've just encountered this due to incorrect pixel scaling (Windows 150% scaling which seems to render as 2x scaling in JT) and couldn't scale down the window vertically. However, maximizing the window made the "connect" button available.

PR #764, opened 2022-10-15
Reunify the standard GUI variants

psiborg112:
The original GUI was never, ever meant to get as tall as it did, and I always held that moving the auth section to the advanced options tab to appease the VS folk was a major design flaw. (And one that I've been almost entirely insulated from by maintaining my own version of the interface.) Now that they have their own UI, this moves the auth section back where it belongs, with a disclaimer to use the VS interface if that's what you're after.


psiborg112:
This may be enough of a fix to put #537 to rest.

src/gui/qjacktrip.cpp
```diff
@@ -108,7 +100,11 @@ QJackTrip::QJackTrip(int argc, bool suppressCommandlineWarning, QWidget* parent)
         About about(this);
         about.exec();
     });
-#ifndef NO_VS
+#ifdef NO_VS
+    m_ui->authNotVSLabel->setText(QStringLiteral(
+        "(This is for JackTrip's inbuilt authentication system. To easily "
```
dyfer, line 105: Suggestion:
`This is for JackTrip's inbuilt authentication system.` -> `This is for JackTrip's inbuilt authentication system, for the Virtual Studio.`

src/gui/qjacktrip.ui
```diff
@@ -440,90 +339,275 @@ To connect to a hub server you need to run as a hub client.</string>
...
-              <cstring>autoPatchComboBox</cstring>
+              <cstring>usernameEdit</cstring>
              </property>
             </widget>
            </item>
            <item row="1" column="0" colspan="2">
-            <widget class="QCheckBox" name="patchServerCheckBox">
+            <widget class="QLabel" name="authNotVSLabel">
+             <property name="text">
+              <string>(This is for JackTrip's inbuilt authentication system. To easily connect to a Virtual Studio server, use the Virtual Studio interface.)</string>
```
dyfer, line 416: Suggestion as above 
`This is for JackTrip's inbuilt authentication system.` -> `This is for JackTrip's inbuilt authentication system, for the Virtual Studio.`
And:
`use the Virtual Studio interface` -> `use the Virtual Studio Mode` (that's the label of the button)

src/gui/qjacktrip.ui
```diff
@@ -440,90 +339,275 @@ To connect to a hub server you need to run as a hub client.</string>
...
+             </property>
+             <property name="text">
+              <string>Browse</string>
+             </property>
+            </widget>
+           </item>
+           <item row="1" column="0" colspan="3">
+            <widget class="QLabel" name="authDisclaimerLabel">
+             <property name="text">
+              <string>(This is a work in progress and needs to be manually configured outside of the app. Only use if you know what you're doing.)</string>
```
dyfer, line 461: No technically part of the changes in this PR, but is this still the case? Is there any documentation on this?

dyfer (commented):
I think this change is good.
I have a couple of suggestions - IMO spelling out explicitly that this is not for Virtual Studio is beneficial, even though it would cause repeating "Virtual Studio" event more...

src/gui/qjacktrip.ui
psiborg112: How about this then? (Forgot the button was directly below because it's not in my builds.)
"(This is for JackTrip's inbuilt authentication system. To connect to a Virtual Studio server, press the Virtual Studio Mode button below.)"

src/gui/qjacktrip.cpp
psiborg112: Then we'd have it three times on a build where it's disabled, the installation of which was probably a conscious choice.

src/gui/qjacktrip.ui
psiborg112: (I've been using it for quite some time to manage which classes can connect to our university jacktrip server at which times.)

src/gui/qjacktrip.cpp
psiborg112: I'm just wary of the explanatory text taking up more vertical screen space than the actual important, actionable parts of that group box.

src/gui/qjacktrip.ui
psiborg112: For the VS enabled build, it's probably less of an issue repeating Virtual Studio, so it could actually be:
(This is for JackTrip's inbuilt authentication system, not for the Virtual Studio. To connect to a Virtual Studio server, press the Virtual Studio Mode button below.)

src/gui/qjacktrip.cpp
psiborg112: That said maybe replacing 'build' with 'version of the program' is a little more layperson friendly?

dyfer: Good point - I was looking at the Virtual Studio build.

src/gui/qjacktrip.ui
dyfer: Thanks! I think the updated version is great.

dyfer (approved):
This looks good to me!

PR #766, opened 2022-10-18
Show Qt Version in about dialog

psiborg112:
Now that people are starting to build with different versions of Qt, show which version the app has been built with in the about dialog box.


dyfer (approved):
Perfect, thanks! I was just thinking why we don't have that info posted anywhere.

PR #769, opened 2022-10-21
Pre-release 1.6.5-rc.1

mattahorton:
There's a lot of good work we should start getting into people's hands. Let's cut a pre-release and start testing with it.


nwang92 (approved):

psiborg112:
Damn. I was going to suggest that #755 should be in any pre-release we make. This is an important change for windows networking which really should make it into the next release version.

PR #772, opened 2022-10-21
Enable Non-ASIO Audio Devices on Windows

dhing1024:


ntonnaett:
This would be much better if you found a generic way of doing this. The code should look the same for all audio backends.

mattahorton:
I appreciate the edit for tone @ntonnaett, but the point is welcome. 

I wonder how we would do that. We could have some kind of key/value store – I don't know what the appropriate C++ store is – that we append to if creation of the RtAudio instance succeeds. That way we don't run into issues with unsupported backends on various platforms. That would be instead of using preprocessor directives to prevent things like WASAPI and DirectSound code from running on macOS and Linux.

ntonnaett:
There's `RtAudio::getCompiledApi(std::vector< RtAudio::Api >& apis)` for getting a vector of all compiled APIs. Of course those APIs could also fail, if the sound server is not running. Then there are `std::string RtAudio::getApiName(RtAudio::Api api)` and `std::string RtAudio::getApiDisplayName(RtAudio::Api api)` for getting the name to the Enum entry.

dhing1024:
I think this pull request is ready for review at this point. These changes allow RtAudio to select audio devices from any audio driver on the machine, rather than just the first one it selects by default. In particular, this will allow Windows users to select non-ASIO devices, if their machine comes with an ASIO driver already installed.

Much of the more recent work has focused on handling cases where the user tries to select two devices from different audio drivers, for example, a WASAPI input and an ASIO output. Even with these changes, this functionality is not available, since it looks like sometimes cross-driver devices are just not going to be compatible, though I couldn't figure out a coherent set of rules why. To address this possibility, if two different drivers are selected, then the user will not be permitted to save their device settings, both in the "Setup" screen and in the "Settings" screen, with some UI features to make it more clear what actions the user should take. (This is only in VS mode; in classic mode the UI doesn't warn the user of the incompatibility and I'm not sure what the right way to do that would be but for now I'm just stopping the JackTrip process).

Will include screenshots in another comment.

dhing1024:
Setup Screen:

https://user-images.githubusercontent.com/26987971/200079413-41e6a2a9-2d0d-4fed-98a9-5e786805d51b.mp4

Settings Screen:

https://user-images.githubusercontent.com/26987971/200079419-dd6ae890-3e5d-47dc-a0b8-5a938afaa6e8.mp4

src/RtAudioInterface.cpp
```diff
@@ -312,40 +359,109 @@ int RtAudioInterface::wrapperRtAudioCallback(void* outputBuffer, void* inputBuff
...
 }
+
+//*******************************************************************************
+void RtAudioInterface::getDeviceList(QStringList* list, bool isInput)
+{
+    list->clear();
+    list->append(QStringLiteral("(default)"));
+
+    std::vector<RtAudio::Api> apis;
+    RtAudio::getCompiledApi(apis);
```
nwang92, line 418: I noticed there's a very short but noticeable delay on your videos when switching devices. Does it make much of a difference if we do this in `setup()` vs each time the device list is retrieved? I imagine the list of APIs available to rtaudio won't change much during a session. If I understand what this function is supposed to do, it seems like it might only matter if you install a new audio backend while jacktrip is already open?

But if this isn't contributing to the delay, then nvm - seems safe to leave it in.

dhing1024: I think on Windows, switching devices just is a little slower and laggy, even without these changes. Versus on Mac I've never really noticed any lag. So I'm not sure how much there is we can do about it without a substantial code refactor.

ntonnaett: Could you close the list independently from the "starting" of the AudioInterface?

dhing1024: Hmm that might be a good way to make it a smoother experience. I'll look into it on Monday.

dhing1024: Took a look at this - it might be possible but I don't think there's a quick and easy way to do it.

nwang92 (approved):

src/gui/vsAudioInterface.cpp
```diff
@@ -138,12 +153,28 @@ void VsAudioInterface::setupAudio()
...
             m_numAudioChansIn  = m_audioInterface->getNumInputChannels();
             m_numAudioChansOut = m_audioInterface->getNumOutputChannels();
             // Setup might have changed buffer size
             m_audioBufferSize = m_audioInterface->getBufferSizeInSamples();
+
+            std::string devicesWarningMsg = m_audioInterface->getDevicesWarningMsg();
+            std::string devicesErrorMsg   = m_audioInterface->getDevicesErrorMsg();
+
+            if (devicesWarningMsg != "") {
+                std::cout << "Devices Warning: " << devicesWarningMsg << std::endl;
```
mattahorton, line 168: It's better to use qDebug here as that logs to the log file

src/gui/vsAudioInterface.cpp
```diff
@@ -138,12 +153,28 @@ void VsAudioInterface::setupAudio()
...
+            }
+
+            if (devicesErrorMsg != "") {
+                std::cout << "Devices Error: " << devicesErrorMsg << std::endl;
```
mattahorton, line 172: also qdebug

src/gui/vsAudioInterface.cpp
```diff
@@ -155,12 +186,27 @@ void VsAudioInterface::setupAudio()
...
             m_numAudioChansIn  = m_audioInterface->getNumInputChannels();
             m_numAudioChansOut = m_audioInterface->getNumOutputChannels();
             // Setup might have changed buffer size
             m_audioBufferSize = m_audioInterface->getBufferSizeInSamples();
+
+            std::string devicesWarningMsg = m_audioInterface->getDevicesWarningMsg();
+            std::string devicesErrorMsg   = m_audioInterface->getDevicesErrorMsg();
+
+            if (devicesWarningMsg != "") {
+                std::cout << "Devices Warning: " << devicesWarningMsg << std::endl;
```
mattahorton, line 201: same qdebug comment. basically, if it's Qt, try to use qDebug

mattahorton (changes requested):
Last comment here. We're missing 1 acceptance criteria.

* Display of list should be separated by "Low Latency Devices (ASIO)" and "All Devices (non-ASIO)"

I don't think those titles are correct, but especially if we're popping a warning that people need to select the same kind of driver, we need to separate them.

I honestly had a hard time knowing which version I was running.

Old, ASIO-only
oldjtdevices

New, all devices
newjtdevices

dhing1024:
Should be good to go now, if there are no other issues. Adds in option groups to the dropdown list for Windows only but leaves in (default) as an option, which has some issues on Windows if ASIO drivers are present. That will be addressed separately.

Mac:
Screen Shot 2022-11-09 at 3 03 33 PM

Windows:
[will add picture here, just need to switch laptops lol]

nwang92 (approved):

mattahorton (approved):
So @dhing1024 I'm going to approve this. 

2 thoughts:
1. I have a lot of devices. Because of that, the list doesn't show the words "Low-Latency" at the top. It seems to prefer the bottom of the list being shown on open. If we could figure out how to prefer the top of the list, that would be ideal. Non-blocking.
2. Has the Windows app gotten _slower_? It's _chugging_.

PR #773, opened 2022-10-21
Show input vu meter in settings

nwang92:
I realized it was pretty trivial to show the input VU meters in the settings (thx to the handy `Meter` QML component) I'm currently hiding it for JACK because I can't seem to start my local JACK server and don't know what the behavior should be.

I don't know if we want this but I figured I'd throw the PR up
Screen Shot 2022-10-21 at 3 28 27 PM

### DIFF
--- a/src/gui/Settings.qml
+++ b/src/gui/Settings.qml
@@ -4,7 +4,7 @@ import QtQuick.Controls 2.12
 Item {
     width: parent.width; height: parent.height
     clip: true
-    
+
     Rectangle {
         width: parent.width; height: parent.height
         color: backgroundColour
@@ -13,12 +13,12 @@ Item {
     property int fontBig: 28
     property int fontMedium: 13
     property int fontSmall: 11
-    
+
     property int leftMargin: 48
     property int rightMargin: 16
     property int buttonWidth: 103
     property int buttonHeight: 25
-    
+
     property string backgroundColour: virtualstudio.darkMode ? "#272525" : "#FAFBFB"
     property string textColour: virtualstudio.darkMode ? "#FAFBFB" : "#0F0D0D"
     property string buttonColour: virtualstudio.darkMode ? "#494646" : "#EAECEC"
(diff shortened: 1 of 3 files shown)


mattahorton:
Where is the audio coming from?

If it's coming from VsAudioInterface, JACK should work. Though, maybe I handled that in the UI by catching an error and switching back to RtAudio. In any case, that should be an easy follow-up.

I think volume should be pretty trivial as well.

mikedickey (approved):

mattahorton:
This is not done. Please do not merge

mattahorton (changes requested):
I'm going to try and make JACK work like it does in setup

mattahorton:
First off, here are a few UI tweaks for a couple purposes:
1. Making the UI work with just JACK
2. Visual clarity

CleanShot 2022-10-24 at 16 38 27@2x
CleanShot 2022-10-24 at 16 38 33@2x

Second, here is how JACK not running should look. There are zero code changes here. I saw an issue where the error that triggers this UI wasn't working, for some reason, but it went away before I could debug it. Now I cannot reproduce.

If someone else can reproduce, the relevant code is somewhere between lines `167-177` of `JackAudioInterface.cpp`. If JackTrip gets through these lines, then line `365` returns a warning that there are no ports.

CleanShot 2022-10-24 at 16 42 26@2x

nwang92:
I pulled your changes and ran a build - so far seems fine to me. I like the placement changes.

I don't feel strongly about this, but it's a bit strange that the stacked order of "Output Device" and "Input Device" isn't the same between Setup and Settings. Just a minor comment now but I don't think we need to address it here. I can't approve my own created PR so I verbally approve.

mattahorton:
> I pulled your changes and ran a build - so far seems fine to me. I like the placement changes.
> 
> I don't feel strongly about this, but it's a bit strange that the stacked order of "Output Device" and "Input Device" isn't the same between Setup and Settings. Just a minor comment now but I don't think we need to address it here. I can't approve my own created PR so I verbally approve.

I agree it's weird, but I don't think it's breaking. I guess I'll approve even though I dismissed my original review?

mattahorton (approved):

PR #786, opened 2022-10-30
Use default keychain for notarization

psiborg112:
Notarytool seems to use the login keychain by default for credential storage. Make sure this is set.

### DIFF
--- a/.github/workflows/jacktriplabs.yml
+++ b/.github/workflows/jacktriplabs.yml
@@ -457,7 +457,8 @@ jobs:
           CERTIFICATE: ${{ secrets.APP_CERT_NAME }}
           PACKAGE_CERT: ${{ secrets.INSTALLER_CERT_NAME }}
           USERNAME: app@jacktrip.org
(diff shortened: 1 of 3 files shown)


psiborg112:
I'm not sure that this will be enough after trying to test locally. Trying to change the login-keychain locally doesn't seem to work. (Their security man page is really not very useful.)

mattahorton:
I've gotten slightly farther than this over on jacktriplabs/jacktrip where I was testing just so I had all the existing secrets. In that attempt, I allowed a keychain path to be passed in, since the script already creates one. But notarytool is complaining about incorrect credentials.

mattahorton:
I just opened #787 with a successful fix.

psiborg112:
Does this fix not work though? It would be the better option because there's no hard coding of the keychain path - it's detected programmatically. Also, your script doesn't account for cases where a keychain doesn't need to be specified.

psiborg112:
Superseded by #788

Issue #785, opened 2022-10-29
1.6.5 deleted?!


dvzrv:
Hi, I'm packaging jacktrip for Arch Linux.

Yesterday there was still a 1.6.5 release and tarball and I have packaged using that: https://github.com/archlinux/svntogit-community/commit/b5a2fc2f96fb4f09102f8cd491e6969048d31046

What happened to that tag? Please don't delete tags! It breaks downstream trust and (reproducible) builds!

dvzrv:
ping @psiborg112 @mattahorton @ntonnaett @cchafe

dvzrv:
This is the source tarball that was generated by github, from which I built:
[jacktrip-1.6.5.tar.gz](https://github.com/jacktrip/jacktrip/files/9895495/jacktrip-1.6.5.tar.gz)

The diffoscope diff to 1.6.5-rc1 is:
[jacktrip-diffoscope.log](https://github.com/jacktrip/jacktrip/files/9895497/jacktrip-diffoscope.log)

dvzrv:
While the diff does not seem harmful I will force downgrade the package now.
A heads up or even response to this ticket would have been nice...

psiborg112:
Releases are handled by the JackTrip Labs crew because of binary signing on mac and windows, and not by the open source contributors. (@mattahorton is probably the person to contact for these things.) In this case they pulled the release because of an issue with with notarization on mac. (I updated a script because apple have changed their notarization workflow. It works fine on my local machine, but not in github actions, and we need to find out why.) I suspect nothing will change in the code between the release being pulled and made available again. But don't quote me on that.

dvzrv:
> In this case they pulled the release because of an issue with with notarization on mac.

Why pull the release that others already rely upon (without notice) in that case though? It is just the notarization for macOS failing, not the build.
For introducing post-release artifact fixup releases for macOS you can e.g. rely upon [build metadata](https://semver.org/#spec-item-10), as you seem to closely follow semver. They order properly after e.g. 1.6.5 but before 1.6.6. This is e.g. how SuperCollider deals with the mess that is notarization in case it fails for the respective release (there is no and never will be a guarantee that this brittle service will always work out of the box).

> I suspect nothing will change in the code between the release being pulled and made available again. But don't quote me on that.

From a downstream perspective this doesn't really matter, as the release tarball has changed once you tag 1.6.5 again and breaks the build (the checksum is locked).
[...]

PR #787, opened 2022-10-31
Set keychain path for notarization

mattahorton:
This [fixes notarization](https://github.com/jacktriplabs/jacktrip/actions/runs/3363802016). It borrows the same app signing keychain created for use in the github actions script.

We'll need to get this to dev and then main and then redo the release creation.

### DIFF
--- a/.github/workflows/jacktriplabs.yml
+++ b/.github/workflows/jacktriplabs.yml
@@ -532,6 +532,8 @@ jobs:
         with:
           name: ${{ env.CLANG_TIDY_NAME }}
           path: ${{ env.CLANG_TIDY_PATH }}/
+      - name: Run tmate
+        uses: mxschmitt/action-tmate@v2
   # release - list of files uploaded to GH release is specified in the *upload* step
   deploy_gh:
     needs: [sign, check-secrets]
--- a/.github/workflows/jacktriplabs.yml
+++ b/.github/workflows/jacktriplabs.yml
@@ -533,7 +533,8 @@ jobs:
           name: ${{ env.CLANG_TIDY_NAME }}
           path: ${{ env.CLANG_TIDY_PATH }}/
       - name: Run tmate
-        uses: mxschmitt/action-tmate@v2
+        if: ${{ failure() }}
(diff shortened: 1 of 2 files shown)


ntonnaett:
@dvzrv pointed out that the checksums for version 1.6.5 is locked downstream. So we might want to skip this version and release 1.6.6.

mattahorton:
That shouldn't block this PR

nwang92 (approved):

psiborg112:
Could you please test #786? This will break my workflow.

psiborg112:
And it relies on the assumption that apple will never change the way its security tool generates keychain names and locations.

dvzrv:
> So we might want to skip this version and release 1.6.6.

As mentioned in https://github.com/jacktrip/jacktrip/issues/785#issuecomment-1297701613 this might not be necessary, as the only downstream (that I am aware of currently) still/ already packaging 1.6.5 is Manjaro which will eventually get the package that I am building anyways, as they don't build their own packages (mostly).

For the future it might be worth looking into using semver's [build metadata](https://semver.org/#spec-item-10) for situations in which not all build artifacts behave the way you expect them to on a release (see https://github.com/jacktrip/jacktrip/issues/785#issuecomment-1297261428).

mattahorton:
> For the future it might be worth looking into using semver's [build metadata](https://semver.org/#spec-item-10) for situations in which not all build artifacts behave the way you expect them to on a release (see [#785 (comment)](https://github.com/jacktrip/jacktrip/issues/785#issuecomment-1297261428)).

Honestly, another part of the solution is to not fail signing all signed artifacts when one platform fails. The reason I felt it necessary to delete the release at the time (agreed in retrospect it was a bad decision), was that we now had unsigned Windows builds sitting in the release where many people have been pointed to for downloads for years.

If we were able to provide the appropriate builds, maybe just delete the macOS ones and then add them back once they've been signed and notarized, that would be better. But our current scripts rely on both platforms succeeding to upload any artifacts at all.

mattahorton:
Closing in favor of #788

PR #788, opened 2022-11-01
Allow the use of a temporary keychain for notarization

psiborg112:
Allow the assembly script to create a temporary keychain for notarization so that it can run regardless of what the system keychain settings are.


psiborg112:
This is an alternative to #786. Only one of them should be merged.

mattahorton (approved):
Tested and working on a repo with the appropriate secrets

PR #791, opened 2022-11-02
Update manifests for 1.6.6

mattahorton:
Final step for releasing 1.6.6


nwang92 (approved):

nwang92:
Not necessary for this PR, but we should also update `edge`

mattahorton:
> Not necessary for this PR, but we should also update `edge`

going to fix that now actually

nwang92 (approved):

Issue #797, opened 2022-11-08
Flatpak builds broken


mattahorton:
[Looks like something about weakjack?](https://github.com/jacktrip/jacktrip/actions/runs/3421793575/jobs/5698402159)

ntonnaett:
This seems to be related to a cve: https://github.blog/2022-10-18-git-security-vulnerabilities-announced/#cve-2022-39253

ntonnaett:
Opened an issue for the gha: https://github.com/flatpak/flatpak-github-actions/issues/106

ntonnaett:
Fixed in flatpak-builder: https://github.com/flatpak/flatpak-builder/pull/497
Flathub included the fix: https://github.com/flathub/flathub/issues/3620

ntonnaett:
There was no flatpak-builder release since Jan. Just asked for a release. This will break our flathub builds too.

PR #798, opened 2022-11-09
VU meters in standard mode

psiborg112:
Adds a VU meter window to the standard GUI.


psiborg112:
If someone could test that this works OK with RtAudio, that would be great. I'm particularly interested in checking the scenario where more audio channels are requested than are available on the interface.

dyfer:
I think that adding VU meters for non-vs mode is great and needed, thanks!

I'm just not sure if I'm convinced by the idea of a separate window.
While this might require more thought, I'd prefer if the meters were integrated in the main window. The main window ideally shouldn't have controls inactive while connected IMO so that one can switch between tabs. OTOH I'm not sure which tab should meters should be added to...

> If someone could test that this works OK with RtAudio, that would be great. I'm particularly interested in checking the scenario where more audio channels are requested than are available on the interface.

You probably know this but FYI - you can always download the regular build from github actions (which has the rtaudio backend). I tried requesting more channels than available on the device and it seems fine (meters show up but no signal appears).

psiborg112:
It's more an issue that RtAudio just ends up being a layer over JACK for me on my main dev machine. I think eventually it would be good to rework the GUI: when it was first created there weren't as many options to include, so tabs made sense. Most of this should probably move to a separate settings window now. And when that happens, I'll move the meters back to the main window.

psiborg112:
And yeah, that's what I suspected would happen. (Empty meters with no signal.) Just wanted to make sure it didn't crash. (Looked like it shouldn't but always worth checking.)

psiborg112:
Unless... Have had an idea. Leave it with me.

psiborg112:
Revised

dyfer:
I think this works better.

I think this is a considerable UI change so I'd like others to chime in before merging this. However, this works for me. Here's a screenshot of JT once connected so that others can take a look without downloading:
Screen Shot 2022-11-09 at 10 04 18


(Also, I'm not sure what's wrong with Flatpak build failure, but I don't think it's related to this PR. @ntonnaett any ideas?)

dyfer (approved):
Works for me, but I'd like another approval before merging this.

ntonnaett:
The flatpak build is failing because a security bug in git was fixed that breaks the flatpak-builder gha thingy. #797

mattahorton:
This is great!

PR #804, opened 2022-11-12
remove extraneous call to readSlotNonBlocking in Regulator.h

cchafe:
remove extraneous call to readSlotNonBlocking in Regulator.h virtual void readBroadcastSlot, leftover lint from initial broadcast method for bufstrategy 3 (PLC)


cchafe:
fails flatpak but should be ok to go ahead with this one (see #797)

mikedickey (approved):

PR #811, opened 2022-11-17
Fix updater crash on Ventura

psiborg112:
The updater code makes use of regular expressions - functionality provided by pcre2 which uses a just in time compiler to work its magic. This doesn't seem to be allowed by default in a hardened runtime anymore. We need to explicitly add a jit entitlement when signing.


nwang92 (approved):

mattahorton (approved):
Not sure if you were ready for review, but this is pretty straightforward

psiborg112:
I was.

PR #817, opened 2022-11-29
Fix for client connection failures on Windows

mikedickey:
Fixes https://github.com/jacktrip/jacktrip/issues/648

I don't claim to understand the comment, but this conditional for Windows causes it to hang for 10 seconds when trying to establish a connection to a hub server. Sadly, 10 seconds is also the amount of time that the hub server waits to receive the first datagram from client. This means that Windows clients are always in a tight race to establish a successful connection in time.

I believe this in particular happens when an IP address is used, causing QHostInfo to perform a reverse lookup, and there is no reverse lookup entry found for it. Windows just seems to deadlock for 10 seconds what that happens.

If this code is good enough for JackTrip.cpp to use when establishing the initial TCP connection to a hub server, it should be good enough for the UDPDataProtocol as well.

### DIFF
--- a/src/UdpDataProtocol.cpp
+++ b/src/UdpDataProtocol.cpp
@@ -127,11 +127,7 @@ UdpDataProtocol::~UdpDataProtocol()
 void UdpDataProtocol::setPeerAddress(const char* peerHostOrIP)
 {
     // Get DNS Address
-#ifndef _WIN32
-    // Don't make the following code conditional on windows
-    //(Addresses a weird timing bug when in hub client mode)
     if (!mPeerAddress.setAddress(peerHostOrIP)) {
-#endif
         QHostInfo info = QHostInfo::fromName(peerHostOrIP);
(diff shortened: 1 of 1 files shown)


psiborg112:
It's a relic from the 1.2 (or maybe 1.3) days. I could never reproduce the issue that it fixed, but @cchafe would see the client reliably hang at waiting for peer. That said, a lot of the networking code has changed since this workaround was put in place. (All platforms now use kernel polling mechanisms for receiving packets, and there's no remaining code that blocks on the main thread.) So my guess is that it should be safe to remove. (Especially since it was uncertain how this worked to fix things on the platform to begin with.)

psiborg112:
Also, not entirely sure this fixes #648 in its entirely since there seem to be a few different issues going on there. (Looks like it probably fixes the original issue, but not the one then reported by @nwang92 where incorrect values are being read from the header.)

PR #823, opened 2022-12-02
Fix Window Resize Rendering

dhing1024:
Should take care of https://app.asana.com/0/1200778084029081/1203387632317029/f

I'm not sure if it makes the issue completely go away but at least on my Windows machine the rendering performs much better on resize. (I also don't really understand what Qt is doing internally here to know why this makes a difference).

I don't think this is an urgent enough fix to warrant trying to squeeze it into 1.6.7. It can wait until after the current release is done.

### DIFF
--- a/src/gui/virtualstudio.cpp
+++ b/src/gui/virtualstudio.cpp
@@ -199,6 +199,7 @@ VirtualStudio::VirtualStudio(bool firstRun, QObject* parent)
     m_view.setSource(QUrl(QStringLiteral("qrc:/vs/vs.qml")));
     m_view.setMinimumSize(QSize(594, 519));
     // m_view.setMaximumSize(QSize(696, 577));
+    m_view.setResizeMode(QQuickView::SizeRootObjectToView);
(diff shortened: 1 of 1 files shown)


ntonnaett:
Could you please give a description of the issue. Asana is not accessible for everyone.

dhing1024:
> Could you please give a description of the issue. Asana is not accessible for everyone.

From Asana: resizing windows shows elements on other screens and is extremely choppy/stuttery
only actually resizes like 50% of the time

I've never seen rendering issues on my Mac, only on my Windows machine. Here is also a screen capture of the original issue:

https://user-images.githubusercontent.com/26987971/205385446-e3701c7c-374c-40d7-a4e5-2ca81db981d9.mp4

ntonnaett:
That's funky. It isn't reproducible for Linux with X11 and Wayland either. But… there are shadows on wayland. I might be able to switch the backend for flatpaks. :)

It seems to work on all platforms and doesn't look like a dangerous thing to do. So I approve.

ntonnaett (approved):

nwang92 (approved):

PR #827, opened 2022-12-02
Fix for Newly Subscribed Studios Showing Up Under "Public Studios"

dhing1024:
Prior to this change, if a new subscriptions was created for a (private) studio and the "Refresh" button was clicked in-app, the new studio would show up under the "Public Studios" list. This was due to the `VirtualStudio::refreshStudios` function updating the list of studios but not the list of subscriptions, so the app would (incorrectly) by default categorize the newly subscribed studio as public.

These changes cause the subscriptions to be updated. Additionally, calling `getServerList` was a side effect of `getSubscriptions`. However, I think it's better for each function to just do 1 thing without side effects, so I decoupled the two.

### DIFF
--- a/src/gui/virtualstudio.cpp
+++ b/src/gui/virtualstudio.cpp
@@ -825,6 +825,7 @@ void VirtualStudio::logout()
 
 void VirtualStudio::refreshStudios(int index, bool signalRefresh)
 {
+    getSubscriptions();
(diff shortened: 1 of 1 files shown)


nwang92:
Works for me!

nwang92 (approved):

PR #834, opened 2022-12-06
Bugfix/commentwarning

ntonnaett:
This fixes a comment warning that should have been an error. I'm wondering what implications this typo has. Is the reverb working right now?

To catch those kind of bugs in the future, I made warnings errors on the gha Linux Meson builds.


psiborg112 (approved):

psiborg112:
Good catch

PR #837, opened 2022-12-06
Include git commit hash in display of version

mikedickey:
We are currently calling things "release candidates" which technically are not that, because if nothing else we have to keep changing the version string to represent if it's an RC or not (and which RC it is). A release candidate should not change in any way prior to it being released as final.

We recently ran into a critical problem that was introduced in the last few commits that were made after 1.6.7 rc2 and before the actual 1.6.7 was tagged, which required doing a fire drill to release 1.6.8. We have a manual pre-release process and test case that would have caught this before release, had we actually been testing the final release packages. Instead we tested "rc2" and then changed it.

I propose that release candidates are the final bits that we ship, no approximations that may change in the last minutes. This requires that we not make any manual commits to change the version strings for release candidates. However, it's still necessary to determine if build A is the final release versus build B, and the standard way to do this is using git commit hashes.

This PR changes the build scripts so that the version string is always (automatically) appended with the git commit hash that was used to build it. Whether any particular build is "rc1" or "rc2" etc would simply be determined by a mapping of commit hashes. i.e. "1.6.9 14f4d619" may be rc1 which "1.6.9 ca16e8bb" may be rc2. The process of making something final would simply be a matter of promoting rc2 to the final release (for example). This could further be represented using tags.

This will help us conform to more standard release process conventions, and avoid releases from being broken and published untested with last minute changes.

### DIFF
--- a/build
+++ b/build
@@ -4,7 +4,8 @@
 # Parse command line options
 clean=1
 install=0
-CONFIG=
+HASH=`git rev-parse --short HEAD`
+CONFIG="DEFINES+=JACKTRIP_COMMIT_HASH=\\\"$HASH\\\""
 UNKNOWN_OPTIONS=
 BUILD_RTAUDIO=0
 RTAUDIO=0
--- a/src/Settings.cpp
+++ b/src/Settings.cpp
@@ -63,6 +63,14 @@
 #include <unistd.h>
 #endif
 
+#ifdef JACKTRIP_COMMIT_HASH
+#define STR(s)       #s
+#define TO_STRING(s) STR(s)
+#define LONG_VERSION gVersion << "-" << TO_STRING(JACKTRIP_COMMIT_HASH)
+#else
+#define LONG_VERSION gVersion
+#endif
+
 #ifdef RT_AUDIO
 #include "RtAudioInterface.h"
 #endif
@@ -415,8 +423,8 @@ void Settings::parseInput(int argc, char** argv)
             break;
         case 'v':
             //-------------------------------------------------------
-            cout << "JackTrip VERSION: " << gVersion << endl;
-            cout << "Copyright (c) 2008-2021 Juan-Pablo Caceres, Chris Chafe." << endl;
+            cout << "JackTrip VERSION: " << LONG_VERSION << endl;
+            cout << "Copyright (c) 2008-2022 Juan-Pablo Caceres, Chris Chafe." << endl;
             cout << "SoundWIRE group at CCRMA, Stanford University" << endl;
 #ifdef QT_OPENSOURCE
             cout << "This build of JackTrip is subject to LGPL license." << endl;
@@ -688,14 +696,14 @@ void Settings::printUsage()
     cout << "" << endl;
     cout << "JackTrip: A System for High-Quality Audio Network Performance" << endl;
     cout << "over the Internet" << endl;
-    cout << "Copyright (c) 2008-2021 Juan-Pablo Caceres, Chris Chafe." << endl;
+    cout << "Copyright (c) 2008-2022 Juan-Pablo Caceres, Chris Chafe." << endl;
     cout << "SoundWIRE group at CCRMA, Stanford University" << endl;
 #ifdef QT_OPENSOURCE
     cout << "This build of JackTrip is subject to LGPL license." << endl;
 #endif
     cout << "JackTrip source code is released under MIT and GPL licenses." << endl;
     cout << "See LICENSE.md file for more information." << endl;
-    cout << "VERSION: " << gVersion << endl;
+    cout << "VERSION: " << LONG_VERSION << endl;
     cout << "" << endl;
     cout << "Usage: jacktrip [-s|-c|-S|-C hostIPAddressOrURL] [options]" << endl;
     cout << "" << endl;


ntonnaett:
I would prefer `git describe --tags` for this. 
Current output:
main: v1.6.8
dev: v1.6.8-7-g32eb9f5c

dev is 7 commits beyond main. I'm not sure about the hash right now. If we remove the v from our tags, we don't even have to parse anything. In meson I can use this directly with vcs_tag: https://mesonbuild.com/Reference-manual_functions.html#vcs_tag

ntonnaett:
git describe puts a g in front of the commit hash. https://git-scm.com/docs/git-describe

mikedickey:
> I would prefer `git describe --tags` for this...

I don't believe the "7 commits behind main" adds any value in this context, and may also be an extra changing variable that requires rebuilds and defeats the purpose. But I'd be OK changing it to have "-" between the version number and hash.

ntonnaett:
What I like about the git describe solution is that releases and prereleases are just their version number. It's also just one command. No manual setting of gVersion.

mikedickey:
> What I like about the git describe solution is that releases and prereleases are just their version number. It's also just one command. No manual setting of gVersion.

I like the idea of being able to get rid of gVersion completely (if that is what you propose), but what would be the purpose of the `-7-g` part of it, other than making it harder to parse and use? Also if a build were to be made off main, it would be a sort of race condition. I think we would want to ensure the version string always fits a specific, controllable format, whatever that may be.

I could go either way on whether we prepend `v` or not. I believe both approaches are equally common.

How about something like this?

```
LONG_VERSION=$(git describe --tags | sed 's/\(v[0-9]*\.[0-9]*\.[0-9]*\).*/\1/')-$(git rev-parse --short HEAD)
echo $LONG_VERSION
v1.6.8-b9e13736
```

ntonnaett:
And although the last rc and final versions might be functionally the same, we have to update the changelog.yml. OTOH we could not mention RCs in the changelog, and just append to the future version number.

I also don't like that I can't see immediately if a build is a rc, a final version or something else if we only use the commit hash.

We also have to think about the organisation of our GHAs. IIRC a rebuild is triggered when pushing to main.

I totally welcome making RCs real release candidates. Don't understand me wrong.

ntonnaett:
I propose generating the version string from the build system with just `git describe --tags`. There's no parsing involved.

What race condition do you have in mind? If two versions have the same commit hash?

mikedickey:
> I propose generating the version string from the build system with just `git describe --tags`. There's no parsing involved.
> 
> What race condition do you have in mind? If two versions have the same commit hash?

As you pointed out, it is non-deterministic if running from main gives you a different output format. The only way to ensure that the build you have is a specific commit is by including the hash, so it would kind of defeat the purpose. And likely create problems downstream. Not to mention it would be extremely unconventional, versus the formats I proposed being quite standard practice for mature engineering organizations.

ntonnaett:
The branch makes no difference. Both tags of last rc and final version would point to the same commit hash. Commit hashes are unique over all branches. Using hashes in version numbers is uncommon. That's the reason the windows build fails right now. Why should I care about mature engineering organizations? Authority is a bad argument. Especially if the version identifier regresses its usefulness.

mikedickey:
> The branch makes no difference. Both tags of last rc and final version would point to the same commit hash. Commit hashes are unique over all branches. Using hashes in version numbers is uncommon. That's the reason the windows build fails right now. Why should I care about mature engineering organizations? Authority is a bad argument. Especially if the version identifier regresses its usefulness.

Maturity != Authority

Conventions emerge when people gain experience from working over longer periods of timing, and learning what works vs what doesn't.

psiborg112:
I'm not loving that as an end user, I can't quickly and easily see if something is a pre-release or not. Having rc in a version number means something to me. (And it's why plenty of other software projects seem to have no trouble using it.) A git commit hash means nothing.

(Ultimately though, I'm maintaining my own builds which include a date based build number, so I probably don't really have a horse in this race.)

nwang92:
Some major issues:
* my `meson` build doesn't seem to show the git hash, maybe the change needs to go somewhere else?
* I feel like the auto-updater won't like this, ex: is `1.6.8-9e847b4` is newer than `1.6.8-b9e1373`?. We can change everything about how the auto-updater works but ultimately the reported version names may need something that can be sortable

Minor concerns:
* the area where we show version may need to be moved due to length:
Screenshot 2022-12-06 at 6 59 19 PM
* personal preference: I like reporting git hash as a separate property, but that may not address any of the concern(s) here

```
$ docker --version
Docker version 20.10.21, build baeda1f
$ minikube version
minikube version: v1.28.0
commit: 986b1ebd987211ed16f8cc10aed7d2c42fc8392f
$ kubectl version --output=json
{
  "clientVersion": {
    "major": "1",
    "minor": "25",
    "gitVersion": "v1.25.4",
    "gitCommit": "872a965c6c6526caa949f0c6ac028ef7aff3fb78",
    "gitTreeState": "clean",
    "buildDate": "2022-11-09T13:28:30Z",
    "goVersion": "go1.19.3",
    "compiler": "gc",
    "platform": "darwin/arm64"
  },
  ...
```

mikedickey:
Here is an alternative approach https://github.com/jacktrip/jacktrip/pull/838

psiborg112:
The more I look at the alternatives being proposed here, the more I feel we're looking for a technical solution for what should be a matter of process. Semantic versioning, which we're currently using with the whole rc thing, addresses all of the issues that are being raised by @ntonnaett and @nwang92. Plenty of other software projects manage to use it without issue. Why reinvent the wheel?

dyfer:
I also thing that the most immediate solution is to fix the process. Why not make RC releases true RCs? I.e. if there are bugs, fix bugs between RCs, and then not add any functional changes between the final RC and the release.

If we want more information about the build, I'd consider adding a separate line in version identifier (a little bit like #838) with branch and commit hash (or tag and commit hash for tagged builds), which would also differentiate releases from non-release builds, but keep the version number in the source like we do currently and fix the issues about RC by adhering to the process.

psiborg112:
What @dyfer said. Once we release an rc we freeze dev. Commits only get made to fix bugs that arise and we release this as a new rc. Then the only thing that changes between the final rc and the release is the version change. (New work can continue on other branches which can be merged post release.)

PR #838, opened 2022-12-07
Replace gVersion with auto-generated buildtime constants

mikedickey:
We are currently calling things "release candidates" which technically are not that, because if nothing else we have to keep changing the version string to represent if it's an RC or not (and which RC it is). A release candidate should not change in any way prior to it being released as final.

We recently ran into a critical problem that was introduced in the last few commits that were made after 1.6.7 rc2 and before the actual 1.6.7 was tagged, which required doing a fire drill to release 1.6.8. We have a manual pre-release process and test case that would have caught this before release, had we actually been testing the final release packages. Instead we tested "rc2" and then changed it.

I propose that release candidates are the final bits that we ship, no approximations that may change in the last minutes. This requires that we not make any manual commits to change the version strings for release candidates. However, it's still necessary to determine if build A is the final release versus build B, and the standard way to do this is using git commit hashes.

This PR eliminates the need to manually maintain the `gVersion` field, replacing it with new `gShortVersion` (semver) and `gLongVersion` (`git describe --tags`) variables that are set at build time. It also introduces two new build time variables: `gCommitHash` and `gBuildTime` for more detailed information about the build.

This modifies the behavior of `-v` so that the "JackTrip VERSION" line will always display semver (x.y.z), which makes it more consistent and easier for downstream things to parse. But it adds a "Build Info" line to display the more detailed information, which is required for anyone testing or debugging a build to know what code was used to generate it. 

This will help us conform to more standard release process conventions, and avoid releases from being broken and published untested with last minute changes.

### DIFF
--- a/build
+++ b/build
@@ -4,7 +4,11 @@
 # Parse command line options
 clean=1
 install=0
-CONFIG=
+BUILD_TIME=$(date -u +"%Y-%m-%d-%H-%M-%S")
+COMMIT_HASH=$(git rev-parse --short HEAD)
+LONG_VERSION=$(git describe --tags)
+SHORT_VERSION=$(echo $LONG_VERSION | sed 's/v\([0-9]*\.[0-9]*\.[0-9]*\).*/\1/')
+CONFIG="DEFINES+=JACKTRIP_BUILD_TIME=\\\"$BUILD_TIME\\\"  DEFINES+=JACKTRIP_COMMIT_HASH=\\\"$COMMIT_HASH\\\" DEFINES+=JACKTRIP_SHORT_VERSION=\\\"$SHORT_VERSION\\\" DEFINES+=JACKTRIP_LONG_VERSION=\\\"$LONG_VERSION\\\""
 UNKNOWN_OPTIONS=
 BUILD_RTAUDIO=0
 RTAUDIO=0
--- a/src/Settings.cpp
+++ b/src/Settings.cpp
@@ -415,8 +415,10 @@ void Settings::parseInput(int argc, char** argv)
             break;
         case 'v':
             //-------------------------------------------------------
-            cout << "JackTrip VERSION: " << gVersion << endl;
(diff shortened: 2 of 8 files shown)


mikedickey:
```
$ jacktrip -v
JackTrip VERSION: 1.6.8
Build Info: v1.6.8-8-g7473b2f8 (7473b2f8) @ 2022-12-07-04-13-43
Copyright (c) 2008-2022 Juan-Pablo Caceres, Chris Chafe.
SoundWIRE group at CCRMA, Stanford University
This build of JackTrip is subject to LGPL license.
JackTrip source code is released under MIT and GPL licenses.
See LICENSE.md file for more information.
```

psiborg112:
#837 is definitely much less intrusive and doesn't create the problem of using 0.0.0 as a default version number. (Which is what is going to be currently happening here for cmake builds. And maybe meson builds, @ntonnaett?)

psiborg112:
Also, if someone downloads the code as a zip file, this happens:
`DEFINES+=JACKTRIP_BUILD_TIME=\"2022-12-07-06-38-20\"  DEFINES+=JACKTRIP_COMMIT_HASH=\"\" DEFINES+=JACKTRIP_SHORT_VERSION=\"\" DEFINES+=JACKTRIP_LONG_VERSION=\"\"`

psiborg112:
Additionally, I think having the time included rather than just the date isn't super useful, because it's going to end up being slightly different for each platform. It's why my build string is of the form 'yyyymmddnn' where n is currently manually set in the very rare circumstance that I've had to make a second release on the same day. So the latest build is '2022111800' for example. (And the real reason for its existence is simply so that I've been able to make multiple releases per version where needed without being out of step with the official version number.)

If you're still using the github commit number then adding the date seems overkill. And most importantly #837 doesn't totally break things when the commit hash is unset.

dyfer:
I also think that this solution is problematic since it needs git information in order to display any proper version. I don't think removing version from the source completely is viable.
Since most conversation about the general issue happened in #837, I'll continue there.

PR #842, opened 2022-12-09
Start and Join Inactive Studios

dhing1024:
Work-in-progress for starting and joining inactive studios.

- Makes the "Join" button a "Start" button that moves you to the "Connected" screen and opens the studio link in the web app. Planning to add a query param `?start=true` to automatically open the Start Studio dialog in the web app.
- Do we like the way this icon looks? Seems a little small to me.
- Currently trying to establish a proof-of-concept that we can use websocket connections with the API server to determine the studio state and connect to the new studio automatically when it is ready. Currently the connection succeeds if the studio was already up and you click "Join" but fails if you click "Start", start the studio, and wait for it to become ready. Seeing these messages, and I'm not sure why: `JackTrip Server Timed Out!` and `Initial TCP Connection Timed Out` even though the host IP and port are available. (May be a blocker for more work on this feature)
- It looks like there was previous work for starting studios automatically. I think most of that code is going to be removed and replaced by whatever implementation comes of this work here.

Still has bugs and a number of different scenarios I haven't looked at yet, right now just trying to get a proof of concept working.


dhing1024:
More learnings: when you click the "Start" button to open Virtual Studio, the call to `completeConnection` ([here](https://github.com/jacktrip/jacktrip/blob/7aaadece6f048ad81496d3f5cd1b2513067e4411/src/gui/virtualstudio.cpp#L1349)) ends up with the TCP connection socket giving a `QAbstractSocket::ConnectionRefusedError`.  If your studio is already active and you press the "Join" button, this doesn't happen. I thought maybe adding a short delay might help, but that seems to not make a difference. I have no idea why there's a difference in behavior between the two ways of connecting - I think everything else is the same.

dhing1024:
Fixed the connection refused error using an exponential backoff retry. Also added a waiting mode to the connected screen that updates automatically using websockets. Once a studio has started it will automatically connect the user through JackTrip. This includes studios for which the user isn't an owner/admin.


https://user-images.githubusercontent.com/26987971/207674176-b6d74d80-4fc4-4686-8bd1-359f72998980.mov


https://user-images.githubusercontent.com/26987971/207674214-7903b5d6-3597-4844-be78-fe64c3e016d4.mov

nwang92:
Some things I found:
* If I click "Leave" on the "Waiting" screen after starting the studio, I go back to the list view but shortly after I'm automatically connected 
* I tried to "Start" a studio that was decommissioning - the "Start Studio" dialog didn't appear but I got this error from the app
Screenshot 2022-12-14 at 11 23 09 AM

dhing1024:
I think these are both fixed now

nwang92 (approved):

mikedickey (approved):

PR #844, opened 2022-12-11
Migrate signing to the main workflow file

dyfer:
I think it would be better for maintainability if we didn't have separate workflows running on JackTrip Labs repo (with signing) and on the main repo.

This PR migrates signing mechanism to the main workflow file. The signing will only happen if the secrets are available, so the same workflow should result in signed builds in JTL and non-signed builds in the regular repo.

**NOTE: this needs to be tested on the JackTrip Labs repo before merging.** One way to do this would be to push this branch to the JTL repo, and then use "workflow_dispatch" trigger to manually trigger the build on that branch. To be extra safe, it might be prudent to also create a temporary release on the JTL repo (e.g. tag if `test-1.6.8-gha`, then delete the release/tag after checking if the signed builds were uploaded), but that's not strictly necessary.

This PR also adds minor maintenance to GHA, replacing `set-output` functionality and updating some actions, which should remove the warnings that are posted currently.


mattahorton:
Amazing. I'll absolutely do the release test over on the JTL repo and come back to approve this if it works

mattahorton:
I don't have time to look at this at the moment, but it failed because it couldn't find certain files on the assemble step for macOS and on the signing step for Windows.

dyfer:
Thanks so much for the review @mattahorton !

I think I found the issue with the macOS build - the order of flags was wrong for the assemble_app script.

As for the Windows signing... there's some stuff going on there - I haven't tried to run that java signing tool that you're downloading and running to do the signing. I saw that it was originally running on Linux... is there any documentation on that tool? If we really need to, we could run it as a separate job in Linux, but it would be cleaner if we could run it on Windows as well.

mattahorton:
Oh are we no longer cross-compiling? That's a big change to include here.

Honestly, I spent a few days getting that signing tool to work and don't remember much about it. It's from eSigner/ssl.com. 

Here's the documentation. https://www.ssl.com/guide/esigner-codesigntool-command-guide/

The linux version is stored in cloud storage and we pull it from there. I have no experience working with the windows version.

dyfer:
> Oh are we no longer cross-compiling? That's a big change to include here.

No sure what you mean. I can see that you were signing the Windows build under Linux, but we were never cross-compiling Jacktrip in GHA, as far as I remember.

> Honestly, I spent a few days getting that signing tool to work and don't remember much about it. It's from eSigner/ssl.com.
> 
> Here's the documentation. https://www.ssl.com/guide/esigner-codesigntool-command-guide/

Thanks! That's very helpful.

Could you put the Windows version (linked on that page) in the place you'd like it to pulled from for signing? Then I would adjust the workflow and you'd be able to test it again on JTL repo, if that works for you.

mattahorton:
@dyfer https://files.jacktrip.org/contrib/CodeSignTool-v1.2.6-windows.zip

here you go

dyfer:
@mattahorton I pushed something that _might_ work. I'm almost certain I messed up either slash direction, or env variable handling, or something else... but if you could try it on the JTL repo that would be great. Thanks!

mattahorton:
This worked! Approving

mattahorton (approved):

dyfer:
> This worked! Approving

Thanks! 
Just to be clear, have you confirmed that the installer is signed? I can see [this](https://github.com/jacktriplabs/jacktrip/actions/runs/3707835786/jobs/6284722860#step:24:636) in the logs and I'm not sure where it came from.

Issue #845, opened 2022-12-12
JackTrip (VS) without RtAudio is broken


dyfer:
Jacktrip segfaults (at least on macOS and Linux - Ubuntu 20.04) when built without RtAudio. Tested with local build from the latest `dev`, but I think this might have been happening for a while (this might be technically a duplicate of #731, but I'm not sure). **This seems to be related to the Virtual Studio gui**, since on macOS the app did originally start properly using the classic gui.

IIUC this effectively means that the recent versions of JackTrip don't work on Linux with the GUI, where it is only used with JACK.

On Linux (local build), the segfault happens before the GUI shows up. On macOS, I get through Virtual Studio login, then the "JackTrip needs your sounds", and then, after clicking OK on that screen, it segfaults:

terminal output:
<details>

```
% builddir/jacktrip
qrc:/vs/vs.qml:126:5: QML Connections: Implicitly defined onFoo properties in Connections are deprecated. Use this syntax instead: function onFoo(<arguments>) { ... }
qrc:/vs/Setup.qml:632:17: Unable to assign [undefined] to QString
qrc:/vs/Setup.qml:741:17: Unable to assign [undefined] to QString
qrc:/vs/Settings.qml:311:17: Unable to assign [undefined] to QString
qrc:/vs/Settings.qml:394:17: Unable to assign [undefined] to QString
SSL Library: Secure Transport, macOS 12.6
Log file location: "/Users/marcin/Library/Application Support/jacktrip/JackTrip/log.txt"
The Sampling Rate is: 48000
---------------------------------------------------------
The Audio Buffer Size is: 128 samples
                      or: 512 bytes
---------------------------------------------------------
zsh: segmentation fault  builddir/jacktrip
```
</details>

crash report on macOS:
<details>

```
-------------------------------------
Translated Report (Full Report Below)
-------------------------------------

Process:               jacktrip [43472]
Path:                  /Volumes/VOLUME/*/jacktrip
Identifier:            jacktrip
Version:               ???
Code Type:             ARM-64 (Native)
Parent Process:        zsh [34961]
User ID:               501

Date/Time:             2022-12-12 10:09:21.7732 +0100
OS Version:            macOS 12.6.1 (21G217)
Report Version:        12
Anonymous UUID:        55069327-BA77-335A-8601-5EFC0C07577B

Sleep/Wake UUID:       E9F028AE-90A0-4CE2-9AE3-0E11CE6AEFBA

[...]

PR #847, opened 2022-12-12
Catching a few cases where neither JACK not RtAudio are available

mattahorton:
This is a bit of ifdef hell, but I'd love a review. 

There's more we can do, but I think one of the things going on with some of our segfaults being reported is either JACK or RtAudio being set without the corresponding necessary definitions.

### DIFF
--- a/src/gui/vsAudioInterface.cpp
+++ b/src/gui/vsAudioInterface.cpp
@@ -65,10 +65,10 @@ VsAudioInterface::VsAudioInterface(int NumChansIn, int NumChansOut,
 {
     QSettings settings;
     settings.beginGroup(QStringLiteral("Audio"));
-    m_inMultiplier       = settings.value(QStringLiteral("InMultiplier"), 1).toFloat();
-    m_outMultiplier      = settings.value(QStringLiteral("OutMultiplier"), 1).toFloat();
-    m_inMuted            = settings.value(QStringLiteral("InMuted"), false).toBool();
-    m_outMuted           = settings.value(QStringLiteral("OutMuted"), false).toBool();
+    m_inMultiplier  = settings.value(QStringLiteral("InMultiplier"), 1).toFloat();
+    m_outMultiplier = settings.value(QStringLiteral("OutMultiplier"), 1).toFloat();
+    m_inMuted       = settings.value(QStringLiteral("InMuted"), false).toBool();
+    m_outMuted      = settings.value(QStringLiteral("OutMuted"), false).toBool();
 #ifdef RT_AUDIO
 #ifndef NO_JACK
     m_audioInterfaceMode = (settings.value(QStringLiteral("Backend"), 0).toInt() == 1)
--- a/src/gui/vsAudioInterface.cpp
+++ b/src/gui/vsAudioInterface.cpp
@@ -57,7 +57,11 @@ VsAudioInterface::VsAudioInterface(int NumChansIn, int NumChansOut,
     , m_inputDeviceName("")
     , m_outputDeviceName("")
     , m_audioBufferSize(gDefaultBufferSizeInSamples)
+#ifdef RT_AUDIO
     , m_audioInterfaceMode(VsAudioInterface::RTAUDIO)
+#else
+    , m_audioInterfaceMode(VsAudioInterface::JACK)
+#endif
 {
     QSettings settings;
     settings.beginGroup(QStringLiteral("Audio"));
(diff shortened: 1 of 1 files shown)


dyfer:
Not a full review, but after building this on macOS without RtAudio there were no segfaults and the app started.

Not sure if that's relevant - the first run after building did _not_ start, emitting this:
<details>

``` sh
% builddir/jacktrip
qrc:/vs/vs.qml:126:5: QML Connections: Implicitly defined onFoo properties in Connections are deprecated. Use this syntax instead: function onFoo(<arguments>) { ... }
qrc:/vs/Setup.qml:632:17: Unable to assign [undefined] to QString
qrc:/vs/Setup.qml:741:17: Unable to assign [undefined] to QString
qrc:/vs/Settings.qml:311:17: Unable to assign [undefined] to QString
qrc:/vs/Settings.qml:394:17: Unable to assign [undefined] to QString
SSL Library: Secure Transport, macOS 12.6
Log file location: "/Users/marcin/Library/Application Support/jacktrip/JackTrip/log.txt"
^C # this is where I stopped it
```
</details>

Then I started Jack and the subsequent starts were fine. Not sure if that's relevant as I'm not able to reproduce this.

dyfer (commented):
I am not very familiar with the implementation, but I was wondering if maybe it would be possible to add a function to `VsAudioInterface` like `getDefaultBackend` or something like that, where we'd do the `#ifdef` dance around `RT_AUDIO` and `NO_JACK`, which could replace at least ifdefs [here](https://github.com/jacktrip/jacktrip/pull/847/files#diff-d0c60d9b6f990a02918f8d909d0799c60c9883df2965760c8bb2f06d929130f2R77-R84) and [here](https://github.com/jacktrip/jacktrip/pull/847/files#diff-d0c60d9b6f990a02918f8d909d0799c60c9883df2965760c8bb2f06d929130f2R77-R84), what do you think?

ntonnaett:
Just for my understanding. This is for the case that you chose a backend that was supported by a previous build and was saved, but isn't supported by the currently used build?! Do I miss other circumstances where this could happen?

IMHO we should get rid of macro hell and move to `if constexpr`. The benefit of `if constexpr` is that all branches of the code are checked by the compiler. It would require to also switch to C++17 for the qmake build. IIRC we actually already have C++17 code in jacktrip. 

The changes right now look good to me. Didn't test. If this fixes #845, we shouldn't wait too long with the merge. Afterwards we could work on a more beautiful overhaul maybe with `bool getDefaultBackend()` and `constexpr bool isBackendAvailable(API)`.

ntonnaett:
I just build with NO_JACK and the audio backend selection includes JACK as an option. Is this expected behavior right now?

mattahorton:
> I just build with NO_JACK and the audio backend selection includes JACK as an option. Is this expected behavior right now?

No

mattahorton:
> I just build with NO_JACK and the audio backend selection includes JACK as an option. Is this expected behavior right now?

No

> Just for my understanding. This is for the case that you chose a backend that was supported by a previous build and was saved, but isn't supported by the currently used build?! Do I miss other circumstances where this could happen?
> 
> IMHO we should get rid of macro hell and move to `if constexpr`. The benefit of `if constexpr` is that all branches of the code are checked by the compiler. It would require to also switch to C++17 for the qmake build. IIRC we actually already have C++17 code in jacktrip.
> 
> The changes right now look good to me. Didn't test. If this fixes #845, we shouldn't wait too long with the merge. Afterwards we could work on a more beautiful overhaul maybe with `bool getDefaultBackend()` and `constexpr bool isBackendAvailable(API)`.

This

> Just for my understanding. This is for the case that you chose a backend that was supported by a previous build and was saved, but isn't supported by the currently used build?! Do I miss other circumstances where this could happen?
> 
> IMHO we should get rid of macro hell and move to `if constexpr`. The benefit of `if constexpr` is that all branches of the code are checked by the compiler. It would require to also switch to C++17 for the qmake build. IIRC we actually already have C++17 code in jacktrip.
> 
> The changes right now look good to me. Didn't test. If this fixes #845, we shouldn't wait too long with the merge. Afterwards we could work on a more beautiful overhaul maybe with `bool getDefaultBackend()` and `constexpr bool isBackendAvailable(API)`.

This is definitely the next thing I want to do.

mattahorton:
> 

@dyfer your experience answers the question @ntonnaett asked. I think this only happens if you run a build now that was unsupported in the last build you ran. 

@cchafe was having this issue when building from source and that seems to match his experience too.

ntonnaett:
> > I just build with NO_JACK and the audio backend selection includes JACK as an option. Is this expected behavior right now?
> 
> No

But at least it doesn't seem to be a regression.

ntonnaett (approved):

mikedickey:
I think this is definitely better, but ideally it should handle a few more cases:

* `NO_JACK` defined and `RT_AUDIO` not defined => compile time error/failure
* calling `VsAudioInterface::setAudioInterfaceMode` at runtime with a value not supported by compile time definitions => error or ignore

ntonnaett:
> 

The first point should be dealt with in the build system. I'll create a PR for meson.

PR #849, opened 2022-12-13
Address segfault due to ifdef hell with a move to c++17

mattahorton:
CleanShot 2022-12-16 at 12 01 24@2x

This **_begins_** work on moving from `#ifdef` of `RT_AUDIO` and `NO_JACK` to `if constexpr`. It uses that pattern to simplify some of the audio setup code and eliminate a couple of instances where VS mode could segfault.

It _does not_ bring that simplification to the underlying `AudioInterface`-based classes yet.

### DIFF
--- a/src/gui/AudioInterfaceMode.h
+++ b/src/gui/AudioInterfaceMode.h
@@ -38,13 +38,13 @@
 enum class AudioInterfaceMode {
     JACK,     ///< Jack Mode
     RTAUDIO,  ///< RtAudio Mode
-    BOTH,
+    ALL,
     NONE
 };
 
 #ifdef RT_AUDIO
 #ifndef NO_JACK
-constexpr AudioInterfaceMode mode = AudioInterfaceMode::BOTH;
+constexpr AudioInterfaceMode mode = AudioInterfaceMode::ALL;
 #else
 constexpr AudioInterfaceMode mode = AudioInterfaceMode::RTAUDIO;
 #endif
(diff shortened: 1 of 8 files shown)


ntonnaett:
OMG, it's happening! I genuinely hate this part of the code and it already looks so much better. I'm wondering if we should already think about what happens when we add more audio backends like Pipewire and maybe use native audio APIs of Windows and macOS directly. Then AudioInterfaceMode::BOTH won't be clear anymore.

src/gui/AudioInterfaceMode.h
```diff
@@ -0,0 +1,83 @@
...
+enum class AudioInterfaceMode {
+    JACK,     ///< Jack Mode
+    RTAUDIO,  ///< RtAudio Mode
+    BOTH,
+    NONE
+};
+
+#ifdef RT_AUDIO
+#ifndef NO_JACK
+constexpr AudioInterfaceMode mode = AudioInterfaceMode::BOTH;
```
ntonnaett, line 47: I know that there are (two?) places where we use have_libjack() for the weakjack case. Maybe it would be good to put that as a runtime check here. AFAIK that wouldn't contradict constexpr. But I don't speak constexpr fluently.

dhing1024 (approved):
I think this looks good, from reading through these changes and building it with both backends disabled. (I'm not an expert though if switching from C++11 to C++17 will have unknown adverse effects).

Re:

> I'm wondering if we should already think about what happens when we add more audio backends like Pipewire and maybe use native audio APIs of Windows and macOS directly. Then AudioInterfaceMode::BOTH won't be clear anymore.

It's probably fine for now as-is. I think that making this codebase extendable to arbitrary numbers of backends might require a lot more review of existing code and modifications to our build systems anyways. But definitely a good thing to be thinking about.

src/gui/AudioInterfaceMode.h
mattahorton: I dooon't quite understand that idea, but I did run into some cases where I still needed preprocessor directives as things were not building any more.

mattahorton:
> OMG, it's happening! I genuinely hate this part of the code and it already looks so much better. I'm wondering if we should already think about what happens when we add more audio backends like Pipewire and maybe use native audio APIs of Windows and macOS directly. Then AudioInterfaceMode::BOTH won't be clear anymore.

I'm happy to replace all instances of `BOTH` with `ALL` right now and then we can adapt things later?

src/gui/AudioInterfaceMode.h
ntonnaett: When using weakjack we don't know the availability of jack at compile time. But it would make sense to have this checked with isBackendAvailable(). AFAIU constexpr evaluates expressions at compile time as long as it works. So for weakjack we would introduce a runtime check for jack but everything else would be evaluated at compile time nonetheless.

ntonnaett:
> I think this looks good, from reading through these changes and building it with both backends disabled. (I'm not an expert though if switching from C++11 to C++17 will have unknown adverse effects).
> 

It would probably make JackTrip faster. For compatibility considerations we might want to think about old Linux distributions where JackTrip should be buildable, but not too much. C++17 is hopefully even supported on Ubuntu 18.04. 
The meson build switched to C++17 a long time ago and is what most/all Linux distributions use.

PR #850, opened 2022-12-14
Support changing output volume via virtual studio API

mikedickey:
Bumping version to 1.6.9


dhing1024 (approved):
Tried it out, looks good

mattahorton (approved):

PR #857, opened 2022-12-18
Use c++17 standard in cmake builds

psiborg112:


ntonnaett:
I just assumed this would have been already the case. We should add the cmake build to our github actions and add `-Werror` (at least for Linux).

ntonnaett (approved):

PR #858, opened 2022-12-20
Create separate arguments for input and output devices

maingig:
Fix for Issue #495. Use separate arguments for input and output devices.

### DIFF
--- a/src/Settings.cpp
+++ b/src/Settings.cpp
@@ -89,7 +89,9 @@ enum JTLongOptIDS {
     OPT_NUMSEND,
     OPT_APPENDTHREADID,
     OPT_LISTDEVICES,
-    OPT_AUDIODEVICE
+    OPT_AUDIODEVICE,
+    OPT_AUDIOINPUTDEVICE,
+    OPT_AUDIOOUTPUTDEVICE
 };
 
 //*******************************************************************************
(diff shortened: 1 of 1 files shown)


ntonnaett:
Thanks for your PR! Looks straight forward. Is this tested with device name including a comma?

Could you please rebase on dev!

maingig:
I tested with these devices:
"Rogue Amoeba Software, Inc.: Loopback Audio 2"
"Rogue Amoeba Software, Inc.: Loopback Audio 1"

Looks like there was one format issue.  Do I need to fix that?

ntonnaett:
I'm happy with this. The naming of the flags is nice for when we get tab completion one day. Let's see what others think. @dyfer @psiborg112 @cchafe

psiborg112:
Looks good to me. Doesn't prevent people from doing things the current way if they're already in the habit of that (or have existing scripts they're using) while fixing the issue.

ntonnaett:
> Looks good to me. Doesn't prevent people from doing things the current way if they're already in the habit of that (or have existing scripts they're using) while fixing the issue.

Do you have an idea for this? Counting the commas? May be something for another PR.

psiborg112:
Oh no, I just meant that it doesn't stop the current option from working the way people are used to while adding the necessary options to fix the comma issue. I don't think there's a more elegant solution than that.

ntonnaett:
Yeah, I agree. But the case for more than one comma for --audiodevice is broken. So we could warn if it's more than 1 and tell the user to use the new flags.

ntonnaett (approved):

psiborg112:
Yes, a warning would probably be good in that circumstance - but definitely for a future PR.

maingig:
Who is responsible for doing the merge? I'm not sure what your workflow is.

PR #861, opened 2023-01-02
Update Manage Studio Button on Connected Screen to Launch Video in Browser

dhing1024:


mattahorton:
we're only changing it on the connected screen?

dhing1024:
It's in the ACs: https://app.asana.com/0/1202679299825895/1203557467327908/f

mattahorton:
We don't do video for self-hosted studios, right? @nwang92 is there anything we should do to address that?

nwang92 (approved):

PR #866, opened 2023-01-12
Reauthenticate After Sign Out

dhing1024:
Use this additional `prompt=login` parameter to force a reauthentication after a user has pressed "Sign Out". Before this, when a user signed out and pressed the sign in button again, they would immediately redirect back to the desktop app without needing to log in via browser again.

https://user-images.githubusercontent.com/26987971/211953821-a90fd8b6-a863-4227-bb75-d87126bda1aa.mov

### DIFF
--- a/src/gui/virtualstudio.cpp
+++ b/src/gui/virtualstudio.cpp
@@ -897,6 +897,7 @@ void VirtualStudio::toVirtualStudio()
                 (*parameters)[QStringLiteral("code")] = QUrl::fromPercentEncoding(code);
             } else if (stage == QAbstractOAuth2::Stage::RequestingAuthorization) {
                 parameters->insert(QStringLiteral("audience"), AUTH_AUDIENCE);
+                parameters->insert(QStringLiteral("prompt"), QStringLiteral("login"));
             }
             if (!parameters->contains("client_id")) {
                 parameters->insert("client_id", AUTH_CLIENT_ID);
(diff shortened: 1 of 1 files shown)


nwang92 (approved):

dhing1024:
Worth noting: this update forces users to re-authenticate when they press the sign-in, in a way it doesn't really address the root cause of the original issue. This is because pressing "Sign Out" doesn't clear the in-browser Auth0 session single sign-on cookie. So when they logged out of the desktop app and pressed "Sign In" again, they are sent to the browser, which already has their Auth0 login session, and they do not need to reauthenticate in order to get the API access token.

We can clear the Auth0 browser session when the user logs out of the desktop app by opening the logout URL (see [auth0 API docs](https://auth0.com/docs/api/authentication#logout)) in their browser on logout. Though if we wanted to do that as well in addition to what is already in this PR, that would have some potential downsides (additional security at the expense of some user friendliness):

- We have to open a browser window when the user presses "Log Out"
- We need to design and deploy a "logout" landing page (a page that might say, for example "you have been logged out"). I'm pretty sure we would be required to provide this page as an "Allowed Logout URL". (see [here](https://auth0.com/docs/authenticate/login/logout/redirect-users-after-logout))
- If we clear the user's Auth0 session layer, then they are logged out of the Virtual Studio and JackTrip Radio web apps as well (technically only if it's the same browser as the one opened by the desktop app, but that's almost always gonna be true)

I think it's probably not worth that extra effort, but I'd still like a second opinion anyways before merging this.

nwang92:
I'm cool with that behavior - I'd rather know explicitly what user I'm trying to sign-in with rather than rely on my browser session. I'm guessing this probably impacts people who change accounts frequently (like us) more than normal users out there so it seems like an acceptable trade-off.

nwang92 (approved):

PR #867, opened 2023-01-12
Use random bind port to avoid client collision

nwang92:
When in Virtual Studio mode, use a random bind port for the jacktrip client process.

This PR also fixes a bug with `checkIfPortIsBinded`. The `QHostAddress::Any` option is used to check if a port is bound on [both ipv4 + ipv6 interfaces](https://doc.qt.io/qt-5/qhostaddress.html#SpecialAddress-enum). So if the port of interest is only in use by ipv4, the exception will not be raised which allows the client to use the port despite if it's already in-use elsewhere. This is particularly problematic when its allowed to proceed, it kicks the older client out and lets the new client in. 

Also dumping this here because it was helpful to understand: https://learn.microsoft.com/en-us/windows/win32/winsock/using-so-reuseaddr-and-so-exclusiveaddruse

### DIFF
--- a/src/JackTrip.cpp
+++ b/src/JackTrip.cpp
@@ -467,11 +467,23 @@ void JackTrip::startProcess(
         // qDebug() << "before mJackTrip->startProcess" << mReceiverBindPort<<
         // mSenderBindPort;
 #endif
-    checkIfPortIsBinded(mReceiverBindPort);
+    if (checkIfPortIsBinded(mReceiverBindPort)) {
+        stop(QStringLiteral("Could not bind %1 UDP socket. It may already be binded by "
+                            "another process on "
+                            "your machine. Try using a different port number")
+                 .arg(mReceiverBindPort));
+        return;
+    }
     if (gVerboseFlag)
         std::cout << "  JackTrip:startProcess before checkIfPortIsBinded(mSenderBindPort)"
                   << std::endl;
-    checkIfPortIsBinded(mSenderBindPort);
+    if (checkIfPortIsBinded(mSenderBindPort)) {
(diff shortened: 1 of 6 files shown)


nwang92:
The port collision changes here work when running 2 jacktrip processes on the same machine, but it still fails in a pretty bad way when running across different machines.

I tried running this on 2 different computers across the same LAN connecting to a cloud hub server:
* launch jacktrip using bind port 4464 on computer A
* computer A connects successfully
* launch jacktrip using bind port 4464 on computer B
* computer B has nothing bound on 4464 so the check passes
* computer B connects successfully
* computer A disconnects with a "Studio stopped" exception

I changed some of the logic in this PR to always go with a randomized port but here's what I'm thinking:
# maybe we should use a wider range (I don't know if there's any history behind `gBindPortHigh` and `gBindPortLow`)
# IMO it's very problematic that computer B wins and kicks out computer A; it's way easier to handle port collisions when computer B hits the error
# I guess the proper way to handle this would be some sort of remote echo service that checks what jacktrip ports are in use by your public IP (something one level higher than https://github.com/jacktrip/jacktrip/blob/b97c6af137572834cbeb1e117d3ca15f53b80be3/src/gui/qjacktrip.cpp#L227-L228). Something similar to a the [WebRTC STUN](https://bloggeek.me/webrtcglossary/stun/) server

nwang92:
Addressed both comments. There's nothing of note to show when it works, but here's a video of what it looks like when it exceeds all the attempts to find an open port:

https://user-images.githubusercontent.com/6201822/214978670-efaa0a61-edcc-44d3-87a9-4f7637e95df6.mov

mikedickey (approved):

PR #874, opened 2023-01-19
Fixing TCP error upon connecting from desktop app start

nwang92:
This should fix the "TCP Socket Error 19". I believe the root cause of this was:
```
newInterval = randomizer.bounded(0, newInterval);
```
which would pick a random number from 0 to `{newInterval}` which governs the sleep period between retries. This has some non-deterministic effects because it could exhaust all retries but not sleep for the full period we need to.

This also fixes a bug where you can start a studio that is shutting down which may have some bad outcomes. I handled this case with a text blurb:
Screenshot 2023-01-19 at 9 21 46 AM

### DIFF
--- a/src/JackTrip.cpp
+++ b/src/JackTrip.cpp
@@ -1077,11 +1077,17 @@ void JackTrip::tcpTimerTick()
     // Use randomized exponential backoff to reconnect the TCP client
     QRandomGenerator randomizer;
     mRetries++;
+    // exponential backoff sleep with 6s maximum + jitter
     int newInterval = 2000 * pow(2, mRetries);
-    newInterval     = randomizer.bounded(0, newInterval);
+    newInterval     = std::min(newInterval, 6000);
+    newInterval += randomizer.bounded(0, 500);
+    QString now = QDateTime::currentDateTime().toString(Qt::ISODate);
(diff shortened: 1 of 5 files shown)


dhing1024:
Would you happen to know why this bug would only seemingly manifest on installed versions of the app, and not local builds from source?

nwang92:
> Would you happen to know why this bug would only seemingly manifest on installed versions of the app, and not local builds from source?

I didn't notice it was a installed version problem - I just tried the build from this branch and hit the same error so I guess it's not fixed. I'll keep digging.

nwang92:
Ok, I think this is fixed now in builds. I switched from `disconnectFromHost` to `abort` to avoid the asynchronous behavior. Docs: https://doc.qt.io/qt-6/qabstractsocket.html#abort

dhing1024 (approved):
Just tested installing on Windows and Mac, and they both seem to not result in TCP Socket Error 19 anymore. I guess it does make sense that asynchronously clearing the client would leave it in a state where it couldn't open a new connection right away, which aligns with the meaning of [TCP Error 19](https://doc.qt.io/qt-6/qabstractsocket.html#SocketError-enum). It's just weird that the issue only manifested on installed versions and not local builds.

PR #875, opened 2023-01-20
Automatically delete QNetworkReplies

nwang92:
I don't know how this works but by default this is false. Seems like a sane feature to turn on?


dhing1024:
Looks like some of these builds are using an older qt version (5.12) which doesn't yet have the `setAutoDeleteReplies` function: https://doc.qt.io/archives/qt-5.12/qnetworkaccessmanager.html

nwang92:
https://media.giphy.com/media/1BXa2alBjrCXC/giphy.gif

PR #883, opened 2023-01-23
Fix windows signing

mattahorton:
So Windows signing has been silently failing since the PR @dyfer made and that I verified to move signing scripts out of `jacktriplabs.yml`. I moved the Windows signing process _back_ to a linux runner but still in the main build workflow, and it works great.

This should clear anything blocking the release of 1.7.0. (Pending an RC.2)

### DIFF
--- a/.github/workflows/jacktrip.yml
+++ b/.github/workflows/jacktrip.yml
@@ -583,7 +583,9 @@ jobs:
           bash -c "set -e; cp $BUILD_PATH/${{ matrix.installer-path }}/JackTrip-${{ steps.set-version.outputs.version }}-${{ steps.set-version.outputs.name }}-installer.msi signed/installer.msi"
           ls CodeSignTool.bat
           ls signed
-          call "dir; dir signed; CodeSignTool.bat sign -credential_id=%CRED_ID% -username=%USERNAME% -password=%PASSWORD% -totp_secret=%TOTP% -output_dir_path=signed -input_file_path=signed\installer.msi"
+          call "dir"
+          call "dir signed"
+          call "CodeSignTool.bat sign -credential_id=%CRED_ID% -username=%USERNAME% -password=%PASSWORD% -totp_secret=%TOTP% -output_dir_path=signed -input_file_path=signed\installer.msi"
           ls signed
           bash -c "set -e; cp signed/installer.msi $BUILD_PATH/${{ matrix.installer-path }}/JackTrip-${{ steps.set-version.outputs.version }}-${{ steps.set-version.outputs.name }}-signed-installer.msi"
           bash -c "ls $BUILD_PATH/${{ matrix.installer-path }}/JackTrip-${{ steps.set-version.outputs.version }}-${{ steps.set-version.outputs.name }}-signed-installer.msi"
--- a/.github/workflows/jacktrip.yml
+++ b/.github/workflows/jacktrip.yml
@@ -576,9 +576,9 @@ jobs:
           unzip CodeSignTool-v1.2.6-windows.zip
           mkdir signed
           echo "before print"
-          ls $BUILD_PATH/
-          ls $BUILD_PATH/${{ matrix.installer-path }}/
-          ls $BUILD_PATH/${{ matrix.installer-path }}/JackTrip-${{ steps.set-version.outputs.version }}-${{ steps.set-version.outputs.name }}-installer.msi
+          ls $BUILD_PATH
+          ls $BUILD_PATH\${{ matrix.installer-path }}
+          ls $BUILD_PATH\${{ matrix.installer-path }}\JackTrip-${{ steps.set-version.outputs.version }}-${{ steps.set-version.outputs.name }}-installer.msi
           echo "before cp"
           bash -c "cp $BUILD_PATH/${{ matrix.installer-path }}/JackTrip-${{ steps.set-version.outputs.version }}-${{ steps.set-version.outputs.name }}-installer.msi signed/installer.msi"
           ls CodeSignTool.bat
--- a/.github/workflows/jacktrip.yml
+++ b/.github/workflows/jacktrip.yml
@@ -583,10 +583,11 @@ jobs:
           bash -c "set -e; cp $BUILD_PATH/${{ matrix.installer-path }}/JackTrip-${{ steps.set-version.outputs.version }}-${{ steps.set-version.outputs.name }}-installer.msi signed/installer.msi"
           ls CodeSignTool.bat
           ls signed
-          call "dir .\signed /S"
-          call "CodeSignTool.bat sign -credential_id=%CRED_ID% -username=%USERNAME% -password=%PASSWORD% -totp_secret=%TOTP% -output_dir_path=signed -input_file_path=.\signed\installer.msi"
+          bash -c "ls signed"
+          bash -c "set -e; ./CodeSignTool.sh sign -credential_id=$CRED_ID -username=$USERNAME -password=$PASSWORD -totp_secret=$TOTP -output_dir_path=signed -input_file_path=signed/installer.msi"
+          # call "CodeSignTool.bat sign -credential_id=%CRED_ID% -username=%USERNAME% -password=%PASSWORD% -totp_secret=%TOTP% -output_dir_path=signed -input_file_path=signed\installer.msi"
           ls signed
-          bash -c "set -e; cp signed/installer.msi $BUILD_PATH/${{ matrix.installer-path }}/JackTrip-${{ steps.set-version.outputs.version }}-${{ steps.set-version.outputs.name }}-signed-installer.msi"
+          bash -c "cp signed/installer.msi $BUILD_PATH/${{ matrix.installer-path }}/JackTrip-${{ steps.set-version.outputs.version }}-${{ steps.set-version.outputs.name }}-signed-installer.msi"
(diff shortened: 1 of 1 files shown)


nwang92 (approved):

dyfer (changes requested):
please wait

dyfer:
So... was there now way to do this on Windows? a bit unfortunate :/

I have a comment regarding the actual workflow, coming up.

mattahorton:
> So... was there no way to do this on Windows? a bit unfortunate :/
> 
> I have a comment regarding the actual workflow, coming up.

Is it about the clang stuff I accidentally added back?

dyfer:
> Is it about the clang stuff I accidentally added back?

Nope, I think there's a way to avoid checking for secrets twice.

.github/workflows/jacktrip.yml
```diff
@@ -597,13 +580,105 @@ jobs:
...
+          then
+            echo "Secrets to sign artifacts were configured in the repo"
+            echo "should_run=true" >> $GITHUB_OUTPUT
+          else
+            echo "Secrets to sign artifacts were not configured in the repo"
+            echo "should_run=false" >> $GITHUB_OUTPUT
+          fi
+  sign:
+    needs: [build, check-secrets]
+    if: needs.check-secrets.outputs.should_run == 'true'
```
dyfer, line 614: I think you might not need the `check-secrets` job at all. Would it work to use this conditional here?
```
if: needs.build.steps.check-secrets.outputs.available == 'true'
```
or something like that - basically use the steps that already run in the main build job.

.github/workflows/jacktrip.yml
```diff
@@ -597,13 +580,105 @@ jobs:
...
+                "${{ secrets.ESIGNER_CREDENTIAL_ID }}" != "" && \
+                "${{ secrets.SSL_COM_PWD }}" != "" && \
+                "${{ secrets.KEYCHAIN_PWD }}" != "" ]]; \
+          then
+            echo "Secrets to sign artifacts were configured in the repo"
+            echo "should_run=true" >> $GITHUB_OUTPUT
+          else
+            echo "Secrets to sign artifacts were not configured in the repo"
+            echo "should_run=false" >> $GITHUB_OUTPUT
+          fi
```
dyfer, line 611: maybe not needed, see the other comment

.github/workflows/jacktrip.yml
```diff
@@ -597,13 +580,105 @@ jobs:
...
+    strategy:
+      fail-fast: false # don't abort if one of the build false
+      matrix:
+        include:
+          - name: Sign Windows artifacts
+            release-name: Windows-x64
+            build-job-name: Windows-x64-qmake-gcc-static-bundled_rtaudio
+            runs-on: ubuntu-latest
+            binary-path: binary
+            installer-path: installer
```
dyfer, line 626: technically we don't have to use a matrix here, since this will only be done for Windows; but it can stay

.github/workflows/jacktrip.yml
```diff
@@ -597,13 +580,105 @@ jobs:
...
+          else
+            VERSION=${GITHUB_SHA::7}
+          fi
+          if [[ -n "${{ matrix.release-name }}" && "$GITHUB_REF" == refs/tags/* ]]; then
+            NAME="${{ matrix.release-name }}"
+          else
+            NAME="${{ matrix.build-job-name }}"
+          fi
+          echo "::set-output name=version::$VERSION"
+          echo "::set-output name=name::$NAME"
```
dyfer, line 651: I think we can re-use `set-version` from the build job. It could then be referenced like `needs.build.steps.set-version.outputs.name` (?)

.github/workflows/jacktrip.yml
```diff
@@ -597,13 +580,105 @@ jobs:
           path: ${{ env.BUILD_PATH }}/${{ matrix.bundle-path }}
       - name: upload installer
         uses: actions/upload-artifact@v2
-        if: matrix.installer-path && (!startsWith(github.ref, 'refs/tags/') || (startsWith(github.ref, 'refs/tags/') && matrix.release-name)) && steps.check-secrets.outputs.available != 'true'
+        if: matrix.installer-path && (!startsWith(github.ref, 'refs/tags/') || (startsWith(github.ref, 'refs/tags/') && matrix.release-name)) && (steps.check-secrets.outputs.available != 'true' || runner.os == 'Windows')
```
dyfer, line 583: I'm getting lost in this conditions... but since we always want to upload the artifact now, wouldn't it be better to simply remove the check for secrets? 
```
if: matrix.installer-path && (!startsWith(github.ref, 'refs/tags/') || (startsWith(github.ref, 'refs/tags/') && matrix.release-name))
```

.github/workflows/jacktrip.yml
```diff
@@ -597,13 +580,105 @@ jobs:
...
+          mkdir signed
+          ./CodeSignTool.sh sign -credential_id=$CRED_ID -username=$USERNAME -password=$PASSWORD -totp_secret=$TOTP -output_dir_path=signed -input_file_path=$BUILD_PATH/${{ matrix.installer-path }}/JackTrip-${{ steps.set-version.outputs.version }}-${{ steps.set-version.outputs.name }}-installer.msi
+          cp signed/JackTrip-${{ steps.set-version.outputs.version }}-${{ steps.set-version.outputs.name }}-installer.msi $BUILD_PATH/${{ matrix.installer-path }}/JackTrip-${{ steps.set-version.outputs.version }}-${{ steps.set-version.outputs.name }}-signed-installer.msi
+          rm $BUILD_PATH/${{ matrix.installer-path }}/JackTrip-${{ steps.set-version.outputs.version }}-${{ steps.set-version.outputs.name }}-installer.msi
+      - name: upload installer
         uses: actions/upload-artifact@v2
-        if: matrix.installer-path && (!startsWith(github.ref, 'refs/tags/') || (startsWith(github.ref, 'refs/tags/') && matrix.release-name)) && steps.check-secrets.outputs.available == 'true'
+        if: matrix.installer-path && (!startsWith(github.ref, 'refs/tags/') || (startsWith(github.ref, 'refs/tags/') && matrix.release-name))
```
dyfer, line 681: I'm not sure if we need the conditional here? The job will only run if the secrets are available...

.github/workflows/jacktrip.yml
dyfer: also, it should only run if we have the installer to sign, so maybe additional condition similar to what we had for signing step previously; or just run on tagged builds

dyfer (dismissed):
In addition to my other comments: maybe I'm missing something, but why are the `analyze build`, `save PR metadata for static analysis`, and `upload static analysis` moved to the signing ~~step~~ job?

.github/workflows/jacktrip.yml
mattahorton: Yeah i was pulling it out of other code

.github/workflows/jacktrip.yml
mattahorton: No because one we named "signed" and the other we don't.

dyfer:
As a side note, it would be great to "squash on merge", this change really should take a single commit IMO.

.github/workflows/jacktrip.yml
mattahorton: I basically solved this by replacing the step with the check-secrets job

.github/workflows/jacktrip.yml
dyfer: ah, gotcha.

.github/workflows/jacktrip.yml
dyfer: ah okay, my bad, sorry.

.github/workflows/jacktrip.yml
mattahorton: Honestly, just because I find it easier to debug things. Hiding it in the step doesn't give me a green check mark on if it passes or not and since it's the first thing that runs, i don't have to wait 10 minutes for a build.

.github/workflows/jacktrip.yml
```diff
@@ -235,27 +261,6 @@ jobs:
           echo "version=$VERSION" >> $GITHUB_OUTPUT
           echo "name=$NAME" >> $GITHUB_OUTPUT
           echo "stamp=$(date '+%Y-%m-%d')" >> $GITHUB_OUTPUT # set timestamp for cache
-      - name: check if secrets are available
```
dyfer, line 238: Why not leave the check-secrets as a step and reference it as I proposed previously?

mattahorton: https://github.com/jacktrip/jacktrip/pull/883#discussion_r1084667531

.github/workflows/jacktrip.yml
dyfer: okay, that makes sense.

dhing1024:
@dyfer I think I accidentally marked one of your comments as unresolved (that you already marked resolved) while clicking through this PR. Can you double check that?

dyfer:
@dhing1024 thanks for checking - all looks fine (all comments were resolved either by me or @mattahorton )

.github/workflows/jacktrip.yml
dyfer: I still think it would be a bit cleaner to reference these values from the build job, OR move these to a separate job, like you did with signing. 
At the same time, maybe it's not worth it and can stay as is.

mikedickey (approved):

Issue #884, opened 2023-01-24
1.6.8 dependencies missing in Fedora


cchafe:
I revved up Fedora 37 on a new machine and successfully did a
dnf install jacktrip

but trying to run it failed

**qrc:/vs/vs.qml:2:1: module "QtQuick.Controls" is not installed 
     import QtQuick.Controls 2.12** 

then did
sudo dnf install qt5-qtquickcontrols
but that wasn't enough

**qrc:/vs/vs.qml:2:1: plugin cannot be loaded for module "QtQuick.Controls": Cannot protect module QtQuick.Controls 2 as it was never registered 
     import QtQuick.Controls 2.12** 

sudo dnf install qt5-qtquickcontrols2

and that did the trick. Runs great.

ntonnaett:
qtquickcontrols2 is listed as a dependency in [our documentation](https://jacktrip.github.io/jacktrip/Build/Meson_build/). So that's a downstream issue IMO.

cchafe:
Perhaps we're missing something that maintainers can use to test their builds.
Is there a convention for that?

ntonnaett:
Good point. If we had (unit) tests, they would probably run those after the build. We should ask distribution packagers, if there are other tactics.

ntonnaett:
It's our fault. I'll create a PR.

Enchufa2:
Downstream maintainer here. Two questions:

- Is QuickControls needed too?
- Jacktrip builds just fine without QuickControls2, which means that it doesn't link against libQt5QuickControls2, which means that this is a runtime dependency, right?

ntonnaett:
What's probably happening is that the QML part is evaluated only at run-time. So we can't catch this at compile-time. But I was baffled by the fact too. Is QtQuick itself also just a runtime dependency? I don't know, but I'm not a Qt professional. However, I would handle both as build dependencies because the qt documentation says that you should add them to your used modules in your build system.

AFAIU @cchafe, Jacktrip doesn't run without the quickcontrols package. The CLI and classic GUI probably would.

ntonnaett:
BTW, thank you for your packaging work! It is much appreciated that we always get the latest version on Fedora.

ntonnaett:
We would have catched this with ahead of time compilation. But the Meson issue for this feature wasn't updated since 2020: https://github.com/mesonbuild/meson/issues/6988

ntonnaett:
This issue is fixed in Fedora. So I'm closing this.

[...]

PR #888, opened 2023-01-24
Add QuickControls2 dependency to Meson and Qmake, and enable ahead-of-time compilation of QMLs

ntonnaett:
#884 happened because of the missing QuickControls2 dependency. I'll open an issue downstream.

Edit: Also turns on ahead-of-time compilation of QMLs for the qmake build.
https://doc.qt.io/qt-6/qtquick-deployment.html#ahead-of-time-compilation

### DIFF
--- a/src/JackTrip.cpp
+++ b/src/JackTrip.cpp
@@ -57,6 +57,7 @@
 #include <QThread>
 #include <QTimer>
 #include <QtEndian>
+#include <QtGlobal>
 #include <cstdlib>
 #include <iomanip>
 #include <iostream>
@@ -1290,7 +1291,7 @@ int JackTrip::clientPingToServerStart()
     connect(&mTcpClient, &QTcpSocket::readyRead, this, &JackTrip::receivedDataTCP);
     connect(&mTcpClient, &QTcpSocket::connected, this, &JackTrip::receivedConnectionTCP);
     // Enable CI builds on Ubuntu 20.04 with Qt 5.12.8
-#ifdef __linux__
+#if QT_VERSION < QT_VERSION_CHECK(5, 13, 0)
     connect(&mTcpClient,
             QOverload<QAbstractSocket::SocketError>::of(&QAbstractSocket::error), this,
             &JackTrip::receivedErrorTCP);
--- a/.github/workflows/jacktrip.yml
+++ b/.github/workflows/jacktrip.yml
@@ -67,7 +67,7 @@ jobs:
             # vcpkg-triplet: x64-mingw-static # used when installing rtaudio with vcpkg (Windows)
             # qt-arch: 'win64_msvc2019_64' # specify when using shared (official) qt libraries (Windows)
             static-qt-version: 5.15.3 # if set, it will enable static Qt build
-            qt-static-cache-key: 'v28' # required for static qt; update this to force rebuilding static Qt
+            qt-static-cache-key: 'v29' # required for static qt; update this to force rebuilding static Qt
             jacktrip-path: jacktrip # needed for binary upload
             binary-path: binary # directory relative to build path; triggers artifact upload for jacktrip binary
             # bundle-path: bundle # directory relative to build path; triggers application bundle creation and upload (macOS)
@@ -96,7 +96,7 @@ jobs:
             novs: true
             weakjack: false
             static-qt-version: 5.15.3
-            qt-static-cache-key: 'v28' # we use the same key as the main Linux build
+            qt-static-cache-key: 'v29' # we use the same key as the main Linux build
             jacktrip-path: jacktrip
             binary-path: binary
             build-system: qmake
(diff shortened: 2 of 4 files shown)


ntonnaett:
Turning on ahead-of-time compilation revealed that we use [required properties](https://doc.qt.io/qt-5/qtqml-syntax-objectattributes.html#required-properties) that were introduced with Qt 5.15. So our Qt 5.12 builds fail now.
@dyfer What do you think?

dyfer:
I think it would be nice to support Qt 5.12 (e.g. to allow building on Ubuntu 20.04 without installing Qt from PPA). But if nobody else things it's needed, we can move the static build to Qt 5.15 (not sure which bugfix version? Ubuntu 22.04 ships with 5.15.3, so maybe not later than that?)

dyfer:
I've pushed Qt update to https://github.com/jacktrip/jacktrip/tree/upgrade-qt-for-static-linux-build
That [build](https://github.com/dyfer/jacktrip/actions/runs/4003838305) doesn't run on Ubuntu 20.04 also doesn't seem to run on Fedora 34. I've danced around the "qt platform plugin" problem before, I don't have the energy to look into that again... but I do want to provide an up-to-date build of JT that's compatible with multiple OSs. Not sure what the best way forward is.

ntonnaett:
@dhing1024 Could we avoid to use the required property feature? I saw you used that [here](https://github.com/jacktrip/jacktrip/blob/4eed6979347a82284adc111cedfd89162b13efec/src/gui/Settings.qml#L293).

dhing1024:
> @dhing1024 Could we avoid to use the required property feature? I saw you used that [here](https://github.com/jacktrip/jacktrip/blob/4eed6979347a82284adc111cedfd89162b13efec/src/gui/Settings.qml#L293).

That sounds good

ntonnaett:
> > @dhing1024 Could we avoid to use the required property feature? I saw you used that [here](https://github.com/jacktrip/jacktrip/blob/4eed6979347a82284adc111cedfd89162b13efec/src/gui/Settings.qml#L293).
> 
> That sounds good

Do you have any idea how to do that? I couldn't find anything equivalent for 5.12.

dhing1024:
I wouldn't know, I'm still a bit new to Qt and not super familiar with nuanced differences between 5.12 and 5.15. But going back I'm actually not sure what the `required` property is doing for us currently.

dyfer:
Actually... I'd like to correct my statement: it seems that it is actually possible to build Qt 5.15 static that _does run on Ubuntu 20.04_. We were missing some packages at build time to build the XCB plugin. @ntonnaett could you run the [static build from here](https://github.com/dyfer/jacktrip/actions/runs/4008494290) and see if you see any issues (that we didn't see before)?

If this works for you @ntonnaett, I suggest that this PR is reabased on https://github.com/jacktrip/jacktrip/tree/upgrade-qt-for-static-linux-build and this might be easier to fix than trying to re-introduce 5.12 compatibility.

ntonnaett:
This is incredible news. I've run it on my computer. Seems fine but I only see subscribed Studios. There are some messages that are new to me. And I don't like the OpenSSL version. I would prefer if we used the latest bugfix release. Is the latest open source version 5.15.8?

```
SSL Library: OpenSSL 1.1.1q  FIPS 5 Jul 2022
Log file location: "/home/nils/.local/share/jacktrip/JackTrip/log.txt"
BDB1539 Build signature doesn't match environment
Cannot open DB environment: BDB0091 DB_VERSION_MISMATCH: Database environment version mismatch
```

dhing1024:
> Seems fine but I only see subscribed Studios.

We took out the ability to start a Virtual Studio in public mode about 2 weeks ago, and finished the migration yesterday evening by updating the database entries on the JackTrip Labs end. They've all been migrated to private studios.

dyfer:
> Is the latest open source version 5.15.8?

[Yes](https://download.qt.io/archive/qt/5.15/).
I went with 5.15.3 since this is the last version shipped with Ubuntu 22.04, but maybe compatibility wasn't broken in a later bugfix release. If you update the version and push to this branch I can test the build with newer Qt on Ubuntu 20.04 to see if it still works (we already know that 5.15.3 works).

ntonnaett:
> > Is the latest open source version 5.15.8?
> 
> [Yes](https://download.qt.io/archive/qt/5.15/). I went with 5.15.3 since this is the last version shipped with Ubuntu 22.04, but maybe compatibility wasn't broken in a later bugfix release. If you update the version and push to this branch I can test the build with newer Qt on Ubuntu 20.04 to see if it still works (we already know that 5.15.3 works).

Let's do this in a separate PR. We should update the qt version also for macOS and Windows. I rebased this branch now.

dyfer:
@ntonnaett looks like Linux shared lib builds need to be updated to Ubuntu 22.04, since 20.04 ships with Qt 5.12.

src/JackTrip.cpp
```diff
@@ -1289,14 +1289,7 @@ int JackTrip::clientPingToServerStart()
     // ----------------------------------------------
     connect(&mTcpClient, &QTcpSocket::readyRead, this, &JackTrip::receivedDataTCP);
     connect(&mTcpClient, &QTcpSocket::connected, this, &JackTrip::receivedConnectionTCP);
-    // Enable CI builds on Ubuntu 20.04 with Qt 5.12.8
-#ifdef __linux__
```
dyfer, line 1293: This `ifdef` could potentially be changed into one based on a Qt version, but I guess now we are indeed going to draw a hard line for Qt 5.15, right?

ntonnaett: I thought about that after I did the change. But I mean, the vs gui is broken with 5.12.

dyfer: VS yes, but "novs" version could still be built on an older system with a version-based conditional.

ntonnaett: True. I'll change that.

dyfer (approved):
Thanks! This is great. And in the process we got a working Linux static build again.

ntonnaett:
Thanks for your review! I'm very happy with what just came from checking why a downstream package missed a dependency.

dyfer:
Just a note: this PR also effectively fixed Linux static build, which was broken for a while. Since it's not part of the PR title, it is not mentioned in the release notes for 1.7.1-beta1. But since nobody complained, then maybe nobody needed it... (?)

PR #892, opened 2023-01-25
GHA: reuse output values

dyfer:
GHA deprecated the old method of setting output value a while back. #883 brought back the old method in the signing step, so I'd like to try my suggestion and re-use the value from the build job.
@mattahorton could you please run this in JTL to confirm the signing works fine?

### DIFF
--- a/.github/workflows/jacktrip.yml
+++ b/.github/workflows/jacktrip.yml
@@ -52,6 +52,9 @@ jobs:
     needs: check-secrets
     runs-on: ${{ matrix.runs-on }}
     name: ${{ matrix.name }}
+    outputs: 
+      version: ${{ steps.set-version.outputs.version }}
+      name: ${{ steps.set-version.outputs.name }}
     strategy:
       fail-fast: false # don't abort if one of the build failse
       matrix:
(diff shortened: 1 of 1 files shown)


mattahorton:
Just back from vacation. I'll try this out

mattahorton:
CleanShot 2023-02-03 at 12 42 28@2x

The "Sign Windows Artifacts" step seems to be pulling the version name from the last build job that finished. In this case, a Linux build finished last because it needed to rebuild Qt.

dyfer:
Ah I see, thanks for catching that. I'll look into this.

dyfer:
I see there's no easy way to reuse the value(s) from the build step, so I left them in the signing step. The change now is to 1) remove unnecessary conditional setting of the name (it is taken straight from the matrix) and 2) setting version, still calculated as before, is updated to use the recommended way of setting step's output value. @mattahorton would you mint testing in JTL again? Thanks!

mattahorton:
CleanShot 2023-02-06 at 14 17 02@2x

So this fails in every case _except_ a release, because the names of PR- and push-driven workflow runs are more complicated than those started by a release being created.

dyfer:
Gotcha, thank you for checking. I forgot the signing runs for regular builds as well... fix forthcoming.

dyfer:
Now I think these are the only changes to be made here to get rid of the deprecation warnings (also updated the checkout action version). Sorry for all the hassle with this!

ntonnaett:
@mattahorton This looks straightforward now. Do you have remaining concerns or should we merge this?

dyfer:
While this is now straightforward, it would be best to run this on JTL repo to double-check.

mattahorton:
except for the v3 change, this is also done in #935. I have no doubt it is fine. Approving.

mattahorton (approved):

PR #894, opened 2023-01-27
Advertise Ubuntu PPA and Debian backports for latest releases

### LINKED ISSUE #525: Advertise JackTrip PPA
@umlaeute was so nice to create a [PPA for us](https://launchpad.net/~umlaeute/+archive/ubuntu/jacktrip/). This is helpful for people that don't have the latest jacktrip version in their Ubuntu repo but want to use it. We should advertise this in our documentation.

ntonnaett:
Resolves #525

### DIFF
--- a/docs/Install.md
+++ b/docs/Install.md
@@ -19,17 +19,25 @@ On Linux the easiest way to install JackTrip is to use the distribution's packag
 If your distribution doesn't include the latest release in their repository, there are different
 alternative options to install from.
 
-Ubuntu users can install from a [PPA repository](https://launchpad.net/~umlaeute/+archive/ubuntu/jacktrip/)
-maintained by the packager of Ubuntu's official JackTrip packages.
+=== "Ubuntu"
 
-```bash
-sudo add-apt-repository ppa:umlaeute/jacktrip
-sudo apt update
-sudo apt install jacktrip
-```
+    Ubuntu users can install from a [PPA repository](https://launchpad.net/~umlaeute/+archive/ubuntu/jacktrip/)
+    maintained by the packager of Ubuntu's official JackTrip packages.
 
-The [GitHub releases page](https://github.com/jacktrip/jacktrip/releases) also includes a binary
-which should run on most Linux distributions (x64). This build is known to not look well under Wayland.
+    ```bash
+    sudo add-apt-repository ppa:umlaeute/jacktrip
+    sudo apt update
+    sudo apt install jacktrip
+    ```
+
+=== "Debian"
+
+    Debian stable users can obtain their latest version from the [Debian backports] repository.
+
+=== "All"
+
+    The [GitHub releases page](https://github.com/jacktrip/jacktrip/releases) also includes a binary
+    which should run on most Linux distributions (x64). This build is known to not look well under Wayland.
 
 ## macOS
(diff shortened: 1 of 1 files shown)


ntonnaett:
@dyfer The Flatpak GHA is running for documentation changes. Do you remember how we prevented this?

dyfer:
@ntonnaett [this](https://github.com/jacktrip/jacktrip/blob/dev/.github/workflows/jacktrip.yml#L16-L21) is what prevents the main build from running on docs changes. Something similar would need to be added for Flatpak I think.

ntonnaett:
> @ntonnaett [this](https://github.com/jacktrip/jacktrip/blob/dev/.github/workflows/jacktrip.yml#L16-L21) is what prevents the main build from running on docs changes. Something similar would need to be added for Flatpak I think.

Thanks! Somehow I thought it was more complicated. Opened a PR for this: #897

dyfer:
This doesn't seem to render right: https://github.com/jacktrip/jacktrip/blob/94f58f2565cc8d5dadf99a0524c264e3c168a6ce/docs/Install.md. I think the new sections shouldn't be indented.

Also, I'd personally prefer this
```
echo "deb http://deb.debian.org/debian bullseye-backports main" \
| tee /etc/apt/sources.list.d/backports.list
```
to be written in a single line. Novice users might be confused by a multiline command - I'm writing from my own experience some time ago. However, this is just a slight preference, not a strong request.

docs/Install.md
```diff
@@ -14,7 +14,40 @@ On Linux the easiest way to install JackTrip is to use the distribution's packag
...
+    echo "deb http://deb.debian.org/debian bullseye-backports main" \
+    | tee /etc/apt/sources.list.d/backports.list
+    apt-get update
+    apt-get install -t bullseye-backports jacktrip
+    ``` 
+
+=== "All"
+
+    The [GitHub releases page](https://github.com/jacktrip/jacktrip/releases) also includes a binary
+    which should run on most Linux distributions (x64). This build is known to not look well under Wayland.
```
dyfer, line 50: > This build is known to not look well under Wayland.

Just checking, is this still the case after we updated to Qt 5.15?

ntonnaett: It's still the case with version 1.7.0. We need to switch to Qt6 for this apparently.

ntonnaett:
@dyfer Quote reply doesn't work… You have to check the mkdocs rendering not the Github one. Use `mkdocs serve` for checking.

I changed the multiline command.

docs/Install.md
```diff
@@ -14,7 +14,39 @@ On Linux the easiest way to install JackTrip is to use the distribution's packag
...
+
+    ```bash
+    sudo add-apt-repository ppa:umlaeute/jacktrip
+    sudo apt update
+    sudo apt install jacktrip
+    ```
+
+=== "Debian"
+
+    Debian stable users can obtain their latest version from the [Debian backports] repository.
```
dyfer, line 35: IMO square brackets in `[Debian backports]` look a little weird. It can stay as is, but I think it might look better as one of these:
"Debian Backports"
Debian "Backports"
`Debian Backports`

dyfer (approved):
Sorry for the confusion about the mkdocs formatting. Looks good! (See comment above for one more suggestion)

docs/Install.md
ntonnaett: That should have been a link.

ntonnaett: Fixed

PR #898, opened 2023-01-31
Reorganize Audio Setup and Settings Screens

dhing1024:
Refactors and reorganizes audio setup and settings screens, largely in preparation for future work giving users the ability to select channels on their input devices.

Using RtAudio backend:

Screenshot_20230131_014810
Screenshot_20230131_014826
Screenshot_20230131_014835

Using JACK backend:
Screenshot_20230131_015257
Screenshot_20230131_015324
Screenshot_20230131_015431

Dark Mode:
Screenshot_20230131_014855
Screenshot_20230131_014927


dhing1024:
Update on this: I changed the audio settings screen so that it has the same layout as the choose devices screen.

Also, I included inverting the slider colors in dark mode to this PR.

nwang92:
This doesn't seem to be a problem for functionality, but when I build + run the app from my terminal I see this error in the logs:
```
qrc:/vs/Settings.qml:644: ReferenceError: warningText is not defined
```

nwang92:
This might be a problem with the app in general, but changing my input volume in a studio "saves" the value internally (confirmed by re-joining the studio) but the settings page does not reflect the change. FWIW, save from settings does get reflected to the connected view so it seems to be working in that direction.

Here's a video where you can see the discrepancy:
https://user-images.githubusercontent.com/6201822/216418530-7606c5a3-4059-4575-b182-87ddfce1503a.mov

dhing1024:
> This doesn't seem to be a problem for functionality, but when I build + run the app from my terminal I see this error in the logs:

Meant to fix this but left it dangling by mistake. Should be good now.

> This might be a problem with the app in general, but changing my input volume in a studio "saves" the value internally

Can we leave this one unresolved for now since it at least working in the forward direction?

nwang92 (approved):

mattahorton:
My 1 request @dhing1024 is to look at #902 and see where I added deletes for `m_inputMeter` and see if it's necessary to add those for `m_outputMeter`. I was fixing a leak.

dhing1024:
@mattahorton just checked and I think everything in #902 made it into here, and there were already the same deletes for `m_outputMeter`

PR #899, opened 2023-02-01
Preview docs in Github Actions

dyfer:
I wanted to have an option of previewing documentation changes even if I don't have access to configured environment with mkdocs installed.

This PR adds a GHA job that runs _only on PRs_ and _only on docs changes_, creating a static version of the docs. They are then uploaded as GHA artifacts. This allows previewing the changes from a static rendering, without the need to run mkdocs locally.

This still requires to download the archive, uncompress it and open `*.html` file(s) in the browser, it does not create an online-viewable preview.

Not sure if others will find this useful... any feedback is welcomed!

### DIFF
--- a/.github/workflows/mkdocs-preview.yml
+++ b/.github/workflows/mkdocs-preview.yml
@@ -19,8 +19,12 @@ jobs:
           echo "use_directory_urls: false" >> mkdocs.yml
       - name: Render docs 
         run: mkdocs build
+      - name: Move to a subdirectory
+        run: |
+          mkdir docs-upload
+          mv site docs-upload/JackTrip-docs-${{ github.sha }}
       - name: Upload docs for preview
         uses: actions/upload-artifact@v3
         with:
           name: JackTrip-docs-${{ github.sha }}
-          path: site/
+          path: docs-upload
--- a/.github/workflows/mkdocs-preview.yml
+++ b/.github/workflows/mkdocs-preview.yml
@@ -1,0 +1,26 @@
+name: Render docs preview
+on:
+  pull_request:
+    paths:
+      - mkdocs.yml
+      - docs/**
+
+jobs:
+  build:
+    name: Render docs
+    runs-on: ubuntu-latest
+    steps:
+      - name: Checkout main
+        uses: actions/checkout@v3
+      - name: Install dependencies
+        run: pip install mkdocs-material mkdocs-macros-plugin
+      - name: Prepare for offline browsing
+        run: |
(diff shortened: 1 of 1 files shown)


ntonnaett (approved):
Nice idea! Looks fine. I assume this only works if it is already merged.

dyfer:
Thanks! It works for PRs that make actual changes to the docs, so not _this_ PR :). Here's an example how it works when made it run of every push: https://github.com/jacktrip/jacktrip/actions/runs/4069416120 (I later fixed the reason for node 12 action warning)

ntonnaett:
> Thanks! It works for PRs that make actual changes to the docs, so not _this_ PR :). Here's an example how it works when I made it run on every push: https://github.com/jacktrip/jacktrip/actions/runs/4069416120 (I later fixed the reason for node 12 action warning)

Of course. Might be really helpful for some people.

dyfer:
@ntonnaett OK to merge? (I'm double checking since I removed the repository check)

ntonnaett:
> @ntonnaett OK to merge? (I'm double checking since I removed the repository check)

Looks good. One additional idea to make it faster to use would be to put everything in a subdirectory in the zip file. But I remember that these things are difficult with GHA.

dyfer:
> One additional idea to make it faster to use would be to put everything in a subdirectory in the zip file.

Like [this](https://github.com/dyfer/jacktrip/actions/runs/4069679099)?

ntonnaett:
> > One additional idea to make it faster to use would be to put everything in a subdirectory in the zip file.
> 
> Like [this](https://github.com/dyfer/jacktrip/actions/runs/4069679099)?

Yes. But… So double subdirectories in artefacts is just something that happens with releases?!

dyfer:
> So double subdirectories in artefacts is just something that happens with releases?!

IIRC... Not exactly. Artifact upload on GHA ("Checks") page _always_ compresses files from the provided path into a zip file, even if it already is a single archive. I'm telling it to upload a directory, so it makes an archive from that.
However, this is not what happens for releases, where files are upload as they are listed, without being additionally packaged into an archive, so if you define a single zip then this is what is being uploaded.

We could presumably fix the double-zip issue by not creating the archive for artifact upload (we'd still need to create it for releases though).

dyfer:
I've added adding files into a subdirectory called the same as the archive (`JackTrip-docs-<sha>`). Looks good?

ntonnaett:
> I've added the commit that moves files into a subdirectory named the same as the archive (`JackTrip-docs-<sha>`). Looks good?

Perfect

PR #904, opened 2023-02-03
Select and Configure Input Channels

dhing1024:
I think this is about ready for others to look at, but it needs to be **\*\*thoroughly\*\*** tested before being merged in. There's a good amount of code refactoring and potential for regression bugs. I'm also a bit worried that there may be race conditions that I haven't run into.

### Windows

This doesn't fix the slowness issue on Windows, especially if there are many devices. When switching input channels or between Mono/Stereo the UI hangs. I think this has been looked at in the past without much success, so I'm leaving this as an unresolved action items for future work.

In terms of UI lag and UX, I'm not sure how it will affect Windows users with lots of devices. For all I know it could be worse with these changes.

### UI

- I made additional vertical space by making the initial window size a little taller to make room for new components. This means that some screens look a little awkwardly off-center vertically. (But just by a small amount). Those are minor issues that can be fixed in a separate PR, in my opinion.

- Some of the text might be adjusted later, though that should be pretty easy to update

- Volume meters should always display 2 channels being lit up. However, in the case of "Mono" and "Mix to Mono" being selected, only 1 channel should be transmitted over the network. In this case the second meter channel is just the first one copied over. If you connect to a VS studio, only 1 channel should light up (and patchbay on the server should only indicate 1 connection)

- This PR refactors the setup/settings screen UI state management that covers every dropdown. For example, if the user is currently on a 2-input channel device, has "1 & 2" and "Mix-to-Mono" selected, then switches to a 1 input device, the only option is "1" and "Mono" which are selected automatically. This is implemented in a new function `VirtualStudio::validateDevicesState`

### Testing

Example with a 4i/4o emulated device on MacOS:

https://user-images.githubusercontent.com/26987971/217408018-b9311b92-1562-4d75-9838-6c2b9c58bd5e.mov

If you're on MacOS, you can configure an emulated device using the "Audio MIDI Setup" tool and creating an aggregate device. In this example, the aggregated device is 4 in - 4 out with audio coming in through channels 3 and 4. An actual physical audio interface would probably be better for testing and QA purposes, but if you don't have one this is also an option.

~~### JackTrip CLI~~

~~This PR adds a `--basesendchannelnum` option to the CLI to specify the starting channel (1-indexed). For example, specifying a value of 3 with 2 input channels tells JackTrip to send input channels 3 and 4. There is currently no option to select mono/stereo options with the CLI - it retains the same behavior as before.~~ removed this

### Misc

- The Stereo to Mono mixer is implemented as a Faust plugin. However, this one operates a little differently from the others in that the number of input channels != number of output channels. The StereoToMono plugin acts before any other ProcessPlugins run.

- ~~Channel numbers are 1-indexed all the way through except for at the RtAudioInterface level~~ changed to 0-indexed

- Most of these changes are at the moment only available through VS GUI mode with the RTAudio backend. The new CLI behavior is an exception.

- When selecting 2 channels, only consecutive stereo pairs starting with an odd numbered channel are available.

- This PR does change the constructor for the `JackTrip` class as well as the audio interface classes


src/gui/Settings.qml
```diff
@@ -625,11 +623,164 @@ Item {
...
+                anchors.left: inputMixModeCombo.left
+                anchors.leftMargin: 2 * virtualstudio.uiScale
+                anchors.right: inputMixModeCombo.right
+                anchors.top: inputMixModeCombo.bottom
+                anchors.topMargin: 8 * virtualstudio.uiScale
+                textFormat: Text.RichText
+                wrapMode: Text.WordWrap
+                text: (() => {
+                    if (virtualstudio.inputMixMode === "stereo") {
+                        return "Treat the channels as Left and Right signals, coming through each speaker separately.";
```
nwang92, line 763: IMO it'd be nice to bring up the fact that mix-to-mono would save bandwidth which could result in a better overall experience?

src/gui/virtualstudio.cpp
```diff
@@ -113,9 +113,21 @@ VirtualStudio::VirtualStudio(bool firstRun, QObject* parent)
...
+    if (m_inputDevice == QStringLiteral("(default)")) {
+        m_inputDevice = "";
+    }
+    if (m_outputDevice == QStringLiteral("(default)")) {
+        m_outputDevice = "";
+    }
+
+    m_baseInputChannel = settings.value(QStringLiteral("BaseInputChannel"), 1).toInt();
+    m_numInputChannels = settings.value(QStringLiteral("NumInputChannels"), 1).toInt();
+    m_inputMixMode = settings.value(QStringLiteral("InputMixMode"), "Mono").toString();
```
nwang92, line 129: Does case matter here? `Mono` vs `mono` - I see `m_inputMixMode` being compared to the lower-case "mono" in other parts of the code

nwang92, line 129: There's probably more audio-specific context I don't have around this, but I wonder if this is the best setting to pick with respect to backwards compatibility?

I could be wrong, but my understanding is right now all existing JackTrip app users are sending 2 stereo channels (assuming their device supports it). If our goal is to provide a better experience on bandwidth-limited environments by sending only 1 channel, would a default of `m_inputMixMode = "mix-to-mono"` achieve what we want while also preventing existing users from changing their setup?

I'm also looking at the fact that the Setup view could never be seen by existing users because you've unchecked the "Ask again next time" box. IDK, it feels like there may be a small chance that we "break" an existing user's setup by changing the default option from 2 channels/stereo -> 1 channel/mono.

mikedickey: I haven't looked at the code, but agree that we need to have 2 defaults:

1. Existing installs/upgrades: migrate to 2 channels / Stereo (channels 1 & 2)
2. Fresh installs: 1 channel/mono (channel 1)

ntonnaett:
I think we should be agnostic to the channel setup. The user might want to send channels 2,3 and 5 and mix them down to one channel for example. That might not be possible with RtAudio but with JACK and Pipewire. If we have a list of input channels, we can just count them to get the number of channels. For RtAudio we can take the lowest number as base channel. We could also make arbitrary channels possible by getting all channels from RtAudio but only using the selected.

src/gui/virtualstudio.cpp
ntonnaett: The mix mode should be an enum (class). Maybe it could be added to AudioInterfaceMode.h

src/RtAudioInterface.cpp
```diff
@@ -170,12 +184,30 @@ void RtAudioInterface::setup(bool verbose)
...
+        mNumInChans = dev_info_input.inputChannels
+                      - static_cast<unsigned int>(AudioInterface::getBaseInputChannel())
+                      + 1;
+        AudioInterface::setNumInputChannels(mNumInChans);
+    }
+
+    if (mNumInChans == 2 && mInputMixMode == "mix-to-mono") {
+        AudioInterface::setNumInputChannels(1);
+    } else if (mNumInChans == 2) {
+        AudioInterface::setNumInputChannels(2);
```
ntonnaett, line 204: All those AudioInterface::function() are member functions not class function. To avoid confusion it is better to remove the class name.

src/RtAudioInterface.cpp
```diff
@@ -69,16 +77,22 @@ RtAudioInterface::~RtAudioInterface()
...
 
 //*******************************************************************************
 void RtAudioInterface::setup(bool verbose)
 {
     // Initialize Buffer array to read and write audio and members
-    mNumInChans  = getNumInputChannels();
-    mNumOutChans = getNumOutputChannels();
-    mInBuffer.resize(getNumInputChannels());
-    mOutBuffer.resize(getNumOutputChannels());
+    mNumInChans  = AudioInterface::getNumInputChannels();
```
ntonnaett, line 90: This looks recursive.

ntonnaett: Argh. It is. That probably happened because RtAudioInterface declares mNumInChans and mNumOutChans a second time. Should be removed.

src/AudioInterface.cpp
```diff
@@ -111,35 +114,31 @@ AudioInterface::~AudioInterface()
     delete[] mAudioInputPacket;
     delete[] mAudioOutputPacket;
 #ifndef WAIR  // NOT WAIR:
-    for (int i = 0; i < mNumInChans; i++) {
+    for (int i = 0; i < mInProcessBuffer.size(); i++) {
```
ntonnaett, line 117: This cleanup is good. We should do this more proactive.

src/RtAudioInterface.cpp
```diff
@@ -417,47 +456,31 @@ int RtAudioInterface::stopProcess()
 
 //*******************************************************************************
 void RtAudioInterface::getDeviceList(QStringList* list, QStringList* categories,
-                                     bool isInput)
+                                     QList<int>* channels, bool isInput)
 {
     RtAudio baseRtAudio;
```
ntonnaett, line 461: I think there's a member mRtAudio for this.

src/RtAudioInterface.h
```diff
@@ -96,14 +101,20 @@ class RtAudioInterface : public AudioInterface
...
     QVarLengthArray<float*>
         mInBuffer;  ///< Vector of Input buffers/channel read from JACK
     QVarLengthArray<float*>
         mOutBuffer;     ///< Vector of Output buffer/channel to write to JACK
     RtAudio* mRtAudio;  ///< RtAudio class if the input and output device are the same
+
+    int mNumInChans;        ///< Number of Input Channels
+    int mNumOutChans;       ///< Number of Output Channels
+    int mBaseInChan;        ///< Base Input Channel
+    QString mInputMixMode;  ///< Input Mixing Mode
```
ntonnaett, line 113: Those are partly redundant and should go to the base class.

src/gui/virtualstudio.cpp
dhing1024: Yes this should be lowercase. I probably will go ahead and make this an enum though

dhing1024: Done here: https://github.com/jacktrip/jacktrip/pull/904/commits/d26741e9d08abd2bf3f2bf1d5d20a5606f56f819

dhing1024:
@ntonnaett I think the most recent commits address most of your comments.

- I was able to remove the redundant member variables in `RtAudioInterface.h`. One of the assumptions this broke was that the Audio Interface member variable `mNumInChans` was being used as both the number of input channels from hardware as well as the number of input channels to network. With mix-to-mono, you can have 2 channels from your microphone and only want to send 1 channel to the network. To get around that, I created local variables in `AudioInterface.cpp` where `mNumInChans` and `mNumOutChans` were being used and adjusted the input channel count if we were in the mix-to-mono mode. I'm not sure if that's the best solution, but I think it works.

- With regards to arbitrary input channel selection, I'm not sure how much additional benefit that would bring over being able to select consecutive input channels. Being able to select channels besides 1 and 2, even if they need to be consecutive, is already an improvement over the existing limitations, and is how I imagine how most people would set themselves up anyways. So maybe we can hold that off for another time?

Another random thing this does: in `vsAudioInterface.cpp`, we were directly updating things like the device to use in the `m_audioInterface` variable from the setter function, which I took out. With the device name it's less important, but with something like the number of channels, it does make a difference. Updating the number of channels, for instance, immediately changes `mNumInChans` in the audio interface class before the audio process can stop, which messes with things like buffer allocations and usually leads to the app crashing. See https://github.com/jacktrip/jacktrip/pull/904/commits/e5f240d1054cfd5a69a77e10bd0e58e9a17a48be#diff-d0c60d9b6f990a02918f8d909d0799c60c9883df2965760c8bb2f06d929130f2

src/RtAudioInterface.cpp
dhing1024: Definitely some work to be done in this area. There might be ways to refactor this so it runs faster, particularly on Windows. This PR is getting a bit thick so maybe I'll hold off on this for another time.

dyfer:
I think being able to send channels above the first two from the interface is a great addition. Here are my thoughts:
- while I think that for mono/stereo only functionality consecutive channel selection might be okay, I think the long term goal should be to select any combination of channels, and also be able to mix them down to mono/stereo - or not. I'd suggest looking how SonoBus does this - it allows selecting non-consecutive channels and then freely mix them up into mono/stereo or send as multiple streams.
  - planning for non-consecutive channel selection would probably also require slightly different approach to the CLI flags. Whether non-consecutive channel selection makes it into JT in this PR, I think ideally the way new CLI flags are set up should allow for non-consecutive selection in the future, without adding new flags.
- this might be the first instance where there's a considerable divergence between VS and Classic GUI functionality. Ideally, I'd like such PR to update both GUIs, but I'm not sure if this is doable.

A typical setup for non-consecutive channel selection (with mixdown to mono/stereo, considering the structure of virtual studio servers) is where there's an instrument (keyboard etc) connected to a pair of line inputs, and a microphone for communication and/or singing. In this case mic might be on input 1, while the keyboard on 3&4 or higher up, depending on the audio interface.

ntonnaett:
> @ntonnaett I think the most recent commits address most of your comments.
> 
>     * I was able to remove the redundant member variables in `RtAudioInterface.h`. One of the assumptions this broke was that the Audio Interface member variable `mNumInChans` was being used as both the number of input channels from hardware as well as the number of input channels to network. With mix-to-mono, you can have 2 channels from your microphone and only want to send 1 channel to the network. To get around that, I created local variables in `AudioInterface.cpp` where `mNumInChans` and `mNumOutChans` were being used and adjusted the input channel count if we were in the mix-to-mono mode. I'm not sure if that's the best solution, but I think it works.

I was already wondering about the local variables. Makes sense now. The whole number-of-channels situation is a mess. Hope we can clean this up at some point, but this probably results in a lot of refactoring.

src/AudioInterface.cpp
```diff
@@ -41,21 +41,25 @@
...
+#include "InputMixMode.h"
 #include "JackTrip.h"
 
 using std::cout;
 using std::endl;
 
 //*******************************************************************************
-AudioInterface::AudioInterface(JackTrip* jacktrip, int NumInChans, int NumOutChans,
+AudioInterface::AudioInterface(JackTrip* jacktrip, int BaseInChan, int NumInChans,
+                               int NumOutChans, int InputMixMode,
```
ntonnaett, line 52:
```suggestion
                               int NumOutChans, InputMixMode InputMixMode,
```

If you use InputMixMode everywhere, you don't need to static_cast anymore.

src/JackAudioInterface.cpp
```diff
@@ -86,12 +86,12 @@ JackAudioInterface::JackAudioInterface(
...
 //*******************************************************************************
 JackAudioInterface::JackAudioInterface(
-    int NumInChans, int NumOutChans,
+    int BaseInChan, int NumInChans, int NumOutChans,
 #ifdef WAIR  // wair
     int NumNetRevChans,
 #endif  // endwhere
     AudioInterface::audioBitResolutionT AudioBitResolution, const QString& ClientName)
-    : AudioInterface(nullptr, NumInChans, NumOutChans,
+    : AudioInterface(nullptr, BaseInChan, NumInChans, NumOutChans, -1,
```
ntonnaett, line 94: Is -1 the InputMixMode? If yes, add the meaning to the enum.

OConnorStP:
Is this capability also going to be available in Classic mode and command-line, for both input and output channels?

dyfer:
I see your point @mikedickey . But I think it would be beneficial to at least keep the CLI arguments the same between the two modes, i.e. choose new CLI argument and format that would support non-consecutive inputs, once implemented in the Classic GUI (or for command-line use), e.g. `-audioinputchannelarray [0,1,4]`. 
Once we add new CLI arguments, it's difficult to remove them in order not to break people's workflow, so I think extra care should be exercised when adding new ones.

OConnorStP:
I agree with Marcin, and would like to argue that this isn't merely an interface feature -- it seems to me to be a change to the way that jacktrip handles channels.  a much-needed one, btw.

dhing1024:
I see the point that the CLI option would want to be agnostic to the user's input setup. I can take that out for now while this feature is still not fully mature to avoid any backwards compatibility issues. I can also look into making some of the underlying logic a little more setup-agnostic as well, but not planning on including output channel setup or classic mode in this PR, to limit scope. Hopefully that satisfies usability needs for our Virtual Studio users while also making it easier to extend this work in the future.

dyfer:
Thanks @dhing1024 
I agree that not including the CLI option for now is better than introducing an option that would need to be reworked later.

ntonnaett:
IMO we should include the CLI option but add to the help message that this flag will eventually be removed if a more flexible solution gets implemented.

dhing1024:
I think these updates address people's comments on the overall design of this feature:
- removes the CLI option
- changes some of the low-level code to allow for more agnostic channel setups by passing `QVarLengthArray`'s of integer channel numbers rather than just the number of desired channels.
- changed channel indexing to 0-based. It seems like that may be a little more standard, but not totally sure about that
- uses the input mix mode enum more generally throughout. I left it as-is with `static_cast` for some of the virtual studio code parts because the enums don't play super well with some of the Qt stuff
- changed the enums to have `0 = InputMixMode::UNSET` option.

Let me know if anyone has additional comments they would like to share. I really appreciate the effort others have put in to look at this PR, since there are a lot of changes here.

nwang92:
Still trying to grok all the code changes here, but from functional testing this seems to work well!

src/gui/Setup.qml
```diff
@@ -1259,11 +1257,164 @@ Item {
...
+                id: inputMixModeHelpMessage
+                anchors.left: inputMixModeCombo.left
+                anchors.leftMargin: 2 * virtualstudio.uiScale
+                anchors.right: inputMixModeCombo.right
+                anchors.top: inputMixModeCombo.bottom
+                anchors.topMargin: 8 * virtualstudio.uiScale
+                textFormat: Text.RichText
+                wrapMode: Text.WordWrap
+                text: (() => {
+                    if (virtualstudio.inputMixMode === "stereo") {
```
nwang92, line 1396: I think these messages have disappeared because `virtualstudio.inputMixMode` is no longer a string

src/gui/Settings.qml
```diff
@@ -626,39 +670,165 @@ Item {
...
+                id: inputMixModeHelpMessage
+                anchors.left: inputMixModeCombo.left
+                anchors.leftMargin: 2 * virtualstudio.uiScale
+                anchors.right: inputMixModeCombo.right
+                anchors.top: inputMixModeCombo.bottom
+                anchors.topMargin: 8 * virtualstudio.uiScale
+                textFormat: Text.RichText
+                wrapMode: Text.WordWrap
+                text: (() => {
+                    if (virtualstudio.inputMixMode === "stereo") {
```
nwang92, line 810: Same here regarding the type of `virtualstudio.inputMixMode`

src/RtAudioInterface.cpp
```diff
@@ -40,27 +40,35 @@
 #include <QString>
 #include <cstdlib>
 
+#include "./InputMixMode.h"
+#include "InputMixMode.h"
```
nwang92, line 44: Duplicate imports here?

src/JackTrip.cpp
```diff
@@ -237,34 +251,66 @@ void JackTrip::setupAudio(
...
         mAudioInterface->setBufferSizeInSamples(mAudioBufferSize);
         mAudioInterface->setup(true);
         // Setup might have reduced number of channels
+
+        // TODO: Add check for if base input channel needs to change
         mNumAudioChansIn  = mAudioInterface->getNumInputChannels();
         mNumAudioChansOut = mAudioInterface->getNumOutputChannels();
+        if (mNumAudioChansIn == 2
+            && static_cast<int>(mInputMixMode)
+                   == static_cast<int>(InputMixMode::MIXTOMONO)) {
```
ntonnaett, line 279: Those static_casts shouldn't be necessary because both are of type InputMixMode.

src/RtAudioInterface.cpp
```diff
@@ -170,12 +196,17 @@ void RtAudioInterface::setup(bool verbose)
...
 
-    if (static_cast<unsigned int>(getNumInputChannels()) > dev_info_input.inputChannels) {
-        setNumInputChannels(dev_info_input.inputChannels);
+    if (static_cast<unsigned int>(baseChanIn) + static_cast<unsigned int>(chansIn)
+        > dev_info_input.inputChannels) {
+        baseChanIn = 0;
+        chansIn    = 2;
+        if (static_cast<unsigned int>(dev_info_input.inputChannels) < 2) {
+            chansIn = 1;
+        }
```
ntonnaett, line 205: The first one looks like dev_info_input.inputChannels is unsigned. So the static_cast the second time  is not necessary. Did you try making baseChanIn, chansIn and chansOut unsigned in the first place?

ntonnaett:
Those last changes are really appreciated. Makes it a lot more future proof IMO. Great work! I'm pretty happy with what we ended up now.

src/gui/Setup.qml
dhing1024: that is definitely what happened there

src/JackTrip.cpp
dhing1024: hmm I thought an early run of GHA was getting caught up on this line, but I guess it must have been some other thing

src/RtAudioInterface.cpp
dhing1024: Yes that turns out to be cleaner

nwang92 (approved):

psiborg112:
A bit late to the party here, but is there any reason why InputMixMode isn't just an enum within AudioInterface.h like all of the other Audio Interface related properties?

psiborg112:
As a follow on from my previous comment, I have a branch (inputMixMode) that does exactly that. It gets rid of InputMixMode.h and moves the enum to AudioInterface.h (which was already an include in every file that InputMixMode.h was used in). Also renamed to inputMixModeT to keep it consistent with previous naming conventions.

mikedickey (approved):

mikedickey:
@psiborg112 since this is a big one that has been hanging around for a while, I think we should merge this and do a separate PR for your changes to inputMixMode?

Issue #896, opened 2023-01-28
Distortion in PLC using FPP of 1024


cchafe:
The symptom is robotic-sounding audio distortion. Bug discovered related to #846
The problem is in bufstrategy 3 when FPP = 1024.

mHist in Regulator.cpp gets set to 3 and that's causing the problem)
setting it to 1 fixes things
(more research needed before I create a PR)

A pipewire host can experience the problem because of the default FPP of 1024. The bug is not related to the choice of audio backend but rarely do hosts run FPP of 1024 these days, so this surfaced because of pipewire.

In debugging this it was noted that signalReceivedNetworkPacket is misleadingly named. It's actually emitted by the audio callback mechanism and would be better called signalTriggerPullPacket.

cchafe:
The current theory is 
set mHist to 2
disable auto
use -q 100 
...............................
On the server I'm seeing the DSP process take about 55ms -- at 1024 a packet is 21ms
it seems we can get away with mCoeffs.resize(trainSamps/4 -2
and that fixes the bug

cchafe:
recommending 
    mCoeffs.resize(trainSamps/8 - 2, 0.0);
and will create a bugfix PR

cchafe:
there's another piece to this
the receiver needs to have greater tolerance when the transmitter is 1024
default -q auto3 is too tight, but -q auto40 works

ntonnaett:
Is this fixed with #906 ?

cchafe:
Yes, but related work is underway to generalize this. I'll "unbug" it for appearance's sake.

cchafe:
I'll be adding to the record here as I get further (with setting pipewire FPP from inside JackTrip via rtaudio)

ntonnaett:
I did some preparatory work / refactoring of the AudioInterface classes to make it easier adding more backends like Pipewire. The Pipewire API is complicated. I need some time to wrap my head around it, especially the threading part.

cchafe:
I have a working "hacktrip" client which is entirely rtaudio/ALSA and sets FPP for pipewire. Code is in https://github.com/cchafe/hacktrip/tree/dev : Hapitrip class, largely undocumented, will try to do that and get it into JackTrip next. BTW, works with either rtaudio 5.2.0 from doc site or from latest github 6beta1 (distinguished by an ifdef in my code)

ntonnaett:
Does this also work with RtAudio/JACK?

cchafe:
[...]

PR #906, opened 2023-02-04
Bugfix/plcBigFPP

cchafe:
fixes issue #896 related to bufstrategy 3 and FPP 1024

the problem was encountered with a linux pipewire laptop client which was running the default 1024
it could not peer with a bufstrategy 3 server without major distortion 

there are two parts to this PR 
it reduces local DSP load for long predictions (by cheating and running with fewer coefficients) 
and it increases receiver mMsecTolerance when a peer is transmitting to it at 1024

1) in the DSP code, ChanData constructor shrinkCoeffsFactor will be 8 if local FPP is 1024

2) in the auto adjust code, calcAuto takes into account mPeerFPPdurMsec
(there was an old comment in the method suggesting this would need to be implemented some day)

This is a deep change and so far runs fine, stress tested with 50% packet loss outgoing from a client

wifi client (running on a linux pipewire machine with default FPP 1024)
`jacktrip  -C <server> --bufstrategy 3 -q auto3 --udprt`

server (mimics a typical VS configuration with FPP 128)
` jacktrip -S -t -z --bindport 4464 --nojackportsconnect --broadcast 97 --bufstrategy 3 -q auto3 -p1 --udprt`

(this is specific to 1024, and could possibly benefit 512, though that's not been looked at)

### DIFF
--- a/src/Regulator.cpp
+++ b/src/Regulator.cpp
@@ -664,6 +664,9 @@ void BurgAlgorithm::predict(std::vector<long double>& coeffs, std::vector<float>
 //*******************************************************************************
 ChanData::ChanData(int i, int FPP, int hist) : ch(i)
 {
+    int shrinkCoeffsFactor = 1;
+    if (FPP == 1024)
+        shrinkCoeffsFactor = 8;
     trainSamps = (hist * FPP);
     mTruth.resize(FPP, 0.0);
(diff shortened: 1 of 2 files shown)


mikedickey (approved):

mikedickey:
Fixed #https://github.com/jacktrip/jacktrip/pull/906

PR #913, opened 2023-02-10
Add refresh button back to settings

mattahorton:
CleanShot 2023-02-10 at 11 24 46@2x


nwang92 (approved):

mattahorton:
Don't merge yet as it would require a new RC. We haven't decided yet if we should do that.

PR #918, opened 2023-02-12
Audiointerface refactory

ntonnaett:
This moves all AudioInterface related files to a new *audio* subfolder. 

~~The AudioInterfaceMode enum class was nice and modern, but you couldn't do things like `AudioInterfaceMode::JACK | AudioInterfaceMode::RTAUDIO`. If there is a method for ORing enum classes, please tell me, because I really don't like the drawback of potentially clashing with C macros.~~

Edit: I mixed now the enum class approach with the possibility of ORing the underlying type with std::to_underlying(). I didn't test this with the real C++23 implementation but the fallback I've added.
https://www.open-std.org/jtc1/sc22/wg21/docs/papers/2021/p1682r3.html

We later have to decide how to proceed with the AudioInterfaceMode setting that is saved as an integer derived by the location in the list of backends in the gui.

There is only one AudioInterfaceMode enum. I deleted *audiointerfaceModeT*.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -133,8 +133,8 @@ set(qjacktrip_SRC
   src/JackTripWorker.cpp
   src/DataProtocol.cpp
   src/UdpDataProtocol.cpp
-  src/AudioInterface.cpp
-  src/JackAudioInterface.cpp
+  src/audio/AudioInterface.cpp
+  src/audio/JackAudioInterface.cpp
   src/JMess.cpp
   src/LoopBack.cpp
   src/PacketHeader.cpp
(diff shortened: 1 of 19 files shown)


ntonnaett:
Thinking about the new AudioInterfaceMode enum again. Maybe it would be better to get rid of AudioInterfaceMode::ALL and NONE instead and stay with the enum class. When we add additional backends ALL becomes quite vague.

ntonnaett:
I just realized that weak_jack should be part of this too. Then we can't use constexpr anymore. So I probably move this to enum class + std::vector.

PR #919, opened 2023-02-12
gha: update clang-format and commit inplace changes

ntonnaett:
This updates clang-format to version 15. Instead of checking the formatting on PRs, it does formatting on push to dev.

### DIFF
--- a/.github/workflows/clang-format.yml
+++ b/.github/workflows/clang-format.yml
@@ -1,9 +1,8 @@
 name: clang-format Check
 on:
   push:
-    paths:
-      - 'src/**'
-  pull_request:
+    branches:
+      - dev
     paths:
       - 'src/**'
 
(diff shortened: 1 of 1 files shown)


dyfer:
I'm not necessarily opposed to this, but I IMO this is a considerable workflow change, as we will potentially get these formatting commits sprinkled in the codebase, and - again potentially - without noticing them until they're already in dev.

Maybe the check in the PR could stay? so that we'll at least see whether changes will be made and PR author can make a devision whether to address that before merging or not.

ntonnaett:
Those formatting commits are already springled in the codebase. One problem with doing this locally is that the formatting differs with the clang-format version. So I thought, we already have those formatting commits, let gha handle that for us.

Getting rid of the formatting commits would be possible by fixups of the illformatted commits. Maybe I'll find how to do this.

dyfer:
> Getting rid of the formatting commits would be possible by fixups of the illformatted commits. Maybe I'll find how to do this.

IIUC think this would be tricky, as it would need to rewrite commits on the remote, which then would cause different commits on the remote and locally (like after a rebase). 

I guess my immediate question here is whether we could still leave notifications in PR about the formatting, or do others feel that's unneeded?

ntonnaett:
We could also add a git prehook. What do you think?

dyfer:
While using a prehook would be a cleaner solution, IIRC that approach uses setting up a specific python environment and that's a bit cumbersome if we're not using python for anything else.
Or, would setting up prehook locally be optional?

ntonnaett:
I think it would be okay to make this optional. Setting up your IDE to clang-format when saving your files is also okay. But it seems that this doesn't fit the workflow of some. Gnome Builder had a huge update where clang-format is not automatically set up anymore. So I think I will add a pre-commit (the python thing you mentioned) setup. Maybe we stay with the old clang-format gha check then?

ntonnaett:
I close this in favour of #920

PR #924, opened 2023-02-15
Updating invite link tooltip to mention referral

nwang92:


mikedickey (changes requested):
I don't think we should put this in the desktop app because it is only optionally updated over time, and we may change or discontinue this offer in the future.

PR #925, opened 2023-02-15
Fix ambiguous call to overloaded funciton in Qt6

psiborg112:
Might be worth having a Qt6 NO_VS build in github actions to check for these kinds of issues. (Sorry, I would offer to do it myself if I knew more about GHA.)

### DIFF
--- a/src/JackTrip.cpp
+++ b/src/JackTrip.cpp
@@ -1316,8 +1316,8 @@ int JackTrip::clientPingToServerStart()
         mAwaitingTcp = true;
         mElapsedTime = 0;
         mEndTime     = 30000;  // Timeout after 30 seconds.
-        mRetryTimer.setInterval(randomizer.bounded(static_cast<int>(0),
-                                                   static_cast<int>(2000 * pow(2, mRetries))));
+        mRetryTimer.setInterval(randomizer.bounded(
+            static_cast<int>(0), static_cast<int>(2000 * pow(2, mRetries))));
         mRetryTimer.setSingleShot(true);
         mRetryTimer.disconnect();
         connect(&mRetryTimer, &QTimer::timeout, this, &JackTrip::tcpTimerTick);
--- a/src/JackTrip.cpp
+++ b/src/JackTrip.cpp
@@ -1316,7 +1316,8 @@ int JackTrip::clientPingToServerStart()
         mAwaitingTcp = true;
         mElapsedTime = 0;
         mEndTime     = 30000;  // Timeout after 30 seconds.
-        mRetryTimer.setInterval(randomizer.bounded(0, 2000 * pow(2, mRetries)));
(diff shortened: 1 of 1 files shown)


ntonnaett:
You might be interested in our clang-format pre-commit hook: https://jacktrip.github.io/jacktrip/DevTools/Formatting/#forget-about-formatting-with-a-git-pre-commit-hook

Where does `2000 * pow(2, mRetries)` come from? Is this even correct? Squared would be `pow(mRetries, 2)`. And then you could just write `mRetries * mRetries`.

ntonnaett:
> Might be worth having a Qt6 NO_VS build in github actions to check for these kinds of issues. (Sorry, I would offer to do it myself if I knew more about GHA.)

I already worked on adding an option to the meson build for building with qt6 or qt5. I don't remember if it was finished.

ntonnaett:
@dhing1024 I think that is your code. Did you mean squared mRetries?

dhing1024:
> @dhing1024 I think that is your code. Did you mean squared mRetries?

I'm pretty sure that was part of work meant to auto-attempt reconnections if the server wasn't immediately available. The idea was to attempt to reconnect based on an exponential backoff timer. See https://github.com/jacktrip/jacktrip/pull/842

So the first time it would try to reconnect after 2 secs, then 4 secs, then 8, and so on.

src/JackTrip.cpp
```diff
@@ -1316,7 +1316,8 @@ int JackTrip::clientPingToServerStart()
         mAwaitingTcp = true;
         mElapsedTime = 0;
         mEndTime     = 30000;  // Timeout after 30 seconds.
-        mRetryTimer.setInterval(randomizer.bounded(0, 2000 * pow(2, mRetries)));
+        mRetryTimer.setInterval(randomizer.bounded(
+            static_cast<int>(0), static_cast<int>(2000 * pow(2, mRetries))));
```
ntonnaett, line 1320:
```suggestion
        mRetryTimer.setInterval(randomizer.bounded(0, 2000 * mRetries * mRetries));
```

ntonnaett: This should give us the desired behavior without the overhead of pow() and the potential casting and (some) bound errors.

ntonnaett: This crashes when using RtAudio/ALSA.

ntonnaett: I think the crash has nothing to do with this change.

ntonnaett (approved):
Your change fixes the Qt6 build. So I approve. ~~But the code there is not doing what is intended and should be fixed. Using `pow(mRetries, 2)` yields the same assert as `mRetries * mRetries`.~~ Had a knot in my brain. But I'm still baffled how this blew up so easy.

Issue #927, opened 2023-02-15
Qt upstream bug comment still relevant?


ntonnaett:
https://github.com/jacktrip/jacktrip/blob/ecce2b04094ea81031a56a9e664a066c4665554c/src/gui/virtualstudio.cpp#L900

The bug mentioned is fixed upstream. This is one of two similar places that break with Qt6. Do we still need to use setModifyParametersFunction()?

ntonnaett:
@dyfer Still interested in working on putting #ifdefs for qt6 support?

dyfer:
Actually, it seems it would be [cumbersome to support both Qt5 and Qt6](https://github.com/jacktrip/jacktrip/issues/767#issuecomment-1286294186) for the VS part. If others are interested in figuring out how to handle the QML part for both Qt5 and Qt6, then I'd be happy to add whatever else is needed. But if we decide to stay at Qt5 for compatibility a bit longer and just switch to Qt6 later, then there would be no point in adding an #ifdef for this.
(The last time I checked `novs` JT built with both Qt5 and Qt6)

ntonnaett:
Oh, I somehow forgot about that comment. Some days ago I thought about working towards a qt5 LTS version that we could offer people with macOS High Sierra etc. and then do the switch to qt6.

PR #928, opened 2023-02-15
#ifdef setModifyParamtersFunction() for Qt6 support

ntonnaett:
This fixes the Qt6 build for Meson with VS enabled. It's still not working. What's missing are separate QML files for Qt6.

#927 relevant


dyfer:
I support this, but wouldn't it be better to wait until we have a solution for QML with Qt6?

ntonnaett:
Yes, I stop working on this.

ntonnaett:
No point in keeping this open. Let's do this with the switch to Qt6.

PR #929, opened 2023-02-16
GHA: correct Qt download link

dyfer:
I think the current logic is broken for Qt 5.15.0-5.15.2, as seen here: https://github.com/jacktrip/jacktrip/actions/runs/4189642206/jobs/7266985031#step:12:29

It seems that `opensource-` in filename is added only for Qt `5.15.3` and up (but not for 5.15.0-5.15.2, and also not for 6.0.0+)


ntonnaett:
What a coincident. I just opened #930. Is this PR still relevant if we update Qt?

ntonnaett (approved):

dyfer:
Yes, I think it's still valid. Thanks!

PR #930, opened 2023-02-16
GHA: Update Qt to version 5.15.8

ntonnaett:
Not sure if I have to do anything about *Windows-x64-qmake-gcc-shared* and *Windows-x64-meson-msvc-shared*.

### DIFF
--- a/.github/workflows/jacktrip.yml
+++ b/.github/workflows/jacktrip.yml
@@ -135,8 +135,8 @@ jobs:
             nogui: false
             novs: false
             weakjack: true
-            static-qt-version: 5.15.2
-            qt-static-cache-key: 'v07'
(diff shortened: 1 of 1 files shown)


ntonnaett:
There's only 5.15.2 for shared Windows… https://www.qt.io/offline-installers

dyfer:
> There's only 5.15.2 for shared Windows…

Yes, Qt 5 above 5.15.2 is only available as source code, so the official binaries won't be available for them.

ntonnaett:
I think that's okay for now. Maybe we could add shared libraries to our qt build. That shouldn't be any real overhead.

ntonnaett:
Thanks for testing!

PR #932, opened 2023-02-17
fix warnings about new qml function syntax

### LINKED ISSUE #931: QML issue
When starting JackTrip (with gui) from the terminal, I'm getting the following message:
```
qrc:/vs/vs.qml:126:5: QML Connections: Implicitly defined onFoo properties in Connections are deprecated. Use this syntax instead: function onFoo(<arguments>) { ... }
Log file location: "/Users/marcin/Library/Application Support/jacktrip/JackTrip/log.txt"
```
I can see this on both macOS and on Windows.
Moreover, the mentioned log file does not seem to get created.

EDIT: I can see this on both 1.6.8 as well as a recent `dev` build.

ntonnaett:
Fixes #931

### DIFF
--- a/src/gui/vs.qml
+++ b/src/gui/vs.qml
@@ -125,23 +125,23 @@ Rectangle {
 
     Connections {
         target: virtualstudio
-        onAuthSucceeded: { 
+        function onAuthSucceeded() {
             if (virtualstudio.showDeviceSetup) {
                 virtualstudio.windowState = "setup";
             } else {
                 virtualstudio.windowState = "browse";
             }
         }
-        onAuthFailed: {
+        function onAuthFailed() {
             loginScreen.failTextVisible = true;
         }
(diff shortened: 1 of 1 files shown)


nwang92:
Did we switch to QT 5.15.x on linux? Looking at my last PR that touched these lines, there was some bug around QT here https://github.com/jacktrip/jacktrip/pull/632

ntonnaett:
We use 5.15 for a long time on distributions and flathub. The static build switched with version 1.7.1.

ntonnaett:
I asked in #927 if we could remove part of #632. Didn't know it was #632 that introduced it.

ntonnaett:
As @dyfer reminded me somewhere else, 5.15 is our minimum version now. So we shouldn't break anything with this.

mikedickey (approved):

PR #933, opened 2023-02-17
Attempting to fix memory-related crash

nwang92:
I'm not sure if this fixes it, but I found that `QVector` does not implement a `detach()` method: https://doc.qt.io/qt-5/qvector-members.html

This change should effectively do the same thing, but manually

### DIFF
--- a/src/Meter.cpp
+++ b/src/Meter.cpp
@@ -145,8 +145,10 @@ void Meter::onTick()
 {
     if (hasProcessedAudio) {
         /* Send the measurements to whatever other component requests it */
-        QVector<float> valuesCopy(mValues);
-        valuesCopy.detach();
+        QVector<float> valuesCopy(mValues.size());
+        for (int i = 0; i < mValues.size(); ++i) {
+            valuesCopy[i] = mValues.at(i);
+        }
         emit onComputedVolumeMeasurements(valuesCopy);
 
         /* Set meter values to the default floor */


ntonnaett:
Do you try to prolong the lifetime?

nwang92:
> Do you try to prolong the lifetime?

Mostly just trying to create a deep copy of `mValues` without using some of the built-in QT abstractions because I'm not sure how they're doing memory management. We've experienced a few crashes on mac like follows:
```
jacktrip(1009,0x10844ce00) malloc: *** error for object 0x7fc507736920: pointer being freed was not allocated

Thread 0 Crashed:: Dispatch queue: com.apple.main-thread
0   libsystem_kernel.dylib        	0x00007fff2093b91e __pthread_kill + 10
1   libsystem_pthread.dylib       	0x00007fff2096a5bd pthread_kill + 263
2   libsystem_c.dylib             	0x00007fff208bf406 abort + 125
3   libsystem_malloc.dylib        	0x00007fff2079f165 malloc_vreport + 548
4   libsystem_malloc.dylib        	0x00007fff207a22aa malloc_report + 151
5   org.jacktrip.jacktrip         	0x0000000105fbb368 Meter::onTick() + 776
6   org.jacktrip.jacktrip         	0x0000000106b69f2a void doActivate<false>(QObject*, int, void**) + 894
```

mikedickey:
I think it does support `detach()`. If it didn't, this code wouldn't compile. See https://github.com/radekp/qt/blob/master/src/corelib/tools/qvector.h

mikedickey:
I suspect that the crash is caused by the enormous amount of memory alloc/dealloc's required to keep updating meters using QVectors. Reference counting container classes are great for convenience, but poor for anything that is happening very frequently within an app.

mikedickey (approved):

mikedickey:
Could also just be a bug in `QVector::detach` but my fear is that it's either a bug with `QVector`'s handling of heap memory or it getting too fragmented from all the ops. Could also be a combination of these, where the custom memory management code it seems to have implemented doesn't account for this type of use. If either are trying, replacing `QVector` is going to be required to fix this.

And do we _really_ need another `vector` implementation in C++ anyway? `std::vector` has been around now for decades. I'd have a lot more confidence in that.

psiborg112:
I have a revision that gets rid of QVector and uses float[]. Can create a pull request for it when #934 is merged. (It's based off that code.)

nwang92:
> I have a revision that gets rid of QVector and uses float[]. Can create a pull request for it when #934 is merged. (It's based off that code.)

Happy to defer to you here then since I'm not sure this will really make a difference

PR #934, opened 2023-02-18
Parse command line for GUI

psiborg112:
Initial work to allow the GUI (standard mode) to parse command line options. Currently works for all required arguments, optional arguments (with a couple of exceptions), audio patching, and authentication options. This is all outlined in the --help screen.

If someone can test that this has no ill effects for VirtualStudio mode, that would be greatly appreciated. (It shouldn't do, but it would be good to test that deeplinks aren't affected.)

### DIFF
--- a/src/Settings.cpp
+++ b/src/Settings.cpp
@@ -273,19 +273,16 @@ void Settings::parseInput(int argc, char** argv)
         case 's':  // Run in P2P server mode
             //-------------------------------------------------------
             mJackTripMode = JackTrip::SERVER;
-            mRunMode      = P2P_SERVER;
             checkMode();
             break;
         case 'S':  // Run in Hub server mode
             //-------------------------------------------------------
-            mJackTripServer = true;
-            mRunMode        = HUB_SERVER;
+            mJackTripMode = JackTrip::SERVERPINGSERVER;
             checkMode();
             break;
         case 'c':  // P2P client mode
             //-------------------------------------------------------
             mJackTripMode = JackTrip::CLIENT;
-            mRunMode      = P2P_CLIENT;
             mPeerAddress  = optarg;
             checkMode();
             break;
@@ -297,7 +294,6 @@ void Settings::parseInput(int argc, char** argv)
         case 'C':  // Ping to server
             //-------------------------------------------------------
             mJackTripMode = JackTrip::CLIENTTOPINGSERVER;
-            mRunMode      = HUB_CLIENT;
             mPeerAddress  = optarg;
             checkMode();
             break;
--- a/src/Settings.cpp
+++ b/src/Settings.cpp
@@ -651,7 +651,6 @@ void Settings::parseInput(int argc, char** argv)
                 optarg = argv[optind++];
             }
             mPassword = optarg;
-            std::cout << mPassword.toStdString() << std::endl;
             break;
         case 'x': {  // examine connection (test mode)
             //-------------------------------------------------------
--- a/src/Settings.cpp
+++ b/src/Settings.cpp
@@ -640,7 +640,10 @@ void Settings::parseInput(int argc, char** argv)
             mCredsFile = optarg;
             break;
         case OPT_AUTHUSER:
-            // Need to manually check if we have our optional argument
+            // Need to manually check if we have our optional argument.
+            // (getopt_long will only find an optional parameter for long arguments if
+            // there's a '=' rather than a space between them. If we don't manually check,
+            // the paramater will be interpreted as an unknown argument.)
             if (optarg == NULL && optind < argc && argv[optind][0] != '-') {
                 optarg = argv[optind++];
             }
--- a/src/Settings.cpp
+++ b/src/Settings.cpp
@@ -291,7 +291,7 @@ void Settings::parseInput(int argc, char** argv)
             break;
         case 'L':  // set optional local host address
             //-------------------------------------------------------
-            mGuiIgnoresArgemunts = true;
+            mGuiIgnoresArguments = true;
             mLocalAddress        = optarg;
             break;
         case 'C':  // Ping to server
@@ -380,17 +380,17 @@ void Settings::parseInput(int argc, char** argv)
             break;
         case 'l':  // loopback
             //-------------------------------------------------------
-            mGuiIgnoresArgemunts = true;
+            mGuiIgnoresArguments = true;
             mLoopBack            = true;
             break;
         case 'e':  // jamlink
             //-------------------------------------------------------
-            mGuiIgnoresArgemunts = true;
+            mGuiIgnoresArguments = true;
             mEmptyHeader         = true;
             break;
         case 'j':  // jamlink
             //-------------------------------------------------------
-            mGuiIgnoresArgemunts = true;
+            mGuiIgnoresArguments = true;
             mJamLink             = true;
             break;
         case 'J':  // Set client Name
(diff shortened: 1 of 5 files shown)


src/Settings.cpp
```diff
@@ -269,24 +273,33 @@ void Settings::parseInput(int argc, char** argv)
         case 's':  // Run in P2P server mode
             //-------------------------------------------------------
             mJackTripMode = JackTrip::SERVER;
+            mRunMode      = P2P_SERVER;
```
ntonnaett, line 276: Why do we need mJackTripMode and mRunMode? Isn't that the same?

src/Settings.cpp
```diff
@@ -367,15 +380,18 @@ void Settings::parseInput(int argc, char** argv)
             break;
         case 'l':  // loopback
             //-------------------------------------------------------
-            mLoopBack = true;
+            mGuiIgnoresArgemunts = true;
```
ntonnaett, line 383: Typo: mGuiIgnoresArguments

src/Settings.cpp
```diff
@@ -875,6 +950,9 @@ void Settings::printUsage()
     cout << " --username                               The username to use when connecting as a hub client (if not supplied here, this is read from standard input)" << endl;
     cout << " --password                               The password to use when connecting as a hub client (if not supplied here, this is read from standard input)" << endl;
     cout << endl;
+    cout << "ARGUMENTS FOR THE GUI:" << endl;
+    cout << " --gui                                   Force JackTrip to run with the GUI. If not using VirtualStudio mode, command line switches in the required arguments, optional arguments (except -l, -j, -L, --appendthreadid), audio patching, and authentication sections will be honoured, and default settings will be used where arguments aren't supplied. Options from other sections will be ignored (and the last used settings will be loaded), except for -V, and the --version and --holp switches which will override this." << endl;
```
ntonnaett, line 954: Typo: holp

src/Settings.cpp
psiborg112: I was wondering how it even managed to compile with such a typo, then realised I must have made it the first time and autocompleted every other time.

src/Settings.cpp
psiborg112: Similar but slightly different. (You'll notice mJackTripMode was never set with -S because it's not needed at this point in the code.) I guess it would be possible to set mJackTripMode to JackTrip::SERVERTOPINGSERVER to show that we're in hub server mode, but the advantage of runModeT is that it matches the more user friendly order that exists in the GUI dropdown. If we slightly repurpose mJackTripMode then we'll need code to translate the order used between the two parts of the program. (Alternatively we could change the order within the jacktripModeT enum if we're sure it won't break any assumptions made about those values elsewhere in the program.)

psiborg112: When I get a chance later today, I'll check the code to see if I can safely change the enum order. If I can, I'll consolidate those two variables.

psiborg112: Scratch that - I'll just do the translation. It's less efficient, but it means there's less chance that a change in one part of the code will break things elsewhere down the line.

psiborg112: (Expect a revision in the not too distant future.)

nwang92:
Ran this build on OSX and deep-links did not seem to be impacted by the new flags

psiborg112:
As soon as someone signs off on this, I'll create a pull request for the meter updates. If anyone wants to get a head start on looking at the code, it's in the meters-and-scripting branch.

ntonnaett:
I'll look into it tomorrow.

src/Settings.cpp
```diff
@@ -596,13 +640,24 @@ void Settings::parseInput(int argc, char** argv)
             mCredsFile = optarg;
             break;
         case OPT_AUTHUSER:
+            // Need to manually check if we have our optional argument
+            if (optarg == NULL && optind < argc && argv[optind][0] != '-') {
+                optarg = argv[optind++];
+            }
```
ntonnaett, line 646: How is this supposed to work? If there's no optional argument it should take the next one, but not if it is a new flag? How is this possible, if there is no optional argument?

src/Settings.h
```diff
@@ -60,7 +60,12 @@ class Settings : public QObject
     Q_OBJECT;
 
    public:
-    Settings(QObject* parent = nullptr) : QObject(parent), mAudioTester(new AudioTester)
+    Settings(bool guiEnabled = false, QObject* parent = nullptr)
+        : QObject(parent)
+#ifndef NO_GUI
+        , mGuiEnabled(guiEnabled)
+#endif
```
ntonnaett, line 67: mGuiEnabled depends on the default initialization value for bools for nogui builds? I think it's easier to reason about it without the ifdef.

src/gui/qjacktrip.cpp
```diff
@@ -1186,9 +1144,217 @@ void QJackTrip::loadSettings()
...
+        m_ui->authCheckBox->setChecked(cliSettings->getUseAuthentication());
+        m_ui->certEdit->setText(cliSettings->getCertFile());
+        m_ui->keyEdit->setText(cliSettings->getKeyFile());
+        m_ui->credsEdit->setText(cliSettings->getCredsFile());
+        m_ui->usernameEdit->setText(cliSettings->getUsername());
+        m_ui->passwordEdit->setText(cliSettings->getPassword());
+
+        settings.beginGroup(QStringLiteral("JitterBuffer"));
+        settings.setValue(QStringLiteral("JitterAnnounce"), true);
+        int bufferStrategy = cliSettings->getBufferStrategry();
```
ntonnaett, line 1232: Typo: getBufferStrategry

ntonnaett (changes requested):
For this kind of changes automatic tests would be really good. We should give the next release some more testing time. Except my other comments this looks good to me.

src/Settings.h
psiborg112: It's just so that in no GUI builds there's no risk of setting the variable to a value that should not be used.

psiborg112: So in the NO_GUI builds, it'll always be false, regardless of whether someone passes something they shouldn't to the constructor. That way I didn't have to put a whole bunch of #idefs elsewhere in the code.

src/gui/qjacktrip.cpp
psiborg112: Damn it. That means it's wrong everywhere. Will change.

src/Settings.cpp
psiborg112: It's related to this sort of issue: [getopt does not parse optional arguments to parameters](https://stackoverflow.com/questions/1052746/getopt-does-not-parse-optional-arguments-to-parameters)

--username and --password take optional parameters. If they're not supplied then the user gets prompted at the command line for them. The problem is that getopt_long wants long options to take the form --username=user when working out whether the next argument supplied is the optarg or whether it's the next command switch. So when users don't include the equals sign (as they typically don't) it treats 'user' as a command switch and then chews the user out for having a non-recognized command. This tests to see if the next string in argv should be interpreted as the optarg for these commands, or whether we should leave the parser to do its thing.

psiborg112: (So this has been busted from the start. I must have just got lucky in the way I was originally testing it. (Only omitting one optional argument which I must have conveniently placed at the end of the command.)

src/Settings.h
ntonnaett: Alright

src/Settings.cpp
ntonnaett: I don't understand your last comment, but with your explanation this sounds reasonable. Could you add a comment that this is about the case without the equal sign?

psiborg112: Sure.

psiborg112: Oh, I mean that this has never properly worked since the introduction of the code. I just got lucky in how I tested it. (It worked for one case, but not in general.)

ntonnaett (approved):

PR #935, opened 2023-02-24
Automatically release to auto-updater

mattahorton:
**To start**:

We ran into an issue where Github was caching the manifest files for multiple days. So we're moving them to files.jacktrip.org.

**What this PR does**: 

On release, github actions will now take the newly created and signed release artifacts, upload them to files.jacktrip.org, download the manifest files, update them with a new release that points to those artifacts, and upload the updated manifest files. 

This takes a lot of the manual work out of releasing with the auto-updater.

The last part of the process that is still manual is moving the signed artifacts over the this repo to replace the unsigned ones.

Again, none of this happens on this repo, as the secrets are not available.

**Coming soon**:

A separate PR (then release) to update the url the desktop app points to when looking for updates.

### DIFF
--- a/.github/workflows/jacktrip.yml
+++ b/.github/workflows/jacktrip.yml
@@ -690,21 +690,6 @@ jobs:
           name: JackTrip-${{ steps.set-version.outputs.version }}-${{ steps.set-version.outputs.name }}-signed-installer
           path: ${{ env.BUILD_PATH }}/${{ matrix.installer-path }}
   # release - list of files uploaded to GH release is specified in the *upload* step
-  deploy_test:
-    needs: [build, sign]
-    runs-on: ubuntu-latest
-    name: 'deploy release'
-    env:
-      DOWNLOAD_PATH: ${{ github.workspace }}/Install
-    steps:
-      - name: download artifacts
-        uses: actions/download-artifact@v2
-        with:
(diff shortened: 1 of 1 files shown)


.github/workflows/jacktrip.yml
```diff
@@ -710,3 +711,155 @@ jobs:
...
+            TYPE="binary"
+          fi
+          echo "version=$VERSION" >> $GITHUB_OUTPUT
+          echo "name=$NAME" >> $GITHUB_OUTPUT
+          echo "extension=$EXTENSION" >> $GITHUB_OUTPUT
+          echo "type=$TYPE" >> $GITHUB_OUTPUT
+      - name: download artifacts
+        uses: actions/download-artifact@v2
+        with:
+          path: ${{ env.DOWNLOAD_PATH }} # no "name" parameter - download all artifacts
```
nwang92, line 779: NIT but this "no name" comment doesn't apply to this section

mattahorton: ah yeah

nwang92:
Overall seems fine to me, although I imagine we'll likely find something when we actually use it :)

nwang92:
Not necessarily something to do as part of this PR, but we should also change the URL of the manifest.json files away from github within the code itself

nwang92 (approved):

PR #936, opened 2023-02-25
Replace QVector in meter code 

psiborg112:
Use a couple of float arrays that only get reallocated if the number of channels changes. (And have a working buffer that is only reallocated if it needs to grow.) This should stop the meter code from churning through memory.

As an added bonus, there's now an option in the classic GUI to execute an external script on connection or disconnection.

### DIFF
--- a/src/Auth.cpp
+++ b/src/Auth.cpp
@@ -39,22 +39,31 @@
 #include <QCryptographicHash>
 #include <QDate>
 #include <QFile>
+#include <QRandomGenerator>
 #include <QTextStream>
 #include <QThread>
 #include <iostream>
 
-Auth::Auth(const QString& fileName, QObject* parent)
+Auth::Auth(const QString& fileName, bool monitorChanges, QObject* parent)
     : QObject(parent)
     , m_days({"Mo", "Tu", "We", "Th", "Fr", "Sa", "Su"})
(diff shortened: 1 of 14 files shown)


psiborg112:
Just some notes for reviewers since a few things ended up smooshed together in here. (Was working on a few things at the same time.) The changes in Auth.h and Auth.cpp (and the related minor change in UdpHubListener.cpp) don't change the way anything currently runs. They just add some functions to allow password files to be edited and written out in the future. (There will eventually be a GUI for this.) Also, pretty sure no one other than me is really using the auth functionality at the moment anyway.

The important changes are in Meter.h and Meter.cpp, and there are some flow on effects for this in the user interface code (both classic and VS) that make use of the Meter object. QVectors are gone and replaced with 3 float arrays that are allocated a single time on the heap and used for the lifespan of the object. They are only reallocated if the number of channels changes, or if the working buffer needs to grow. (The biggest benefit of this code change is that we're no longer allocating a new buffer every time that compute is called.) While it shouldn't be possible to change the number of channels used once the code has started running, for safety I've delayed the deletion of the old arrays so that any code still accessing them has more than enough time to wrap up what it's doing.

mikedickey:
Very nice!

mikedickey (approved):

Issue #852, opened 2022-12-15
VirtualStudio interface issue with Qt 5.12


dyfer:
After building locally on Ubuntu 20.04, I can't load the VS interface, despite the app compiling properly and launching. I'm guessing this has to do with compatibility with Qt 5.12.8 (the version available for Ubuntu 20.04) and Qt Quick. 

I'm getting a blank screen (no way to go through the setup or switch to classic interface) and the following output:

``` 
$ builddir/jacktrip
qrc:/vs/vs.qml:102:5: Type Setup unavailable
qrc:/vs/Setup.qml:603:26: Expected token `:'
SSL Library: OpenSSL 1.1.1f 31 Mar 2020
Log file location (...)
```

_Originally posted by @dyfer in https://github.com/jacktrip/jacktrip/issues/845#issuecomment-1353224096_

dyfer:
@ntonnaett I assume this is a suggestion for the Linux static build, correct?

I just want to make sure everybody is aware - this issue means that JackTrip with VirtualStudio cannot be run on older Ubuntu (20.04LTS and earlier - that's one before the latest LTS version) _even if compiled locally_(*). My preference would be to fix compatibility with Qt 5.12 and have appropriate solution to use QuickControls only when compiling for Qt >=5.15. 

But since apparently nobody noticed/complained about it until now, then maybe there's no real-world need for it. I just want to make sure everybody is aware of the consequences of not supporting Qt 5.12 anymore. I'm happy to update the static Linux build to Qt 5.15 if that's what we want.
(Also, this doesn't affect Windows and macOS where we only use Qt 5.15).

(*) maybe it's possible to install QT 5.15 from a PPA on Ubuntu 20.04, but I'm not sure how practical or problematic that is.

ntonnaett:
Maybe we can avoid the required property feature and stay with 5.12 for now. QuickControls isn't the problem here. With ahead-of-time compilation enabled, we would have catched that at compile time.

As you said, nobody complained until now. We don't have any clue about what distributions are actually used. We should try to serve as many distributions as possible, but we should also avoid investing time that doesn't serve anybody. IMHO we should use the same Qt version on all platforms. Maybe something like distrobox or toolbx is a good way to serve old distributions. I'll look into that when I have time. The OSB path wouldn't have helped here because those packages depend on the qt version of the distribution.

dyfer:
[...]

PR #939, opened 2023-02-28
Clean up main

psiborg112:
Main has accumulated a lot of VirtualStudio specific code which, combined with all of the ifdefs being used, has made it a little hard to get a quick overview of what is happening at launch. This PR encapsulates a lot of the deeplink and instance checking code in its own class.

Couple of notes: I've left the author field in vsInit.h and vsInit.cpp blank because it's not really my code. If anyone knows who originally authored it, and could add their name there, that would be great. Also, is the code that checks for an existing instance of the program really only meant to be run under windows? And as always, if someone could test that deeplinks still work as expected, that would be great.

### DIFF
--- a/src/gui/vsInit.cpp
+++ b/src/gui/vsInit.cpp
@@ -82,7 +82,7 @@ void VsInit::checkForInstance(QCoreApplication* app, QString& deeplink,
         Qt::QueuedConnection);
     // Create instanceServer to prevent new instances from being created
     void (QLocalSocket::*errorFunc)(QLocalSocket::LocalSocketError);
-#ifdef Q_OS_LINUX
+#if (QT_VERSION < QT_VERSION_CHECK(5, 15, 0))
     errorFunc = &QLocalSocket::error;
 #else
     errorFunc = &QLocalSocket::errorOccurred;
--- a/src/gui/vsInit.cpp
+++ b/src/gui/vsInit.cpp
@@ -56,15 +56,13 @@ QString VsInit::parseDeeplink(QCoreApplication* app)
     }
 }
 
-void VsInit::checkForInstance(QCoreApplication* app, QString& deeplink,
-                              QSharedPointer<VirtualStudio> vs)
+void VsInit::checkForInstance(QString& deeplink)
 {
-    m_vs = vs;
     // Create socket
-    m_instanceCheckSocket.reset(new QLocalSocket(app));
+    m_instanceCheckSocket.reset(new QLocalSocket(this));
     // End process if instance exists
     QObject::connect(
-        m_instanceCheckSocket.data(), &QLocalSocket::connected, app,
+        m_instanceCheckSocket.data(), &QLocalSocket::connected, this,
         [=]() {
             // pass deeplink to existing instance before quitting
             if (!deeplink.isEmpty()) {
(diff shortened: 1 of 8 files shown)


src/main.cpp
```diff
@@ -285,11 +283,10 @@ int main(int argc, char* argv[])
     QSharedPointer<QJackTrip> window;
 
 #ifndef NO_VS
-    QString deeplink = QLatin1String("");
+    QString deeplink;
     QSharedPointer<VirtualStudio> vs;
 #ifdef _WIN32
-    QSharedPointer<QLocalServer> instanceServer;
-    QSharedPointer<QLocalSocket> instanceCheckSocket;
+    QScopedPointer<VsInit> vsInit;
```
ntonnaett, line 289: This instance doesn't seem to be used but just reset for Windows. Could be removed here and declared later when used.

mattahorton:
> Also, is the code that checks for an existing instance of the program really only meant to be run under windows? 

Correct.

psiborg112:
@mattahorton thanks for confirming that

src/gui/vsInit.cpp
```diff
@@ -0,0 +1,164 @@
...
+                }
+            }
+            emit QCoreApplication::quit();
+        },
+        Qt::QueuedConnection);
+    // Create instanceServer to prevent new instances from being created
+    void (QLocalSocket::*errorFunc)(QLocalSocket::LocalSocketError);
+#if (QT_VERSION < QT_VERSION_CHECK(5, 15, 0))
+    errorFunc = &QLocalSocket::error;
+#else
```
dyfer, line 87: VS GUI [requires Qt 5.15](https://github.com/jacktrip/jacktrip/issues/852) anyway, AFAIR, so this could be possibly removed, I think.

mattahorton:
I've just tested this. It seems that, on Windows, now the existing instance of JackTrip _also_ closes when the 2nd instance is run. 

The expected behavior is that the original instance continues running and the new one is prevented.

psiborg112:
@mattahorton does it close or does it crash? Is there any debug output?

psiborg112:
@OConnorStP This is purely a virtualstudio thing. I'm just trying to clean up the code so that there isn't VS stuff where there really shouldn't be.

mattahorton:
> @mattahorton does it close or does it crash? Is there any debug output?

No crash output. No extra output at all. It just closes.

psiborg112:
So what's the history of this code? Who authored it initially? (And can they take a look at it now?) What's the rationale behind it being only a windows thing?

psiborg112:
OK, so it definitely crashes and I suspect the nested lambda is the issue. I think that by the time the internal lambda is called it has gone out of scope and no longer exists. I might have to actually derive vsInit from QObject so that these lambdas can be redefined as slots. (Which, to be honest, would also have the added benefit of making the code easier to follow.)

psiborg112:
As an aside, I've tested the original code, and one undesired outcome is that if you're in classic mode then the second instance will launch a VS login screen if you're not already logged in. Which it really shouldn't do.

psiborg112:
So that should fix it. (There was a null pointer issue because vs wasn't set in main when this was called.) This is some much needed cleanup, and should definitely still be merged once approved. But, having said that, neither this code nor the code that it replaces is fit for purpose and it needs to be rethought. It doesn't even remotely take into account the fact that VirtualStudio mode is not the only mode the program operates in. You could be in the middle of a classic mode session, connected to a server, and this code will still bring up the VS window. There's no reason for the user to assume that this is the same instance of the program that they're actively using, and if the VS login window gets stuck (as it did most of the time when I was testing the impacts of this) and the user closes it to get back to what they were doing, they end up closing the entire program.

This needs to be fixed, and it should not be on the open source developers to fix it. I suspect that a lot of the testing of new features by the VirtualStudio team is being done entirely within VS mode. Which is fine for things that don't impact the rest of the project. (And the architecture is modular enough that this will often be the case.) But it's unfair when your changes put an undue maintenance burden on the open source team. (Many of whom are time poor already.) This discovery already comes off the back of us having a 1.8.0 beta release that does not work at all when using JACK, and that doesn't work as expected with the RtAudio backend outside of VS mode. No one is expecting you to contribute improvements back to the standard interface or the command line. But please don't break these for us.

psiborg112:
Rant aside, this latest commit should fix the issue that I've identified. (I'm on the road at the moment and haven't had a chance to actually test it yet, so absolutely do not merge without testing.)

mattahorton:
I've been thinking a lot about regression testing for JackTrip and what that would take. I know nothing about the frameworks for that in Qt/C++. But it's clear there is a need.

mattahorton (approved):

psiborg112:
It would be great to have automated tests, but even those wouldn't really have been necessary with #904. All it would have taken would have been to run the far reaching code (that touched far more than just VS mode and RtAudio) in any other configuration - the bugs happened instantly and 100% of the time. (Could not ask for a more reproducible bug.) I think it's fair for code reviewers to make the assumption that code has actually been run once it reaches the pull request stage, so that their focus is on more technical issues. And if it hasn't for some reason then the author needs to clearly state any limitations to their testing. (It's why I always stress where I haven't been able to test VS features, like deeplinks, where code I write butts up against their implementation.) I think there needs to be an understanding that if your code is outside of specific GUI territory and reaches into common areas then it absolutely needs to be tested in other modes, even if that means you need to collaborate with someone who can do the tests you can't.

PR #942, opened 2023-02-28
Fix Output Device Selection Bug

dhing1024:
When working on the input channels feature, I mistakenly commented out one more line than I intended to, resulting in the user's selection of output device not being respected.


mikedickey (approved):

mikedickey:
This fixes the issue on my laptop

PR #948, opened 2023-03-04
Fix crash when using JACK

psiborg112:
I've taken the path of least resistance to fix this, but I think we need to rethink the constructor for the base AudioInterface class. Channel selection has no meaning when using JACK and passing QVarLengthArrays doesn't make a lot of sense here. I'd suggest it would probably be worth either having an overloaded constructor, or (possibly better) having the channel lists as member variables of the derived RtAudioInterface class instead. (And having the getter and setter functions defined as part of this class rather than the base class.)

### DIFF
--- a/src/AudioInterface.h
+++ b/src/AudioInterface.h
@@ -258,13 +258,9 @@ class AudioInterface
     void computeProcessToNetwork(QVarLengthArray<sample_t*>& in_buffer,
                                  unsigned int n_frames);
 
-    JackTrip* mJackTrip;  ///< JackTrip Mediator Class pointer
-    int mNumInChans;      ///< Number of Input Channels
-    int mNumOutChans;     ///<  Number of Output Channels
     QVarLengthArray<int> mInputChans;
     QVarLengthArray<int> mOutputChans;
-    inputMixModeT mInputMixMode;  ///< Input mixing mode
-#ifdef WAIR                       // wair
+#ifdef WAIR               // wair
     int mNumNetRevChans;  ///<  Number of Network Audio Channels (net comb filters)
     QVarLengthArray<sample_t*>
         mNetInBuffer;  ///< Vector of Input buffers/channel read from net
@@ -298,6 +294,11 @@ class AudioInterface
     AudioTester* mAudioTesterP{nullptr};
 
    protected:
+    JackTrip* mJackTrip;          ///< JackTrip Mediator Class pointer
+    int mNumInChans;              ///< Number of Input Channels
+    int mNumOutChans;             ///<  Number of Output Channels
+    inputMixModeT mInputMixMode;  ///< Input mixing mode
+
     void setDevicesWarningMsg(warningMessageT msg);
     void setDevicesErrorMsg(errorMessageT msg);
 
--- a/src/AudioInterface.h
+++ b/src/AudioInterface.h
@@ -98,6 +98,8 @@ class AudioInterface
      * \param NumInChans Number of Input Channels
      * \param NumOutChans Number of Output Channels
      * \param AudioBitResolution Audio Sample Resolutions in bits
+     * \param processWithNetwork Send audio to and from the network
+     * \param jacktrip Pointer to the JackTrip class that connects all classes (mediator)
      */
     AudioInterface(
         QVarLengthArray<int> InputChans, QVarLengthArray<int> OutputChans,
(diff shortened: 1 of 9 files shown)


psiborg112:
This is a major regression, so if this could get reviewed ASAP, that would be great.

psiborg112:
Scratch that. I decided to do more than the bare minimum on this. Part of the reason this error occurred in the first place is because there were duplicate private member variables in the base and derived classes. This changes these from private to protected where it makes sense to do so. (I'd argue the same can probably be done for the mInputChans and mOutputChans arrays as well if someone wanted to do that.) I've also made it explicit in JackAudioInterface.h where things are being overridden. I got rid of what I thought was an unnecessary overloaded constructor, but it appears it's being used in the VS side of things. So I'm going to re-order some arguments so that a default nullptr value can be passed as the JackTrip pointer. Bear with me...

psiborg112:
And this should be good to go.

cchafe:
Thanks for much for the fix! I agree we are overdue for an overhaul in this area and some future proofing to boot. I'm getting something that looks like an off-by-1 when using RtAudio. This is with 2 channels for in and out and default connections.
Correct (Jack backend)
fix-jack-JackBackend

Incorrect (RtAudio backend)

fix-jack-RtAudioBackend

psiborg112:
That might be to do with 0 vs 1 indexing of channels? I thought it was addressed in later revisions to #904 but I might be wrong. (I'm beginning to suspect that most of the testing for this was probably done in virtual studio mode...)

psiborg112:
That should hopefully fix it. (At least, it does on mine. I'm also in the same situation as you are where JACK is my main audio server and RtAudio just hooks into that.)

psiborg112:
Have tested on mac with RtAudio using coreaudio, and it seems to work properly there as well.

psiborg112:
I just realised that this means we have a beta release out there that will crash entirely when used with JACK, and that won't give the channel assignments that are expected when using RtAudio in anything but VS mode. We need to get a replacement release out as soon as we can.

src/AudioInterface.h
```diff
@@ -98,15 +98,17 @@ class AudioInterface
      * \param NumInChans Number of Input Channels
      * \param NumOutChans Number of Output Channels
      * \param AudioBitResolution Audio Sample Resolutions in bits
+     * \param processWithNetwork Send audio to and from the network
+     * \param jacktrip Pointer to the JackTrip class that connects all classes (mediator)
```
ntonnaett, line 102: The jacktrip parameter is redundant here. Why did you change the order of parameters?

ntonnaett:
Sir, do you have a moment to learn about our lord and savior clang-format pre-commit hook?

https://jacktrip.github.io/jacktrip/DevTools/Formatting/#pre-commit-configuration

cchafe:
That fixed it (7b5399057b2b6539e31c41d7eff3f3f2d4eea920). Tested in non-VS GUI with both Jack and RtAudio backends. There should be a beta-2 as soon as the above comment on AudioInterface.h gets sorted out.

src/AudioInterface.h
psiborg112: I probably didn't need to. It was really only in RtAudioInterface and JackAudioInterface that it made an actual difference, but it does seem to make sense for consistency and clarity that the constructor parameter order for the base class should match that used in the derived classes.

And the change was to get rid of an overloaded constructor in the derived classes that was needlessly complicating things. It just ran the original constructor but with a nullptr as the jacktrip pointer. (Because it's being run directly from vsAudioInterface.) This constructor also set processWithNetwork to false without any documentation of this. Which means we had two constructors which look like they do the same thing (just with effectively a default nullptr being passed to the parent class) but which actually have different default behaviours. That's just asking for things to break further down the line.

ntonnaett: Yeah, that sounds bad. To me it felt right to have the jacktrip pointer at pole position, but I don't care. Could you remove the redundant \param? Didn't test it myself, but sounds like it's well enough tested that we can merge this.

psiborg112: Yeah, I kind of agree on that. But it's not uncommon with the Qt classes for the parent pointer to be one of the final arguments, so that at least fits with that kind of paradigm. Was wondering what you were on about with the redundant param and then realised I left the first one in there. Will fix that now. (Thanks for spotting that.)

ntonnaett: Good to know. Seeing `this` as the first (or actually second?) argument in this context is confusing too. So maybe it's not that bad of an idea to put it in the end.

ntonnaett (approved):

PR #949, opened 2023-03-06
Fix updater release scripts

mattahorton:
The `github.event_name = 'release'` check was failing because we never actually run our actions in response to a release being published. This fixes that problem. 

We _may_ reconsider enabling the `deploy_gh` job on tagged commits and instead run it only on releases, but I don't have enough perspective as to what that would break.

### DIFF
--- a/.github/workflows/jacktrip.yml
+++ b/.github/workflows/jacktrip.yml
@@ -694,7 +694,7 @@ jobs:
           path: ${{ env.BUILD_PATH }}/${{ matrix.installer-path }}
   # release - list of files uploaded to GH release is specified in the *upload* step
   deploy_gh:
-    if: startsWith(github.ref, 'refs/tags/') # run on tagged commits
+    if: startsWith(github.ref, 'refs/tags/') && github.event_name == 'release' # run on releases only
     needs: [build, sign]
     runs-on: ubuntu-latest
     name: 'deploy release'
--- a/.github/workflows/jacktrip.yml
+++ b/.github/workflows/jacktrip.yml
@@ -713,27 +713,6 @@ jobs:
           draft: true
         env:
           GITHUB_TOKEN: ${{ secrets.GITHUB_TOKEN }}
-  release_test:
-    if:  github.event_name == 'release'
-    runs-on: ubuntu-latest
-    name: 'test release if statement'
-    steps:
-      - name: print
-        run: echo "release if passed"
-  tag_test:
-    if:  startsWith(github.ref, 'refs/tags/')
-    runs-on: ubuntu-latest
-    name: 'test tag if statement'
-    steps:
-      - name: print
-        run: echo "tag if passed"
-  secrets_test:
-    if:  needs.check-secrets.outputs.should_run == 'true'
-    runs-on: ubuntu-latest
-    name: 'test secrets if statement'
-    steps:
-      - name: print
-        run: echo "secrets if passed"
   deploy_jtl:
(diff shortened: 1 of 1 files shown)


nwang92 (approved):

dyfer:
Does this mean that we will have 2 (3?) jobs running concurrently for the release? one from the pushed tag, one triggered by the release itself, (on top of one that was likely triggered by the latest push before tagging)? Would the artifacts be uploaded twice to the release? (I assume they'd just be overridden, so maybe that's not a practical issue, but something to consider I guess)

Maybe what you have here is the cleanest, but one could also consider running this on all tags in the JTL repo:
```yaml
deploy_jtl:
    if:  startsWith(github.ref, 'refs/tags/') && needs.check-secrets.outputs.should_run == 'true'
```

mattahorton:
> Does this mean that we will have 2 (3?) jobs running concurrently for the release? one from the pushed tag, one triggered by the release itself, (on top of one that was likely triggered by the latest push before tagging)? Would the artifacts be uploaded twice to the release? (I assume they'd just be overridden, so maybe that's not a practical issue, but something to consider I guess)
> 
> Maybe what you have here is the cleanest, but one could also consider running this on all tags in the JTL repo/when secrets are available:
> 
> ```yaml
> deploy_jtl:
>     if:  startsWith(github.ref, 'refs/tags/') && needs.check-secrets.outputs.should_run == 'true'
> ```

Can't do that because any created tags would get uploaded to JTL servers and pushed to the updater.

But the concerns you brought up are the ones I share.

mattahorton:
I _do_ know that the artifacts will just get overwritten and will have the same bits as the other build. 

We could limit the existing tag-based scripts to only run if it's _not_ a release?

dyfer:
> Can't do that because any created tags would get uploaded to JTL servers and pushed to the updater.

Though that only happens when secrets are available (== on JTL repo) and tags typically result in releases... but of course it's possible to push a non-release tag, so I understand this is too risky.

> I _do_ know that the artifacts will just get overwritten and will have the same bits as the other build.

Good to know.

> We could limit the existing tag-based scripts to only run if it's _not_ a release?

Yeah, we could do that, OR we could disable running on tags altogether and leave running on releases only (and pushes do dev/main as before). After all, we mostly care for running on a release. And the workflow can be triggered for any branch manually anyway.

mattahorton:
> I'm suspecting that one practical outcome of the last suggestion (don't run on tags) would be that releases would not be created automatically if a tag is pushed from e.g. command line. This is just a speculation though and I'm not sure if this is important in our case.

That is interesting. I haven't considered that use case.

mattahorton:
@dyfer I changed Github release deployment to only run on releases.

dyfer:
So, does this mean that release can only be created using GitHub UI, and not by simply pushing a tag?

I don't oppose this, but I want it to be clear. Additionaly, it would be nice to document it _somewhere_ (I don't think the release process is documented at all currently), not necessarily as part of this PR.

@ntonnaett as you've also worked on GHA, do you have any thoughts on this?

Again, I'm fine with the change, but I think it's important that we are aware what it does.

mikedickey (approved):

mattahorton:
> So, does this mean that release can only be created using GitHub UI, and not by simply pushing a tag?
> 
> I don't oppose this, but I want it to be clear. Additionaly, it would be nice to document it _somewhere_ (I don't think the release process is documented at all currently), not necessarily as part of this PR.
> 
> @ntonnaett as you've also worked on GHA, do you have any thoughts on this?
> 
> Again, I'm fine with the change, but I think it's important that we are aware what it does.

yes

PR #951, opened 2023-03-08
Reinstate backend warning in the classic GUI

psiborg112:
While it's great that we have a choice of audio backends now, and users can run the program without JACK, our behaviour of silently falling back to the RtAudio backend when JACK isn't found is causing a lot of confusion amongst users outside the VirtualStudio community. (As I predicted when we first discussed this in #477.) It's gotten to the point that people are mistakenly under the impression that JackTrip installs JACK, and are having to rely on the cryptic WEAK-Jack messages the library generates to do any sort of trouble shooting, unaware that some of the output they're seeing means that JACK isn't properly installed. (See #941.) This pull request reinstates an error message that has remained in my builds.

When JACK isn't found, the user will be confronted with the following warning:

> An installation of JACK was not found. JackTrip will still run using a different audio backend (RtAudio) but some more advanced features, like the ability to run your own hub server, will not be available.
> 
> (If you install JACK at a later stage, these features will automatically be re-enabled.)

They're also given the option to silence this warning going forward.

### DIFF
--- a/src/gui/qjacktrip.cpp
+++ b/src/gui/qjacktrip.cpp
@@ -333,9 +333,7 @@ QJackTrip::QJackTrip(Settings* settings, bool suppressCommandlineWarning, QWidge
     // Check if Jack is actually available
     if (have_libjack() != 0) {
 #ifdef RT_AUDIO
-#ifdef PSI
         bool usingRtAudioAlready = m_ui->backendComboBox->currentIndex() == 1;
-#endif  // PSI
         m_ui->backendComboBox->setCurrentIndex(1);
         m_ui->backendComboBox->setEnabled(false);
         m_ui->backendLabel->setEnabled(false);
@@ -349,7 +347,6 @@ QJackTrip::QJackTrip(Settings* settings, bool suppressCommandlineWarning, QWidge
             "JACK was not found. This means that only the RtAudio backend is available "
             "and that JackTrip cannot be run in hub server mode.");
 
-#ifdef PSI
         QSettings settings;
         settings.beginGroup(QStringLiteral("Audio"));
         if (!settings.value(QStringLiteral("HideJackWarning"), false).toBool()) {
@@ -385,7 +382,6 @@ QJackTrip::QJackTrip(Settings* settings, bool suppressCommandlineWarning, QWidge
             settings.setValue(QStringLiteral("UsingFallback"), false);
         }
         settings.endGroup();
-#endif  // PSI
 #else   // RT_AUDIO
         QMessageBox msgBox;
         msgBox.setText(


cchafe:
One case to consider is a linux pipewire machine. No explicit jack install but it does run with jack because of pipewire's jack provisioning. That's a bit of an outlier because there is no jackd present but the jack backend is functional and utilities like jack_bufsize are present and working. It isn't rtaudio. What mechanism is used to detect that jack isn't present?

mattahorton:
I'm fine with this.

OConnorStP:
I like the message for the scenarios it covers.  This seems like a good example of providing "documentation when it's most needed."  Is there a way to sense scenarios like Chris's Pipewire case into the error processing, thus providing even better documentation-when-it's-most-needed?

ntonnaett:
As we only check for a JACK installation, this should also work for the Pipewire case.

psiborg112:
The weakjack library is used to determine whether JACK is available or not. I'm not familiar with pipewire, but from my understanding it provides JACK emulation, which should mean that weakjack detects it. Actually the easy way to know will be whether you've been able to run the GUI before in JACK mode on a pipewire only install - the detection code has always been there. It will just have silently switched you to RtAudio if JACK wasn't detected.

ntonnaett:
Wait. We shouldn't depend on weakjack on Linux.

dyfer:
> Wait. We shouldn't depend on weakjack on Linux.

At least not currently. 
If we were to have backend switching with Jack / PulseAudio / Alsa (?) etc support, then weak linking would make sense... but not yet.

dyfer:
> > Wait. We shouldn't depend on weakjack on Linux.
> 
> At least not currently, as we're not building JT with RtAudio on Linux. But in the future, if/when we have backend support for PulseAudio / Alsa etc, weak linking with Jack would make sense...

ntonnaett:
You don't need weak linking for that.

dyfer:
No, but it makes sense from the user perspective. One shouldn't need to have JACK installed if one were to run JT with PulseAudio or Alsa backend. IIUC, without weak linking JT wouldn't start if Jack is not installed. 

AFAIU this case would be valid only for OSs without pipewire.

cchafe:
I built today's test on a fedora machine with 
qmake-qt5 -config rtaudio
Pipewire machine, jack not installed, and the GUI offers both Jack and RtAudio backends.
Both work.

dyfer:
> Pipewire machine, jack not installed, and the GUI offers both Jack and RtAudio backends.
> Both work.

Thanks for testing!
This is what I'd expect, with Pipewire installed. 
Do you know which "intermediate" backend does the RtAudio use in this case? Is it ALSA?

It would be more tricky to support OSs without pipewire, where ideally we'd need weak linking for all the backends and load them on demand.

psiborg112:
To clarify as well, the detection code is within a USE_WEAK_JACK ifdef. It will never be tripped if it's been built without weak jack, because it'll be assumed that JACK is a requirement for the program to run.

mikedickey (approved):

PR #952, opened 2023-03-09
Include rtaudio in static Linux build with GUI

mikedickey:


psiborg112:
Maybe this should be added as an additional configuration rather than a change to the existing one? Otherwise there are no GUI enabled builds that don't also have RtAudio. (And it'll be one less configuration we'll be testing for.)

mikedickey:
> Maybe this should be added as an additional configuration rather than a change to the existing one? Otherwise there are no GUI enabled builds that don't also have RtAudio. (And it'll be one less configuration we'll be testing for.)

We could so long as there is a need for it. What is the use case for having a build without support for RTaudio?

ntonnaett:
The flathub build is without support for RtAudio. Pipewire is the only backend that is supported.

psiborg112:
Thanks @ntonnaett. The use case is to test that nothing that we've changed in the code breaks that configuration for those who want to build it in that fashion.

dyfer:
This also means that the build will be linked against any audio libraries that RtAudio is linking against, but I'm not sure what's the latest on the backends that are used by RtAudio on Linux. Have this been investigated? Which additional libraries would we link against and would that cause issues for people who want to use the static build?

dyfer:
(If we indeed agree on this change, I'd be happy to add a configuration to test non-rtaudio build; I think it would be better to have it _not_ use static qt to avoid additional long builds, when cache is not available)

psiborg112:
Yes, a non static test build would be great.

mikedickey:
I'll update this to revert the change and add a new config for static with jack+rtaudio

mikedickey:
@mattahorton please note possible conflict with the CD automation you are working on for releases. We would like to use static with Jack+RtAudio for Linux downloads from Virtual Studio

dyfer:
> I'll update this to revert the change and add a new config for static with jack+rtaudio

My suggestion was to add a non-static jack-only build for testing, as opposed to have two static builds.

Do you have any insights on the linking questions I posted earlier?

mikedickey:
@dyfer you are correct that there are more shared dependencies introduced by static linking rtaudio. In particular:

	libasound.so.2 => /lib64/libasound.so.2 (0x00007fd253baa000)
	libpulse-simple.so.0 => /lib64/libpulse-simple.so.0 (0x00007fd253ba3000)
	libpulse.so.0 => /lib64/libpulse.so.0 (0x00007fd253b4d000)
	libpulsecommon-15.0.so => /usr/lib64/pulseaudio/libpulsecommon-15.0.so (0x00007fd251573000)
	libsndfile.so.1 => /lib64/libsndfile.so.1 (0x00007fd250cbd000)
	libasyncns.so.0 => /lib64/libasyncns.so.0 (0x00007fd251402000)
	libgsm.so.1 => /lib64/libgsm.so.1 (0x00007fd250a12000)
	libFLAC.so.8 => /lib64/libFLAC.so.8 (0x00007fd2509b7000)
	libvorbis.so.0 => /lib64/libvorbis.so.0 (0x00007fd250985000)
	libvorbisenc.so.2 => /lib64/libvorbisenc.so.2 (0x00007fd2508da000)
	libopus.so.0 => /lib64/libopus.so.0 (0x00007fd25087e000)
	libogg.so.0 => /lib64/libogg.so.0 (0x00007fd250873000)
	libmpg123.so.0 => /lib64/libmpg123.so.0 (0x00007fd250815000)
	libmp3lame.so.0 => /lib64/libmp3lame.so.0 (0x00007fd25079d000)

mikedickey:
Updated to add a shared build without rtaudio, for testing

dyfer:
Thanks for the list @mikedickey 
That means that the binary will no longer run on OSs without PulseAudio (and without PipeWire, which, IIUC, provides PulseAudio backend). Is everyone okay with that? After a quick web search, it seems unlikely to that there are mainstream distros without PulseAudio (or PipeWire), but maybe full time Linux users have a better idea here.

I'm not sure what all the libsndfile dependencies are doing there, but I guess it's okay...?

mikedickey:
seems to work on both ubuntu and fedora, which i think is going to cover the vast majority.

dyfer (approved):
I think this is fine, i.e. it still runs on my Ubuntu system.

However, I have not tested RtAudio performance.

ntonnaett:
I wouldn't be worried about the availability of pulseaudio but all these codec libraries and libsndfile might be missing.

mikedickey:
@OConnorStP these builds don't run on RPI anyway due to chipsets (x64 vs arm), so this change will not impact RPI in any way. We currently don't create any RPI builds ourselves.

OConnorStP:
apt install Jacktrip...  works on an Rpi.  it's 1.3, but it works.  so somebody's doing a build.

dyfer:
@OConnorStP packages in managers like `apt` are typically maintained (built) by maintainers of the given distribution. In particular, I'm not aware that any JT developers are actively involved in publishing .deb or .rpm packages. The closest we are to this is in the Flatpak world, but I don't think our Flatpak packages are published anywhere (?)

ntonnaett:
I try to keep flatpak manifests in our repo and in Flathub in sync. So you could say our Flatpak packages are published on Flathub. They get even built for aarch64.

https://flathub.org/apps/details/org.jacktrip.JackTrip

dyfer:
Cool, thanks for the info!

OConnorStP:
I agree -- the Flathub build is very cool.  It loaded right up on Debian 11 and the GUI ran fine (first time in a long time for that).  

But here's an interesting thing.  I then went on to launch Ardour (which in turn launches Jack).  I then fired the Flathub Jacktrip as a Hub Server from the command line as I normally would.  Again, it ran fine. 

But it didn't see the running instance of jackd when I came in as a Hub client -- instead it came back with the "perhaps the JACK server isn't running?" error and quit.  But Jack was running.  Any ideas about that?   I've attached a screen grab.

Flatpak install

dyfer:
Let's continue this discussion in #955

PR #959, opened 2023-03-13
Fix Compiler Warnings When Building Without RtAudio

dhing1024:
Fixes some compiler warnings in builds not using RtAudio. To test and change flags with meson build system:

`meson configure -Drtaudio=disabled` or ``meson configure -Drtaudio=enabled` (in your build directory)

### DIFF
--- a/src/gui/virtualstudio.cpp
+++ b/src/gui/virtualstudio.cpp
@@ -471,7 +471,6 @@ int VirtualStudio::inputMixMode()
 }
 #endif
 
-
 QString VirtualStudio::outputDevice()
 {
 #ifdef RT_AUDIO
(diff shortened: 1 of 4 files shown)


ntonnaett:
Thanks for taking the time fixing those warnings! It seems you found a cleaner way than what I would have done. Do you also want to fix the warnings from vsAudioInterface.cpp with this PR?

dhing1024:
I can add them in. I think it shouldn't be too much different from the main vs class

ntonnaett:
After the build has no warnings anymore, I can add a meson build to our checks with rtaudio disabled that errors out for warnings the same as we do for the current build.

ntonnaett (approved):

PR #963, opened 2023-03-15
Add Warnings to Connected Screen

dhing1024:
Adds warnings to the connected screen if devices are low latency (the screenshot below is only for visualization purposes; the devices I have wouldn't actually produce this warning)

Screenshot 2023-03-15 at 4 02 19 PM


mikedickey (approved):
tested and works as expected for me on windows

PR #967, opened 2023-03-17
Update version and changelog for 1.8.0 release

mattahorton:


docs/changelog.yml
```diff
@@ -6,16 +6,22 @@
   - (added) when using the classic GUI, command line options are now parsed
   - (added) VS mode - Selecting and configuring device channels
   - (added) Classic mode - Warning for machines without JACK installed
+  - (added) VS mode - high latency warning for non-ASIO devices
+  - (added) Meson rtaudio build in GHA
```
ntonnaett, line 10: A Meson build with rtaudio disabled was added.

ntonnaett:
`(updated) Flathub - improve latency by defaulting to 256 samples buffer size` would be nice. Although it didn't happen in this repo, Flathub users see this changelog when updating.

docs/changelog.yml
mattahorton: gotcha. will update. i misread

dhing1024 (approved):

PR #971, opened 2023-03-18
Allow IP check to work with IPv6 only

psiborg112:
Realised when looking over the IP check code that there's an assumption that an IPv4 address will always be available. This allows the code to report an address when on an IPv6 only network.


mikedickey (approved):
code lgtm and it works for my ipv4 address

PR #972, opened 2023-03-19
make buffer size and sample rate settable for Pipewire

ntonnaett:
The command line flags for setting the buffer size and sample rate were only available to the RtAudio backend. Making these flags available to all backends, let's us set the PIPEWIRE_LATENCY environment variable before the JACK client gets setup.

PIPEWIRE_LATENCY needs buffer size and sample rate. Getting the default values for these from JACK requires a JACK client. Instead of opening an ephemeral JACK client, ~~we require setting both buffer size (-F) and sample rate (-T) for setting PIPEWIRE_LATENCY instead.~~ we set the undefined value to the JackTrip's own default value.

PS: This also fixes -T segfaulting.

### DIFF
--- a/src/Settings.cpp
+++ b/src/Settings.cpp
@@ -155,10 +155,11 @@ void Settings::parseInput(int argc, char** argv)
         {"clientname", required_argument, NULL, 'J'},  // Run in JamLink mode
         {"remotename", required_argument, NULL, 'K'},  // Client name on hub server
         {"appendthreadid", no_argument, NULL,
-         OPT_APPENDTHREADID},  // Append thread id to client names
+         OPT_APPENDTHREADID},                       // Append thread id to client names
+        {"srate", required_argument, NULL, 'T'},    // Set Sample Rate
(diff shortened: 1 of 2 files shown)


cchafe:
Delightful! And it now works both ways: with or without -T
Ready for a merge?

ntonnaett:
Yes, let's merge. I'm really happy how minimal invasive this is. And I'm impressed how good Pipewire became. 32 samples buffer size has no glitches for me.

cchafe (approved):

dyfer:
Great!
Sorry I didn't have a chance to test this.
Just curious, did that by any chance work without the env variable, just with setting the buffer size in JACK backend? or is it not happening through the JACK api at all, just with the env var?

cchafe:
It's 100% env var and it's a super simple solution -- the right thing for any audio backend is able to plumb to Pipewire , Jack, RtAudio, RtAudioALSA. I can run it down to 16 FPP (which seems to be an improvement to Pipewire which used to have a low end of 64).

PR #978, opened 2023-03-25
mkdocs: make docs look jacktrippy

ntonnaett:
Small changes to make our docs look a little bit more like they belong to jacktrip. This also fixes the search.


dyfer (approved):
This looks nice!

PR #979, opened 2023-03-27
Fix Windows deeplink when open

mattahorton:
So the std::cout `VirtualStudio::show()` was somehow colliding with the socket and instead of the deeplink URL, the characters "SSL Lib" or "SSL L" were going across the local socket that checks for the existing instance of the app.

I switched it to qDebug since it's already in `virtualstudio.cpp`. Solved the problem.

### DIFF
--- a/src/gui/virtualstudio.cpp
+++ b/src/gui/virtualstudio.cpp
@@ -324,7 +324,7 @@ void VirtualStudio::show()
     if (m_checkSsl) {
         // Check our available SSL version
         QString sslVersion = QSslSocket::sslLibraryVersionString();
-        std::cout << "SSL Library: " << sslVersion.toStdString() << std::endl;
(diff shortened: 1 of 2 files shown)


nwang92 (approved):

mikedickey (approved):

psiborg112:
While this might fix the problem, this isn't really a solution - the code is one added std::cout from this bug being triggered again. This either needs to be documented somewhere in the code, or preferably (given the asynchronous nature of the operation and the fact that this is not isolated to a single portion of the code) find a way to avoid std::cout getting routed to the local server.

psiborg112:
I guess the first question would be whether this was happening before the cleanup in main.cpp, and if not, what about the encapsulation has changed the interaction between QLocalServer and std::cout?

PR #981, opened 2023-03-27
AudioSettings should show the currently selected device

nwang92:
I'm seeing that virtualstudio.cpp is reporting the correct device names and indices, but they are somehow not making it into the QML.

Moving these funcs into AudioSettings.qml seem to fix it.


dhing1024 (approved):
was gonna look into this today, along with the other bug I'm working on. but looks like you beat me to it!

PR #982, opened 2023-03-28
vsInit cleanup

psiborg112:
Get rid of complex nested lambdas so that the code is easier to read. Also adds a warning related to #979 and updates authorship of some files.

@mattahorton, I'm going to need your help testing this. (Deeplink functionality/instance checking on windows.) Cleaning up this code will hopefully help us to find and fix the root cause of #979.


mattahorton:
Thanks, Aaron. I'll test it out

mattahorton (approved):

PR #983, opened 2023-03-28
Fix Refresh Button on Windows

dhing1024:
I think this fixes the issue with refresh button on Windows. The issue I think stems from the fact that there are some situations with `ASIOStart` and `ASIOStop` in `externals/rtaudio/include/asio.cpp` getting called at the wrong times. I was running into an issue where refreshing devices would stop the audio stream from the driver (without restarting it). Then if I tried to do something like switch devices, it would try to stop the stream again before restarting it and causing a crash. At least that's my best guess as to why there's a bug, and I think this fixes it by making sure to restart the audio stream when refreshing devices.

As an aside, I think something about my ASIO setup on Windows isn't quite working correctly, so I think I'll need someone to test connecting to a studio and seeing if audio still works there normally. This might also have some performance improvements on Windows.


mikedickey (approved):
This looks good to me. I tested the build from actions on my Windows machine and it fixes the refresh button problem for me. I was also able to connect to studios and get sound. Seems good!

mattahorton (approved):

PR #989, opened 2023-04-01
bufstrategy 4, runs PLC directly in audio callback (the original way)…

cchafe:
bufstrategy 4, runs PLC directly in audio callback (the original way) rather bufstrategy 3 where it runs in a workerThread (the current way)

The problem is windows does not play nice with our new separate workerThread computation of PLC. 
The symptom is continuous distortion when using bufStrategy 3.
A new bufStrategy 4 reverts to the earlier, original, non-workerThread methods.
This fixes PLC in windows, and might fix similar problems encountered with macos, or some other.
So... don't use --bufstrategy 3 in a windows hub client, use --bufstrategy 4
(everything else about the CLI and GUI arguments works the same)

The fix involves overloading Regulator::pullPacket to have it both ways.

Regulator::pullPacket() does the PLC heavy work.
For bufStrategy 3 it is called from the workerThread slot RegulatorWorker::pullPacket()
For bufStrategy 4 it is called from the overloaded Regulator::pullPacket(int8_t* buf) 

Total explanation:

The audio callback calls **receiveNetworkPacket** which then calls
**readSlotNonBlocking(ptrToReadSlot)**
    if bufStrategy 3
      returns the result from workerThread
    if bufStrategy 4
      calls pullPacket(ptrToReadSlot)
        which calls pullPacket() and then returns the result directly      

**receiveNetworkPacket** then, if in bufStrategy 3 mode, emits 
  signalReceivedNetworkPacket which calls pullPacket() in workerThread

Interface changes are in 
Settings and qjacktrip.ui which has two items now instead of just the first one
**3 (experimental - in own thread) 
4 (experimental - in callback)**

or CLI:

jacktrip -C jackloop128.stanford.edu **--bufstrategy 3** -q auto3 --udprt
jacktrip -C jackloop128.stanford.edu **--bufstrategy 4** -q auto3 --udprt

### DIFF
--- a/src/gui/qjacktrip.ui
+++ b/src/gui/qjacktrip.ui
@@ -1906,7 +1906,7 @@ To connect to a hub server you need to run as a hub client.</string>
      <x>0</x>
      <y>0</y>
      <width>409</width>
-     <height>27</height>
+     <height>30</height>
     </rect>
    </property>
   </widget>
@@ -1986,7 +1986,7 @@ To connect to a hub server you need to run as a hub client.</string>
   <tabstop>disconnectScriptBrowse</tabstop>
  </tabstops>
  <resources>
-  <include location="qjacktrip_novs.qrc"/>
+  <include location="qjacktrip.qrc"/>
  </resources>
  <connections/>
 </ui>
--- a/src/gui/qjacktrip.ui
+++ b/src/gui/qjacktrip.ui
@@ -7,14 +7,14 @@
     <x>0</x>
     <y>0</y>
     <width>409</width>
-    <height>997</height>
+    <height>961</height>
    </rect>
   </property>
   <property name="windowTitle">
    <string>JackTrip</string>
   </property>
   <property name="windowIcon">
-   <iconset resource="qjacktrip_novs.qrc">
+   <iconset resource="qjacktrip.qrc">
     <normaloff>:/qjacktrip/icon.png</normaloff>:/qjacktrip/icon.png</iconset>
   </property>
   <widget class="QWidget" name="centralWidget">
@@ -109,7 +109,7 @@
        </size>
       </property>
       <property name="currentIndex">
-       <number>3</number>
+       <number>0</number>
       </property>
       <property name="tabsClosable">
        <bool>false</bool>
--- a/src/gui/qjacktrip.ui
+++ b/src/gui/qjacktrip.ui
@@ -7,14 +7,14 @@
     <x>0</x>
     <y>0</y>
     <width>409</width>
-    <height>961</height>
+    <height>997</height>
    </rect>
   </property>
   <property name="windowTitle">
    <string>JackTrip</string>
   </property>
   <property name="windowIcon">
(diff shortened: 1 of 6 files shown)


src/gui/qjacktrip.ui
```diff
@@ -1408,7 +1408,12 @@ for better quality at the expense of latency.</string>
           </item>
           <item>
            <property name="text">
-            <string>3 (experimental)</string>
+            <string>3 (experimental - in own thread) </string>
+           </property>
+          </item>
+          <item>
+           <property name="text">
+            <string>4 (experimental - in callback)</string>
```
psiborg112, line 1416: This should be the only change needed in the .ui file. (Not sure how some of the others crept in.) If you're able to revert all but this one that would be great.

psiborg112 (changes requested):

src/gui/qjacktrip.ui
cchafe: those were side effects of me editing using designer, please check if I reverted  'em ok
lesson learned!

psiborg112: I use designer as well, but usually it behaves itself a bit better. There's just another couple at line 1909 and 1989. (Otherwise it's looking good.)

cchafe: I wonder if all these nits might be due to display differences or me touching the designer objects awkwardly, I missed those last ones down there at the bottom of meld -- fixed now

cchafe: I tested the installer artifact under win10 and it works as intended

ntonnaett:
I'm wondering what the reason for the behavior on Windows is. Maybe it's just about thread priorities?

cchafe:
I played around with thread priorities at some length until I gave up. Using HackTrip for the framework there was no improvement. I suspect the same for JackTrip.

psiborg112:
Looks good to me now.

psiborg112 (approved):

ntonnaett:
This should probably merged after 1.8.1 is released.

mikedickey:
I tested using classic mode on both Mac and Windows, simulating loss by by running `tc qdisc` on the hub server.

Running in-thread (4) on Windows sounds much better to me.

It's less of a clear distinction for me on OSX, likely because it does a better job with threading. For some reason they both sound kind of bad for the first few seconds, and then get better. After that I think it sounds slightly better on OSX.

The only thing I wonder is if this is more of a client versus hub server thing, in that perhaps we should always run in-thread (4) when connecting as a client, and using a separate worker (3) when running as a hub server? IIRC, it only takes a few clients for hub server to fail with the in-thread model (4). If so, then I wonder if it may be better to just have it act differently depending on what mode jacktrip is running in, versus introducing a new buffer strategy?

I also really want to try out a single regulator thread, which is set to run as high priority. I think that may work better universally. But that is a topic for another PR..

mikedickey (approved):

PR #991, opened 2023-04-03
Update vcpkg

mattahorton:
Updates vcpkg in github actions to a commit with the msys2 fix.


dyfer:
Is there a reason not to do it inside the "install dependencies for Windows" step, [like we do with rtaudio](https://github.com/jacktrip/jacktrip/blob/688bc255681c121bc479b19042efddb72d0b5f02/.github/workflows/jacktrip.yml#L361-L363)?

dyfer:
I guess I was too late :/

PR #992, opened 2023-04-03
Use realtime priority for regulator worker thread

mikedickey:
also adding statistics on worker thread underrun when -I (iostats)

### DIFF
--- a/src/JackTrip.cpp
+++ b/src/JackTrip.cpp
@@ -675,6 +675,8 @@ void JackTrip::completeConnection()
         QObject::connect(this, &JackTrip::signalReceivedNetworkPacket, workerPtr,
                          &RegulatorWorker::pullPacket, Qt::QueuedConnection);
         mRegulatorThreadPtr->start();
+        QObject::connect(this, &JackTrip::signalAudioStarted, workerPtr,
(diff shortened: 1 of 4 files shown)


mikedickey:
I am seeing that with default scheduling priority, PLC is falling behind with only a few clients. Bumping its priority is producing MUCH better results for me.

I still think it's probably better to run inside the jack callback (4) for clients, but for hubserver this is required.

mikedickey:
FWIW i did also try using a single regulator thread, but it does not work very well at all. After a few clients with sustained (1%) loss, it starts to fall behind and is never able to catch back up. https://github.com/jacktrip/jacktrip/commit/bc3abd85d463da3aa9248a4824a9eb51165da503

nwang92 (approved):

ntonnaett:
Maybe we want to do this more fine-grained. It looks like this uses the same priority as we use for the udp thread. How does this actually interact with the realtime udp thread option?

mikedickey:
> Maybe we want to do this more fine-grained. It looks like this uses the same priority as we use for the udp thread. How does this actually interact with the realtime udp thread option?

Maybe, but realtime udp threads don't seem to scale well in hubserver mode anyway. At least as things currently stand, you should only use realtime udp threads with bufstrategy 4.

PR #993, opened 2023-04-04
In virtual studio mode, always use bufferStrategy 4 for loss concealment

mikedickey:
bufferStrategy 3 is better for hubserver, while 4 is better for clients

An alternative approach here would be to just eliminate 4 and make 3 behave differently depending on which mode jacktrip is running

### DIFF
--- a/src/gui/virtualstudio.cpp
+++ b/src/gui/virtualstudio.cpp
@@ -1492,16 +1492,21 @@ void VirtualStudio::completeConnection()
 #else
         int inputMixMode = -1;
 #endif
-        JackTrip* jackTrip = m_device->initJackTrip(
-            m_useRtAudio, input, output,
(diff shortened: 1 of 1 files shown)


nwang92 (approved):

ntonnaett:
IMHO eliminating strategy 4 sounds better. It isn't really a buffer strategy in the first place.

mikedickey:
> IMHO eliminating strategy 4 sounds better. It isn't really a buffer strategy in the first place.

I tend to agree with you, but having the two strategies does help for testing, at least for now. 

I think this change is important so long as we have the two in place. I'm going to merge this for now, and modify my other PR that merges the two strategies so that it reverts it.

See https://github.com/jacktrip/jacktrip/pull/994

Issue #938, opened 2023-02-27
Loss concealment + small buffer sizes leads to skipping in the broadcast queue


nwang92:
I've only been able to reproduce this on JackTrip Radio, but if you start a studio with loss concealment and a buffer size of <= 64, the radio playback sounds like it's skipping.

The radio playback is built from the broadcast ports of each client, so I suspect the problem maybe be how loss concealment w/ low buffer size impacts broadcast queues?

Note this doesn't happen with using buffer size of <= 64 and other buffer strategy settings.

nwang92:
I took a few recordings from jack to confirm the behavior. Example CLI reference:
```bash
$ jack_capture --port cd24c603-6edf-4d9c-8f90-1b12407b3157:receive_1 --port cd24c603-6edf-4d9c-8f90-1b12407b3157:receive_2
```

I attached the recordings here, where `jack_capture_normal.wav` was recorded from normal ports and `jack_capture_broadcast.wav` was recorded from the broadcast ports
[jack_capture.zip](https://github.com/jacktrip/jacktrip/files/10853203/jack_capture.zip)

mikedickey:
Fixed in 1.9.0

PR #996, opened 2023-04-06
moving push to broadcast ringbuffer into pushPacket()

mikedickey:
so that shimFPP can do its buffer size adjustment magic

### DIFF
--- a/src/Regulator.cpp
+++ b/src/Regulator.cpp
@@ -372,6 +372,8 @@ void Regulator::shimFPP(const int8_t* buf, int len, int seq_num)
 //*******************************************************************************
 void Regulator::pushPacket(const int8_t* buf, int seq_num)
 {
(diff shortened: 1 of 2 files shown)


mikedickey:
The bug here is that if A & B connect where the buffer size for A > B, and B is using PLC (bufferstrategy of 3 or 4), the broadcast output for B will be skipping etc.

This seems to fix it for me. I tested all three combinations:
* A & B have the same buffer size
* A > B
* B > A

I don't really follow all the FPP adjustment logic though, and it seems to simple. So @cchafe I would love your eyes on this before merging.

nwang92 (approved):

mikedickey:
Fixes https://github.com/jacktrip/jacktrip/issues/938

cchafe:
Good one!

PR #1000, opened 2023-04-19
Bump Xcode from 11.7 to 13.2.1 on OSX 11

mikedickey:
Available versions here https://github.com/actions/runner-images/blob/main/images/macos/macos-11-Readme.md

11.7 is quite old, and seems to be creating bad bytecode for PLC


ntonnaett:
What is the impact of the minimum supported macOS host version?

psiborg112:
Setting it with an environment variable would be the other way to go about it. Pretty sure MACOSX_DEPLOYMENT_TARGET is what you're after.

PR #1005, opened 2023-04-24
Fix openssl v3 compatibility on Linux via qt bugfix patch

mikedickey:
from https://bugreports.qt.io/browse/QTBUG-103820


mikedickey:
Just applying the patch alone doesn't seem to solve the issue..

dyfer:
Hm. It seems it's not straightforward to install openssl v3 on Ubuntu 20.04 (and IIUC Virtual Studio is not supported with openssl v1, correct?). In that case maybe the static build should only support Ubuntu 22.04 and above? you can try changing the build system to 22.04 to check if that fixes the issue.

mikedickey:
> Hm. It seems it's not straightforward to install openssl v3 on Ubuntu 20.04 (and IIUC Virtual Studio is not supported with openssl v1, correct?). In that case maybe the static build should only support Ubuntu 22.04 and above? you can try changing the build system to 22.04 to check if that fixes the issue.

Testing locally I was able to build static openssl on 20.04 and get qt to use that, and confirmed that it works on 22.04.. Just need to iron out any issues translating this into github actions..

dyfer:
Just a note regarding caching, once this PR is good to go: if there's more than a week planned between merging this and making another release, I'd suggest to merge it to `main` first (and then `main` -> `dev`). That way new PRs should have access to the new cache and thus avoid long build times.

mikedickey:
> Just a note regarding caching, once this PR is good to go: if there's more than a week planned between merging this and making another release, I'd suggest to merge it to `main` first (and then `main` -> `dev`). That way new PRs should have access to the new cache and thus avoid long build times.

I'd like to get it included in 1.9.0 (next week?)

mikedickey:
Confirmed working on Ubuntu 22.04 and even on ChromeOS / Linux. Sadly, ChromeOS uses a VM and "VirtIO" sound driver which introduce _terrible_ latency. I'm getting better latency when connecting audio via browser (WebRTC) on ChromeOS versus running the JackTrip app.

dyfer:
Thanks, it seems to work as advertised, I tested on Ubuntu 20.04 and got

```
(...)
SSL Library:  "OpenSSL 3.1.0 14 Mar 2023"
```

Regarding the implementation:

> I'd like to get it included in 1.9.0 (next week?)

I still think it would be better to merge this into `main` first (probably needs to be rebased) so that the PRs between now and the release don't all rebuild Qt

> Add "build nomake" command which can be used to just generate Makefiles without calling "make". This makes it easier to run multi-process "make -j4"

This is minor... I think adding "nomake" option is fine/useful, but is this the cleanest way to go about multithreading? Maybe it would be better to add an argument for num of threads to the script args? And/or, add `nomake`, but change the default in the script to `make -j4` so that the script can be run as is in the CI? 

I believe this change also changes the make command on windows in the CI (the [script sets that](https://github.com/jacktrip/jacktrip/pull/1005/files#diff-44575cf5b28512d75644bf54a517dcef304ff809fd511747621b4d64f19aac66L125) at `mingw32-make`, while in this PR `make` is used, similarly to Linux/macOS). Although, since it seems to build fine then maybe that's okay...? I'm not sure.

mikedickey:
> I still think it would be better to merge this into `main` first (probably needs to be rebased) so that the PRs between now and the release don't all rebuild Qt

Why would GitHub be rebuilding qt for other PRs? Does it not preserve older cache entries?

> This is minor... I think adding "nomake" option is fine/useful, but is this the cleanest way to go about multithreading? Maybe it would be better to add an argument for num of threads to the script args? And/or, add `nomake`, but change the default in the script to `make -j4` so that the script can be run as is in the CI?

I too like this better than "nomake" and just replaced it with a "-j#" option.

dyfer:
> > I still think it would be better to merge this into `main` first (probably needs to be rebased) so that the PRs between now and the release don't all rebuild Qt
> 
> Why would GitHub be rebuilding qt for other PRs? Does it not preserve older cache entries?

Caches are scoped to each branch, BUT the cache from the default branch is available to all. Our default branch is `main`, so caches done in `dev` will _not_ be available to other branches and PRs.

> > This is minor... I think adding "nomake" option is fine/useful, but is this the cleanest way to go about multithreading? Maybe it would be better to add an argument for num of threads to the script args? And/or, add `nomake`, but change the default in the script to `make -j4` so that the script can be run as is in the CI?
> 
> I too like this better than "nomake" and just replaced it with a "-j#" option.

I like this, thanks!

mikedickey:
> > Why would GitHub be rebuilding qt for other PRs? Does it not preserve older cache entries?
> 
> Caches are scoped to each branch, BUT the cache from the default branch is available to all. Our default branch is `main`, so caches done in `dev` will _not_ be available to other branches and PRs.
> 
> EDIT we also "refresh" cache by running the CI weekly. This also happens on the default branch.

I don't think we can rebase this very cleanly because it also depends on the universal binary build changes and other things which have already merged to dev. How about we merge this to dev, and then immediately merge dev to main, even though we aren't technically doing a new release quite yet, to ensure all the PRs can use all of the recent qt updates?

dyfer:
> I don't think we can rebase this very cleanly because it also depends on the universal binary build changes and other things which have already merged to dev.

Ah, I see.

> How about we merge this to dev, and then immediately merge dev to main, even though we aren't technically doing a new release quite yet, to ensure all the PRs can use all of the recent qt updates?

Wouldn't that mess up the release process? We'd end up with "non-stable" state in the `main` branch. But since this will be for a short period of time, maybe that's okay.

Another idea would be to merge this into dev, and then make a PR cherry-picking only CI changes into main. But I'm not sure whether that would create problems when merging main back to dev after the release?

.github/workflows/jacktrip.yml
```diff
@@ -408,42 +411,66 @@ jobs:
           tar -xf qt.tar.xz
           # move qt to a known location
           mv qt-everywhere* $QT_SRC_PATH
-      - name: build static Qt on Linux / macOS
-        if: matrix.static-qt-version && steps.cache-static-qt.outputs.cache-hit != 'true' && runner.os != 'Windows'
+      - name: build static Qt on Linux
+        if: matrix.static-qt-version && steps.cache-static-qt.outputs.cache-hit != 'true' && matrix.name == 'Linux-x64-qmake-gcc-static'
         run: |
+          # Note that other static Linux runners will fail on first run. This is intentional to prevent two runners
+          # from trying to build and update the same cache item. Just re-run the failed jobs after cache is populated.
```
dyfer, line 418: I just noticed it and I like it! I was thinking about some solution like this. Thanks!

mikedickey:
> > How about we merge this to dev, and then immediately merge dev to main, even though we aren't technically doing a new release quite yet, to ensure all the PRs can use all of the recent qt updates?
> 
> That would work for caching, but wouldn't that mess up the release process? We'd end up with "non-stable" state in the `main` branch. But since this will be for a short period of time, maybe that's okay.
> 
> Another idea would be to merge this into dev, and then make a PR cherry-picking only CI changes into main. But I'm not sure whether that would create problems when merging main back to dev after the release?

I don't think it would have any practical impact, short of having a bug fix PR off main that we need publish in a patch release. But I could take a stab at cherry-picking the PRs into main after this merges to dev. It may just cause some odd conflicts to pop up when we try merging dev -> main.

dyfer:
> I don't think it would have any practical impact, short of having a bug fix PR off main that we need publish in a patch release. But I could take a stab at cherry-picking the PRs into main after this merges to dev. It may just cause some odd conflicts to pop up when we try merging dev -> main.

Okay, let's merge this into `dev` then. If you later open a PR merging dev->main we could check with others if anybody has any opposition to that... but it looks like it would be benign enough.

dyfer (approved):

PR #1008, opened 2023-04-27
Show error message when single studio limit is reached

nwang92:
Looks like this:
https://user-images.githubusercontent.com/6201822/234994788-0cb3d77b-818f-4f4f-ae33-12ffef0b3a1d.mov


mikedickey (approved):
works for me

PR #1009, opened 2023-04-28
Remove yes/no at first open

mattahorton:
Adds a preprocessor def `VS_FTUX` that disables the Yes/No screen on launch. This was discussed in the jitterbug list. 

It does not remove the classic GUI from the build, and in fact, you can still access the classic GUI from settings in VS mode.

There is also a workflow change to make jacktriplabs/jacktrip builds use the `VS_FTUX` flag.

I'll need help testing CMake. 

CleanShot 2023-04-28 at 13 00 45@2x


mattahorton:
I updated the preprocessor definition name to `VS_FTUX` to more accurately describe a Virtual Studio mode first-time UX.

mikedickey (approved):

PR #1010, opened 2023-04-30
Fixing Play Test Tone button on Linux

mikedickey:
It turns out that it is unsafe to include multiple faust-generated C++ files in the same object. They contain multiple symbols which use the same name, including APIUI and #ifdef statements. Mixing these together leads to very unpredictable behaviour, including crashes.

In particular, I discovered the the vsAudioInterface class was including Tone.h, Meter.h and Volume.h, and each of those was including their own faust-generated c++ files. This was causing symbol conflicts in the generated object code. It's actually quite suprising that it only seemed to have a side-effect on Linux, and only in it "not working." I think we were just getting lucky.

I've reworked the Tone, Meter and Volume classes so that the faust generated c++ code is only used in their .cpp files. This ensures that each of them uses the correct symbols, and the duplicate names do not cause conflicts.

### DIFF
--- a/src/Meter.cpp
+++ b/src/Meter.cpp
@@ -38,7 +38,37 @@
 
 #include "Meter.h"
 
+#include <iostream>
+
 #include "jacktrip_types.h"
+#include "meterdsp.h"
+
+//*******************************************************************************
+Meter::Meter(int numchans, bool verboseFlag) : mNumChannels(numchans)
+{
+    setVerbose(verboseFlag);
+    for (int i = 0; i < mNumChannels; i++) {
+        meterP.push_back(new meterdsp);
+    }
+}
+
+//*******************************************************************************
+Meter::~Meter()
+{
+    for (int i = 0; i < mNumChannels; i++) {
+        delete static_cast<meterdsp*>(meterP[i]);
+    }
+    meterP.clear();
+    if (mValues) {
+        delete mValues;
+    }
+    if (mOutValues) {
+        delete mOutValues;
+    }
+    if (mBuffer) {
+        delete mBuffer;
+    }
+}
 
 //*******************************************************************************
 void Meter::init(int samplingRate)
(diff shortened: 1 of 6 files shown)


psiborg112:
Are we using Faust generated code in a problematic manner elsewhere, or is it just in the vsAudioInterface class that it's an issue?

mikedickey:
> Are we using Faust generated code in a problematic manner elsewhere, or is it just in the vsAudioInterface class that it's an issue?

Great question! Why don't we run into similar issues with Limiter, Compressor and (especially) Reverb? All three are also being included by Effects. This could be more of a Faust version mismatch problem than anything: if `APIUI` is declared the same way across all the Faust c++ files, the conflict is benign.

Just spot checking this theory, it seems to hold. I can see that most of the Faust c++ files were generated with `2.41.1` while others use `2.50.6`:

```
$ grep "Code generated with Faust" *.h
compressordsp.h:Code generated with Faust 2.41.1 (https://faust.grame.fr)
freeverbdsp.h:Code generated with Faust 2.41.1 (https://faust.grame.fr)
freeverbmonodsp.h:Code generated with Faust 2.41.1 (https://faust.grame.fr)
limiterdsp.h:Code generated with Faust 2.41.1 (https://faust.grame.fr)
meterdsp.h:Code generated with Faust 2.41.1 (https://faust.grame.fr)
stereotomonodsp.h:Code generated with Faust 2.50.6 (https://faust.grame.fr)
tonedsp.h:Code generated with Faust 2.50.6 (https://faust.grame.fr)
volumedsp.h:Code generated with Faust 2.41.1 (https://faust.grame.fr)
zitarevdsp.h:Code generated with Faust 2.41.1 (https://faust.grame.fr)
zitarevmonodsp.h:Code generated with Faust 2.41.1 (https://faust.grame.fr)
```

So long as the files included together were generated using the same version, it may be fine. The crossing of versions (in this case both `tonedsp.h`, `volumedsp.h` and `meterdsp.h` are included in `vsAudioInterface`) may be what triggers the issue.

nwang92 (approved):

josmithiii:
For more discussion, see https://github.com/jacktrip/jacktrip/pull/1011#issuecomment-1534014181

PR #1011, opened 2023-05-01
Moving faust-generated code out of header files

mikedickey:
For remaining ProcessPlugins..

We recently discovered it is unsafe to includes these in headers because it can lead to symbol conflicts.

See https://github.com/jacktrip/jacktrip/pull/1010

### DIFF
--- a/src/Compressor.cpp
+++ b/src/Compressor.cpp
@@ -37,6 +37,81 @@
 
 #include "Compressor.h"
 
+#include "compressordsp.h"
+
+//*******************************************************************************
+Compressor::Compressor(int numchans,  // xtor
+                       bool verboseIn, float ratioIn, float thresholdDBIn,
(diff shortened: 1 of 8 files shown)


nwang92 (approved):

josmithiii:
This is surprising since the headers have include-guards and should not be allocating anything (just interfaces).  What is the easiest way to reproduce the problem if I go back to a commit before this workaround?  I want to see the build log and see what the multiple symbols are.  I can fix problems in APIUI.h et al. for the Faust folks.

josmithiii:
Ok, I can see how using Faust headers from two different Faust versions in the same executable could cause problems.   The APIUI class probably changed its layout.  The build process should make that impossible and enforce only one Faust version per compilation.  In my work I've taken to putting all my Faust code in its own JUCE-style module.  Both including (for inlining) and linking are still possible.

PR #1014, opened 2023-05-02
Enable Local Monitoring of Inputs

dhing1024:
Adds a local monitoring slidebar and functionality to the connected screen. Key features:

- The behavior can depend on the input mixing mode. For example, when using stereo, both channels are copied individually to the monitor. When using mix-to-mono, the monitor receives the preprocessed mixed audio in BOTH speakers (similar to how volume meters behave for mix-to-mono by copying the values to both bars)

- The monitor volume is independent of the input volume. That means you might still hear yourself even if your "input volume" that you're sending to the network is off or muted.

- Adds a dedicated `ProcessPlugin*[]` array specifically for monitoring plugins in addition to the "to network" and "from network" plugin arrays. It might be useful to have more than 1 plugin on the monitoring chain (perhaps for a separate volume meter for the monitor or for feedback detection)

- The number of monitor chans (`nChansMon` throughout `AudioInterface.cpp`) is the minimum between the # of hardware inputs and hardware outputs. It does not take into account the # of network inputs and network outputs (which can be different from the # of hardware inputs/outputs, if for example using `mix-to-mono`)

- This doesn't add anything to the "Settings" or "Audio Setup" screens, at least for the time being. On those screens vertical space is probably going to be an issue. I'm also wondering if the monitor should have it's own volume meters (like on my physical audio interface), though it might be confusing. Apparently distinguishing between input and output is already difficult for some Virtual Studio users and adding new monitoring features will probably make that even more confusing. (maybe we should Figma this?)

- There isn't muting of the monitor (at the moment), though users can just slide the bar all the way to the left. I'm not yet sure if / how we should add that in.

Screenshot 2023-05-02 at 11 18 44 AM

ASIO is busted on my Windows machine for some reason and I can't really fix it, so I'm not able to test ASIO. I did try my non-ASIO devices on Windows and there is a noticeable audio delay, compared to my Mac. It's a good demo of why people should use ASIO drivers if they have one, since there's only hardware delay and no network delay. If someone can confirm that this works with ASIO devices that would be wonderful.

This implementation takes into accounts learnings from #1010


mikedickey:
I tested on Windows with ASIO drivers (UA Volt) and latency is great/unnoticeable.

mikedickey:
Some thoughts on this:

- I think it's fine (possibly best) that the monitor slider does not have its own corresponding meters. Also no need for a monitor "Mute"
- I think monitor slider would be better in the output section, since this impacts only what you hear versus what you send to the studio
- Could we put the text "STUDIO" to the left of the existing output volume slider to further emphasize the difference between these?
- Instead of "MON" maybe label this as "YOU" or "SELF"?
- If my input device is set to a single channel (Mono), I only hear myself on the left. I think it should be both sides, same as if I am using "Mix to Mono"

dhing1024:
These comments should be addressed now - here is the updated UI

Screenshot 2023-05-03 at 12 23 25 PM

mikedickey:
Latest build crashes for me (on Windows) if I have a single input channel selected.

mikedickey (approved):
works for me

nwang92 (approved):

PR #1015, opened 2023-05-02
Clean Up Code Enclosed in WAIR Ifdefs

dhing1024:
I was wondering if it would be a good idea to remove code surrounded by `#ifdef WAIR` preprocessor directives. Some reasoning behind this:

- when I add WAIR to the qmake build by uncommenting out [this line](https://github.com/jacktrip/jacktrip/blob/0ff391cc5011a439aa7d2607460d787a873d5e94/jacktrip.pro#L47), the compiler  errors out on [this line ](https://github.com/jacktrip/jacktrip/blob/0ff391cc5011a439aa7d2607460d787a873d5e94/src/JackTripWorker.cpp#L52)trying to import a nonexistent file. That line was added [in this commit](https://github.com/jacktrip/jacktrip/commit/33688be53e829f120e673df271b673ee4f262f0b#diff-39dbdacec5c8e0515ba0094e1701d115ba64169aca7a47d338ab9b78eecb2936), I think a merge from a separate repository hosting platform, after which that file can't be found - see [here](https://github.com/jacktrip/jacktrip/tree/33688be53e829f120e673df271b673ee4f262f0b). So it seems like this has been broken at least 3 or 4 years, perhaps longer.

- Taking these out makes the code feel less bloated and easier to work with. Especially when adding new features we don't have to read through unused code and figure out how to write an equivalent WAIR version of the same feature, particularly in files such as `AudioInterface.cpp`.

- As far as I know we haven't really been testing anything that gets added to WAIR code as a result of new feature development. (I haven't been). It doesn't seem like it's something we officially maintain or support.

I have no idea what the WAIR definition is actually used for - it just seems to me like it's been broken for a while and isn't of any use. If someone has any insight as to why it should be kept we don't need to take these out. I also left in the `WAIRTOHUB` defines since that does seem like it's used, particularly in server mode.


psiborg112:
Ultimately it should probably be the decision of @cchafe but I don't think it should be removed. (I've never found it particularly onerous to work around its existence.) WAIR stands for wide area internet reverberation, and there's a paper on it in the papers folder of the repository. It's not a configuration that any end users are going to encounter, but it is a concept that I'm interested in playing around with at some stage when I finally have some more time.

cchafe:
Here's the [paper](https://www.frontiersin.org/articles/10.3389/fdigh.2018.00027/full) -- the extra clutter is annoying to me, too and there's no work going on with WAIR at the moment. It most certainly needs revision after all the updates since then. My worry is getting back to it later will be doubly difficult if it's removed and things drift apart further.

dhing1024:
Going to close this PR without merging. We had a short discussion about it during our open source meeting and @dyfer had a really good suggestion that we could use build automation to do some basic smoke testing of WAIR code. That won't necessarily make it bug free but will at least enforce a minimum standard of maintenance. 

If @cchafe has some extra time, he might be able to get WAIR code a little more up-to-date. I'll turn this into as an issue instead and hopefully it can get resolved before this code becomes too stale.

PR #1016, opened 2023-05-03
Enforce Same ASIO Device on Windows

dhing1024:
Doesn't require any changes to UI code. But we may want to update the help article at https://help.jacktrip.org/hc/en-us/articles/4409919243155


mikedickey:
I'm still getting the same scary popup messages when using a build from github actions

dhing1024:
Still a WIP

dhing1024:
I think this should be good now, but not totally sure. Will need someone other than me to test ASIO.

mikedickey:
If I change either the input or output device to an ASIO device, the backend seems to change the other device to match the same one, but it doesn't update the selection list shown in the user interface for it. The selection list still displays the old device I had selected previously, even though it's no longer being used.

mikedickey (approved):

PR #1017, opened 2023-05-03
Output Volume Meters Include Monitor

dhing1024:
Requires #1014. Moves the output volume meter to the `ProcessPluginsToMonitor` chain (as opposed to the `ProcessPluginsFromNetwork` chain) and allows `Meter()` to measure from the second parameter in the `compute()` function.


nwang92 (approved):
Seems to work for me on OSX

mikedickey (approved):
Works for me on Windows

PR #1012, opened 2023-05-01
Refresh devices when opening the popup

nwang92:
There's a small downside here which is calling "refreshDevices" disconnects you from audio briefly. I guess this is because it's unsafe to reload the device list when it's wired up to meters + other things?

PR #1022, opened 2023-05-05
Feature/adding icons to connected view

nwang92:
Porting the layout + icon changes out of https://github.com/jacktrip/jacktrip/pull/1012 and into a separate PR

### DIFF
--- a/src/gui/Connected.qml
+++ b/src/gui/Connected.qml
@@ -159,9 +159,9 @@ Item {
     Item {
         id: deviceSettings
         visible: showReadyScreen && isUsingRtAudio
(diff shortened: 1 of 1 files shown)


src/gui/Connected.qml
```diff
@@ -651,30 +689,6 @@ Item {
                         color: textColour
                     }
 
-                    Text {
```
nwang92, line 654: FYI I removed this because the text is pretty lengthy and pushes things around in the pop-up. But this message appears in the connected view already

mikedickey (approved):

PR #1029, opened 2023-05-06
Updating manifests to match what is on files.jacktrip.org

mikedickey:
Adding script to help automate this

Without this, everyone who has a release older than 1.8.0 has to upgrade twice to catch up. I think we should just manually keep these in sync for a few more months to smooth the migration process.


mikedickey:
Note that the diff shows everything has changed because of whitespace differences between when we were editing these manually versus auto-generating them.

nwang92 (approved):

PR #1035, opened 2023-05-12
Detect Feedback Loops

dhing1024:
Detects feedback loops and mute the user's audio input if a feedback loop is found. This uses a Faust plugin that does a running FFT-128 computation to calculate spectral power vs time, then uses the results as a heuristic for determining whether an audio feedback loop exists. (See `Analyzer::checkForAudioFeedback` for implementation details)

The FFTs are done outside of the main process callback in a separate thread since it may be a more expensive computation, and to avoid situations where the audio process callback doesn't finish executing in time. This implementation uses a shared buffer, implemented as a ring buffer that can grow if necessary, between both threads. The main process callback adds audio samples to the ring buffer, and the second thread reads from it regularly using a timer. The ring buffer is protected from concurrent reads and writes using a mutex.

There are two instances of `Analyzer`, one acting on the input signal and another acting on the monitor's combined input/output signal. The `Analyzer` acting on the input signal checks for feedback from the studio (I think this would only happen if the studio loopback volume is high enough) and the other `Analyzer` checks for feedback between the local speaker/microphone circuit (if the monitor is on and the user isn't using headphones).

https://github.com/jacktrip/jacktrip/assets/26987971/0b521444-536d-44ea-abd4-ac053412dc51

I was able to able to get it to work reasonably well using stuff I have at home, though I don't know how well that will translate to performance more broadly. One of the risks is that there will be too many false positives, which may cause unwanted muting during a live session. If we need to make adjustments we should err on the side of minimizing false positives.


mikedickey:
I think we need to be careful to not have any mutexes or memory allocations happening inside audio callbacks, since these can be cpu intensive and possibly blocking, both of which cause xruns/audio glitches. Given that the need here seems to be for a single-producer, single-consumer, a wait-free ring buffer implementation is possible. See the sample implementation [here](https://www.boost.org/doc/libs/1_64_0/doc/html/atomic/usage_examples.html), although now we can use `std::atomic` instead of `boost::atomic` (its predecessor).

src/gui/virtualstudio.cpp
```diff
@@ -1617,6 +1623,13 @@ void VirtualStudio::completeConnection()
         connect(this, &VirtualStudio::updatedMonitorVolume, m_monitor,
                 &Monitor::volumeUpdated);
 
+        // Setup output analyzer
```
mikedickey, line 1626: I understand that feedback could be generated either by use of your monitor signal getting picked up by your input, or by the studio signal (a) looping back your input signal + (b) your output signal getting picked up by input. But I'm not following on why a second detector would be required. If the first case is true, wouldn't the feedback also show up on your input signal?

mikedickey (changes requested):

dhing1024:
> I think we need to be careful to not have any mutexes or memory allocations happening inside audio callbacks, since these can be cpu intensive and possibly blocking, both of which cause xruns/audio glitches. Given that the need here seems to be for a single-producer, single-consumer, a wait-free ring buffer implementation is possible. See the sample implementation [here](https://www.boost.org/doc/libs/1_64_0/doc/html/atomic/usage_examples.html), although now we can use `std::atomic` instead of `boost::atomic` (its predecessor). You could also make each element an array of buffer_size, so that each callback only has to push 1 element.


What happens if the buffer isn't large enough to hold all of the queued samples? I don't think we'd be able to resize the buffer if it is too small. If we use a wait-free ring buffer, we won't drop any samples due to locking as we do now (currently, a few samples will be dropped from time to time if the consumer is reading from the buffer concurrently - see [here](https://github.com/jacktrip/jacktrip/blob/64554ef3aa6c3435584ee9e6a0d396ff74dbfb55/src/Analyzer.cpp#L170)). On the other hand, if the amount of memory is fixed at initialization, we might end up dropping a lot of samples if the buffer runs out of space.

We also only allocate a new `mSumBuffer` in the callback if the current size is not large enough, which is similar to how it's done in [Meter.cpp](https://github.com/jacktrip/jacktrip/blob/34c50c09f105be5cdfc6ad2d665ad528e7b823ce/src/Meter.cpp#L128)

src/gui/virtualstudio.cpp
dhing1024: I was trying to see if we could distinguish between the two forms of feedback using two separate instances of `Analyzer`. I think using only 1 also works, perhaps even a little better. In that case we'd probably want to mute both the monitor and input volume since we can't tell the two apart.

If we have only 1 instance it would need to go before the input volume plugin or after the monitor plugin - not sure which is better but I kept the one after the monitor.

nwang92:
I tested this out and it seems to work beautifully. I guess the only concern would be how well it works for normal day-to-day use but that's probably something we should test with beta versions maybe?

Other that that, this seems fine to me.

nwang92 (approved):

src/Analyzer.cpp
```diff
@@ -0,0 +1,369 @@
...
+//*******************************************************************************
+void Analyzer::onTick()
+{
+    // cannot process audio if the no samples have been added to the ring buffer yet
+    if (!hasProcessedAudio) {
+        return;
+    }
+
+    const uint32_t samples = mCircularBuffer.size();
+    std::vector<float> fftBuffer;
```
mikedickey, line 151: can we make this a class member so that resize() only performs memory operations when necessary to grow, versus every call to `onTick`?

src/Analyzer.h
```diff
@@ -0,0 +1,113 @@
...
+    int mNumChannels;
+    bool mIsMonitoringAnalyzer = false;
+    bool hasProcessedAudio     = false;
+    QTimer mTimer;
+
+    void* mFftP;              // Faust plugin
+    uint32_t mFftSize = 128;  // FFT size parameter
+
+    // ring buffer that doesn't require locking
+    WaitFreeRingBuffer<float, 4096> mCircularBuffer;
```
mikedickey, line 97: I think we may need a bigger number for the size. At 48kHz, we'll push 4800 sums every 100 milliseconds. Maybe `81920` to be safe (up to 768kHz)?

mikedickey (changes requested):

src/Analyzer.cpp
dhing1024: done

src/Analyzer.h
dhing1024: done

mikedickey (approved):

PR #1037, opened 2023-05-17
Use Device Authorization Flow

dhing1024:
Refactors the user authentication and reimplements the authentication flow.

- Implements and uses the [device authorization flow](https://auth0.com/docs/get-started/authentication-and-authorization-flow/device-authorization-flow) for Virtual Studio users. There's no direct implementation of this flow in the [Qt OAbstractOAuth classes](https://doc.qt.io/qt-5/qabstractoauth2.html), even in Qt6, so this is implemented as a separate class (vsDeviceCodeFlow.cpp/h). Ideally we would be using a library rather than implementing our own security flow, but this seems to be the next best option.
- Adds a class `vsAuth.cpp/h` to track user authentication state in the app. I added this as a separate class to keep the main Virtual Studio class from getting even more bloated. The auth state is accessible in QML under `auth.*`, rather than `virtualstudio.*`
- Adds a separate class (vsApi.cpp/h) for Virtual Studio API calls. This is needed because the existing code uses the existing QOAuthCode flow authorizer (`m_authenticator`) to make API calls. This PR stops making use of `m_authenticator`.
- For Virtual Studio @jacktrip.org users, you don't have to log out and log back in switching from prod to test mode.

There are a few different styles of the auth flow we could do. When the user opens the browser, we have the option to (1) open `/activate` in browser and have the user type in the code, or (2) open `/activate` in browser and have the code automatically get filled out for them. We can also choose to (1) open the browser automatically when clicking "Sign In" or (2) requiring them to press "Open Browser" or visit the URL.

https://github.com/jacktrip/jacktrip/assets/26987971/82b8f1c0-522b-4cd3-a6ef-93efeeba0bc5

https://github.com/jacktrip/jacktrip/assets/26987971/a3a2762e-c1a0-478b-8f2b-8addd275993b

From experimenting a bit with the style and feel of it, this PR currently implements it as
- Clicking the URL or the "Open Browser" buttons from in-app goes to `/activate` and pre-fills the code
- If you open the browser separately and type in the URL manually, the user will need to type the code in
- We don't automatically open the browser for the user. I think this would be a little too aggressive in that it asks you to confirm or type in the code from the app, which means you would have two switch focus between the browser and app to log in. That's difficult especially if the browser opens full-screen, which is likely, and may overall just be too confusing.

Regardless of how we implement this on this repository, we'll need to make some adjustments to the Auth0 sign-in page as well. There are some minor formatting issues there.

Relevant Asana tickets:
https://app.asana.com/0/1204190814361100/1204518663542885/f
https://app.asana.com/0/1204414944562207/1203982516427023/f
https://app.asana.com/0/1200778084029081/1204534315242903/f


nwang92:
I think the "pre-filled code with redirect" is the right move because that makes it as easy as possible to sign in. I don't mind the currently implemented browser flow because it prepares me for what to do. It does add extra clicks, but if the "Sign In" button were to open a browser itself then it could hide the prompt on what I'm supposed to do so it seems like an ok trade off in my opinion.

nwang92:
NIT but maybe color the "success" screen with a green check so it looks ok on both light/dark mode? I grabbed this screenshot on dark mode when it was transitioning
Screenshot 2023-05-17 at 4 24 50 PM

Also there was a point where I was stuck on the "Success" screen for like 3 seconds and I wasn't sure if it got stuck or not. Maybe some subtitle there with a message along the lines of "Loading app" would help.

nwang92:
Another small annoyance of mine is that I can't select the 8-character code with a cursor. I'm not sure if that's a QML limitation on the text elements but it would be nice to have a way to copy the code so I don't need to manually type it all in.

nwang92 (approved):
This seems good from what I can tell. I like the refactored vsApi class!

nwang92:
This is more of an external comment, but do you happen to know if this page is customizable at all? I feel like we should include a link to `/studios` or something. The current "landing page" embedded into the app (prior to this branch) had something similar. But since this is a hosted page, it's not something we need to change immediately.

dhing1024:
> This is more of an external comment, but do you happen to know if this page is customizable at all? I feel like we should include a link to `/studios` or something. The current "landing page" embedded into the app (prior to this branch) had something similar. But since this is a hosted page, it's not something we need to change immediately.

The page is customizeable - in the same way that we can customize login and signup screens. So we can probably do something to make it less awkward at the end.

Another thing to note: I suspect that the fact that the widget content isn't vertically centered is due to a rendering bug with Auth0 and not from our own custom templating system. I'm filing a bug report with Auth0 so hopefully they can address it.

PR #1038, opened 2023-05-19
Use a wait-free packet queue for Regulator worker thread

mikedickey:
Support for using a shared Regulator worker thread (disabled)

Changing default buffer strategy for virtual studio mode to 3 (worker=true) with auto3

Added reusable WaitFreeRingBuffer and WaitFreeFrameBuffer classes

Fix regulator stats with bufstrategy 4


mikedickey:
This is a pretty extensive update, triggered by realizing that some audio devices cannot handle running a single `pullPacket` call in the time alloted for the callbacks. I previously thought (for bufstrategy 4) that it would be fine to do this for clients because it's only 1 instance (versus servers with many), but further testing proved that to be false. It works sometimes, for some (higher quality? faster?) audio interfaces, but not other times.

nwang92:
From some functional testing, this seems to clear up the static/crackles for me with my Blue Yeti + Apple built-in mic. Still looking over the code to see if I understand the changes

mikedickey:
I've tested this across a variety of devices and operating systems. It's not perfect, but it seems to be a massive improvement.

jacktrip.pro
```diff
@@ -209,6 +209,8 @@ HEADERS += src/DataProtocol.h \
            src/CompressorPresets.h \
            src/Limiter.h \
            src/Regulator.h \
+           src/WaitFreeRingBuffer.h \
+           src/WaitFreeFrameBuffer.h \
```
nwang92, line 213: I think these need to get added to the meson.build file too?

mikedickey: I wasn't sure if that was only for QOBJECT mocking or not, but I added them in just in case.

nwang92 (approved):

PR #1039, opened 2023-05-19
Improvements and Fixes for Device Flow Code

dhing1024:
Follow up to #1037 

- Resolves issues with the back button appearing (and acting buggy) on the `vsftux` builds
- Removes the need for an extra sign up screen. This is done by attempting to log in (either through the code flow protocol or via the refresh token) whenever we switch to the login screen.
- If there's a refresh token stored, this doesn't initialize the code flow protocol. If this token is valid, the user shouldn't need to do anything to move to the next screen. If this token is invalid, this is gracefully handled by displaying an error message and initializing the code flow procedure.

Relevant tickets:
https://app.asana.com/0/1204518798614303/1204632412091470/f
https://app.asana.com/0/1204518798614303/1204632410251062/f
https://app.asana.com/0/1204518798614303/1204632410251060/f

PR #1040, opened 2023-05-23
Consolidating text and moving things up about 50px

mikedickey:
I think this would give us a better chance of fitting the key parts "above the fold" for small resolution (or high scaled) screens.

Screenshot 2023-05-22 at 9 09 36 PM


dhing1024:
I'm doing some refactoring of this qml file for other needs - I will incorporate this idea into #1039

PR #1043, opened 2023-05-25
Optimize PLC regulator performance by using double instead of long double

mikedickey:
This appears to give a 2-3x performance boost

### DIFF
--- a/src/Regulator.cpp
+++ b/src/Regulator.cpp
@@ -599,7 +599,7 @@ bool BurgAlgorithm::classify(double d)
     return tmp;
 }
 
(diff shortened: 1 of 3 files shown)


cchafe (approved):
Using long double precision is a remnant and no longer needed because 1) we use a larger number of recent packets in the training and 2) prediction is only two packets, the current missing one plus a future second one which gets cross-faded with the actual future.

For the record, here's where the extra precision matters 
double
double100Hz512FPP
long double
longDouble100Hz512FPP
The graphs show Burg with double precision, 100Hz input, a sequence of several predicted packets that starts on the second packet, trained on the first packet.

PR #1045, opened 2023-05-29
Bump version for 1.10.0 beta1 release

mikedickey:


cchafe (commented):
Should we count up 1.10, 1.11 ahead of 2.0.0? -- sounds like it's not 1.9.1-beta1

nwang92 (approved):

mikedickey:
> Should we count up 1.10, 1.11 ahead of 2.0.0? -- sounds like it's not 1.9.1-beta1

Yes, I believe that is most standard for semantic versioning. It's more than a patch (1.9.1) and doesn't seem to qualify for a major version bump (2.0.0). See https://semver.org/spec/v2.0.0.html

```
A normal version number MUST take the form X.Y.Z where X, Y, and Z are non-negative integers, and MUST NOT contain leading zeroes. X is the major version, Y is the minor version, and Z is the patch version. Each element MUST increase numerically. For instance: 1.9.0 -> 1.10.0 -> 1.11.0.
```

mikedickey:
For final release, I'd stil llike to include https://github.com/jacktrip/jacktrip/pull/1035 and possibly some more tweaks to Virtual Studio auth flow, but I think they can wait for a beta2.

PR #1049, opened 2023-05-31
Optionally Disable Feedback Detection

dhing1024:
Requires #1035

Adds a popup notification when feedback occurs, and allows the user to disable it under advanced settings.

Screenshot 2023-05-31 at 4 05 29 PM

Screenshot 2023-05-31 at 4 05 39 PM


nwang92 (approved):

src/gui/virtualstudio.cpp
```diff
@@ -277,6 +280,9 @@ VirtualStudio::VirtualStudio(bool firstRun, QObject* parent)
     m_view.engine()->rootContext()->setContextProperty(
         QStringLiteral("updateChannelComboModel"),
         QVariant::fromValue(m_updateChannelOptions));
+    m_view.engine()->rootContext()->setContextProperty(
```
nwang92, line 283: The only relatively minor comment I have is this combo box could just be a check box so we can remove `m_feedbackDetectionOptions`

dhing1024: I think we can probably just leave this in - primarily for the sake of uniformity on that settings page. If we really don't like it we can probably just change it later.

mikedickey (approved):

PR #1051, opened 2023-06-01
Optimize circular buffer for feedback detection

mikedickey:
Push and pull buffers of samples rather than one at a time

Add buffer size to ProcessPlugin::init and other related code cleanup

### DIFF
--- a/docs/changelog.yml
+++ b/docs/changelog.yml
@@ -1,3 +1,7 @@
+- Version: "1.10.0-beta2"
+  Date: 2023-06-02
+  Description:
+  - (added) VS Mode automatically mute when feedback is detected
 - Version: "1.10.0-beta1"
   Date: 2023-05-29
   Description:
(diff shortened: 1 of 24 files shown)


mikedickey:
I've observed that the new feedback detection feature still roughly doubles jacktrip's CPU usage on my M2 MacBook. I tracked this down specifically to the Faust code, or this line:

```static_cast<fftdsp*>(mFftP)->compute(samples, &data, mAnalysisBuffers);```

Not sure if there is way to optimize that any further?

cchafe:
In fftdsp.dsp he FFT size is 128
the documentation says it will perform a sliding FFT over a real input signa
does that mean full hops?
(if sliding means each sample, then that's problematic)

mikedickey:
> In fftdsp.dsp he FFT size is 128
> the documentation says it will perform a sliding FFT
> does that mean full hops?
> (if sliding means each sample, then that's problematic)

Hi @cchafe, I believe @demarcod1 is out today, but I will forward this to him via Slack

nwang92 (approved):

mikedickey:
I'm going to merge to roll a beta2 today with the latest. Feedback detection definitely needs more work, but this will help get it more testing. Going to leave it disabled by default for now.

PR #1054, opened 2023-06-06
Adjust Feedback Detection Sensitivity

dhing1024:
Hopefully this is a little better? I tried to define the thresholds a little better in relation to what the expected FFT outputs would be, and adjust the sensitivities that way.

Note that we are squaring in our Faust callback to get a power spectrum. The FFT outputs complex values with a magnitude of less than or equal to 128 for (size N = 128 and non-clipping input signals). We square to get real valued magnitudes and use that for the analysis.


mikedickey:
I think this is better. It clears up the false positives for me.

mikedickey (approved):

PR #1055, opened 2023-06-06
Improve FFT Performance

dhing1024:
I haven't actually tried seeing how this performs CPU-wise, but this should regardless be an improvement to the existing stuff. With these changes, we now calculate a single FFT-128 on ever `onTick` using only the latest 128 samples. (This means that the FFT algorithm doesn't need read every sample, but rather only chunks of 128 samples every 100ms)

FFT-128 (128 samples) -----> 100ms without FFT -------> FFT-128 (128 samples)

Using https://github.com/d1vanov/Simple-FFT to do the computations. Since we need to  make modifications to `fft_settings.h` in order to make this work, I didn't use the normal git submodules feature for this, rather just adding them as individual files. There's probably a better way to do that, perhaps by maintaining our own fork of the FFT repository or something, but this seems to be a low-cost way of doing the trick.

It might be worth taking another look at the sensitivity. I don't expect these changes to modify behavior but empirical confirmation can never hurt.


ntonnaett:
Why didn't you choose fftw? I'm surprised that the Meson build doesn't fail.

dhing1024:
> Why didn't you choose fftw? I'm surprised that the Meson build doesn't fail.

It looked like it would be easier to integrate into the existing codebase. @mikedickey also said that it has better software use licensing terms.

nwang92:
This is neat, from some playing around I do see noticeable CPU improvement on my M1 macbook from peaking at 16-18% on dev down to 8-12% on this branch.

I'm not sure what the best practices are on integrating libraries. We did take the same approach here by embedding [dblsqd](https://github.com/Mudlet/dblsqd-sdk-qt/tree/master/dblsqd), and if we continue to do this I would say we should remove the `benchmark-tests` and `unit-tests` directories unless they serve a functional purpose.

nwang92:
On a functional testing of the sensitivity, I'm able to trigger a pretty nasty feedback loop and the detection logic doesn't seem to be kicking in anymore.

mikedickey:
This seems to completely eliminate the CPU regression. At least on my MacBook, there is no measurable difference in CPU usage between enabling versus disabling it.

mikedickey:
> On a functional testing of the sensitivity, I'm able to trigger a pretty nasty feedback loop and the detection logic doesn't seem to be kicking in anymore.

@nwang92 Are you comparing against the latest in dev or beta2? This is a lot less sensitive versus beta2, but I believe that change was introduced by https://github.com/jacktrip/jacktrip/pull/1054/files

I'm not seeing any difference in sensitivity of this branch versus dev.

nwang92:
I'm using this branch and attempting to force a feedback loop using the built-in speakers + mic. The in-app monitor volume is 0 but there is some studio loopback.
Screenshot 2023-06-08 at 3 58 59 PM
Screenshot 2023-06-08 at 4 02 24 PM
Screenshot 2023-06-08 at 4 02 56 PM

mikedickey:
> I'm using this branch and attempting to force a feedback loop using the built-in speakers + mic. The in-app monitor volume is 0 but there is some studio loopback.

If you run the same test using a build off dev branch, do you get the same results? If so, we should merge this and keep tweaking sensitivity separately.

ntonnaett:
The easiest solution for Flatpak is to wrap the new dependency in a cmake subproject.

nwang92 (approved):

mikedickey (approved):

Issue #1036, opened 2023-05-14
New QML errors with 1.9.0


ntonnaett:
That's with the default configuration on Flatpak. I thought we test this in CI. At least the qmake build should have failed.

```
qt.qpa.qgnomeplatform: Could not find color scheme  ""
Qt: Session management error: None of the authentication protocols specified are supported
qrc:/vs/AudioSettings.qml:218: TypeError: Cannot read property 'text' of undefined
qrc:/vs/AudioSettings.qml:527: TypeError: Cannot read property 'text' of undefined
qrc:/vs/AudioSettings.qml:218: TypeError: Cannot read property 'text' of undefined
qrc:/vs/AudioSettings.qml:527: TypeError: Cannot read property 'text' of undefined
qrc:/vs/Connected.qml:366: TypeError: Cannot read property 'text' of undefined
qrc:/vs/Connected.qml:556: TypeError: Cannot read property 'text' of undefined
file:///usr/lib/qml/QtQuick/Controls.2/org.kde.desktop/ComboBox.qml:117: TypeError: Type error
file:///usr/lib/qml/QtQuick/Controls.2/org.kde.desktop/ComboBox.qml:117: TypeError: Type error
file:///usr/lib/qml/QtQuick/Controls.2/org.kde.desktop/ComboBox.qml:117: TypeError: Type error
file:///usr/lib/qml/QtQuick/Controls.2/org.kde.desktop/ComboBox.qml:117: TypeError: Type error
file:///usr/lib/qml/QtQuick/Controls.2/org.kde.desktop/ComboBox.qml:117: TypeError: Type error
file:///usr/lib/qml/QtQuick/Controls.2/org.kde.desktop/ComboBox.qml:117: TypeError: Type error
file:///usr/lib/qml/QtQuick/Controls.2/org.kde.desktop/ComboBox.qml:117: TypeError: Type error
file:///usr/lib/qml/QtQuick/Controls.2/org.kde.desktop/ComboBox.qml:117: TypeError: Type error
file:///usr/lib/qml/QtQuick/Controls.2/org.kde.desktop/ComboBox.qml:117: TypeError: Type error
file:///usr/lib/qml/QtQuick/Controls.2/org.kde.desktop/ComboBox.qml:117: TypeError: Type error
file:///usr/lib/qml/QtQuick/Controls.2/org.kde.desktop/ComboBox.qml:117: TypeError: Type error
[...]

PR #1056, opened 2023-06-07
Addressing some QML errors

nwang92:
I think this fixes some (but not all) of the errors in https://github.com/jacktrip/jacktrip/issues/1036

Also moving `Colorize` to not reference `parent` because that doesn't work in Qt6


ntonnaett:
The number of error messages has shrunk in half. The interface seems less glitchy. Two things I currently see: The Settings button has no text. And the margin between meter and fader blocks are much too big. I approve this because it seems to be moving in the right direction and you might want to open another PR for more work on this. But we should have this fixed for the next release. I'm currently holding the update on Flathub because the VS gui is too broken with the current stable version.

ntonnaett (approved):

PR #1063, opened 2023-06-11
Fixed issue with RTAudio scanning interfaces several times at startup

mikedickey:
I suspect this was causing a lot of problems on windows, but at a minimum this should significantly reduce Windows startup time and improves responsiveness when using RTAudio.

It now only scans interfaces when necessary, and not before trying to log into Virtual Studio (which caused what looks like login failures when there are actually audio interface issues). It also now logs whenever it is scanning devices, and prints out the results.

Eliminated duplicated code by caching the results of device scans.

Fixed some qml bugs/warnings with classic mode and eliminates rootContext()->setContextProperty() calls moving things into QObject properties.

    Defer settings and audio initialization outside of QJackTrip constructor

    There currently is a strange coupling between the QJackTrip gui and the
    Virtual Studio gui, such that they are both running at the same time.
    This should probably be addressed, but is for another PR..

    This one simple defers the settings initialization for QJackTrip, which
    includes RTAudio initialization and scanning, so that it is only done
    when the QJackTrip interface is being used.

    Previously, QJackTrip would try to run RTAudio device scans even before
    the initial window was painted. This likely conflicted with the same
    work being performed by the Virtual Studio interface. At best, this
    would cause unnecessary delays at startup. At worst, I believe it was
    causing crashes on Windows, especially when there are ASIO drivers
    present, since that framework seems to be extremely brittle and touchy.

### DIFF
--- a/src/gui/virtualstudio.cpp
+++ b/src/gui/virtualstudio.cpp
@@ -66,6 +66,13 @@ VirtualStudio::VirtualStudio(bool firstRun, QObject* parent)
     , m_showFirstRun(firstRun)
     , m_inputMeterLevels(2, 0)
     , m_outputMeterLevels(2, 0)
+    , m_inputComboModel(QJsonArray::fromStringList(QStringList(QLatin1String(""))))
+    , m_outputComboModel(QJsonArray::fromStringList(QStringList(QLatin1String(""))))
+    , m_inputChannelsComboModel(
+          QJsonArray::fromStringList(QStringList(QLatin1String(""))))
+    , m_outputChannelsComboModel(
+          QJsonArray::fromStringList(QStringList(QLatin1String(""))))
+    , m_inputMixModeComboModel(QJsonArray::fromStringList(QStringList(QLatin1String(""))))
 {
     QSettings settings;
     m_updateChannel =
@@ -199,7 +206,6 @@ VirtualStudio::VirtualStudio(bool firstRun, QObject* parent)
 
     m_bufferSize     = settings.value(QStringLiteral("BufferSize"), 128).toInt();
     m_previousBuffer = m_bufferSize;
-    refreshDevices();
     m_previousInput  = m_inputDevice;
     m_previousOutput = m_outputDevice;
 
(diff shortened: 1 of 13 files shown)


mikedickey:
TODO list:
- [ ] Test builds on Windows (with and without ASIO)
- [ ] Test builds on Linux
- [ ] Test on computers with lots of channels available
- [ ] Test using JACK audio backend
- [ ] Test builds without rtaudio available
- [ ] Ensure device, channels and mix mode selections are saved & restored correctly

psiborg112:
That should simplify things. (Basically makes the changes that I suggested above.) Let me know if it's introduced any regressions, but it shouldn't have. In terms of both interfaces being created at the same time, it shouldn't really cause an issue as long as they're just doing basic interface set up in the constructor. (Which is now the case on the qjacktrip side of things with these modifications.) Probably worth double checking things on the VS side at some stage, although given the need for the auth sequence, I suspect it's likely to be all good.

mikedickey:
> > That should simplify things. (Basically makes the changes that I suggested above.) Let me know if it's introduced any regressions, but it shouldn't have. In terms of both interfaces being created at the same time, it shouldn't really cause an issue as long as they're just doing basic interface set up in the constructor. (Which is now the case on the qjacktrip side of things with these modifications.) Probably worth double checking things on the VS side at some stage, although given the need for the auth sequence, I suspect it's likely to be all good.
> 
> I haven't had a chance to look at your changes, but our specific requirement would be that QJackTrip does not scan interfaces at construction time. On windows, this is a very expensive process and we've found many scenarios where it causes things to crash. We need to be able to defer anything involving audio until after construction.

Ahh I think I see what you did now. Much cleaner, thanks!

src/gui/virtualstudio.cpp
```diff
@@ -1224,38 +1234,30 @@ void VirtualStudio::refreshStudios(int index, bool signalRefresh)
 void VirtualStudio::refreshDevices()
 {
 #ifdef RT_AUDIO
-    if (!m_vsAudioInterface.isNull()) {
-        m_vsAudioInterface->closeAudio();
-        setAudioReady(false);
-    }
-
-    refreshRtAudioDevices();
-    validateDevicesState();
```
dhing1024, line 1233: Why did `refreshRtAudioDevices()` and `validateDevicesState()` get moved to within `restartAudio`? Just seems a little weird to me that it wouldn't be handled directly by this function.

mikedickey: I moved these to `startAudio` because:
* both need to be called before the rest of the code in `startAudio` such as `setInputDevice`
* `restartAudio` calls `startAudio` and `restartAudio` is already called from here;  there is no need to do it twice
* `startAudio` is called via other paths than just `refreshDevices`

mikedickey: Note also the removal of `refreshDevices` from the constructor. That was causing it to scan and initialize audio immediately at startup, before it tried to do anything. With this PR, the audio initialization is being deferred until after authentication succeeds, so that we can properly report on what is happening when and have a better shot at tracking down the root cause of issues.

psiborg112:
Sorry, should have said 'basic user interface set up in the constructor' rather than just throwing around 'interface' for multiple concepts. Clear as mud.

PR #1070, opened 2023-06-14
Script environment variables

psiborg112:
Set some environment variables that connection scripts can rely upon.


ntonnaett:
That's a nice feature with a pretty small footprint. IMHO the environment variables should have a prefix like JT_.

psiborg112:
> That's a nice feature with a pretty small footprint. IMHO the environment variables should have a prefix like JT_.

Not a bad idea. Will make the change.

mikedickey (approved):

Issue #1069, opened 2023-06-13
jacktrip-1.9 fails to configure without GUI (meson)


umlaeute:
we've been getting build errors for the Debian packages (that build both a GUI-version of jacktrip and a (pure) cmdline version), as in https://salsa.debian.org/multimedia-team/jacktrip/-/jobs/4225141

the issue is obviously, that in non-GUI mode, there are no resources nor UI to be compiled, but the meson qt modules explicitly requires the `sources` parameter for [`qt.compile_resources`](https://mesonbuild.com/Qt6-module.html#compile_resources) resp. [`qt.compile_ui`](https://mesonbuild.com/Qt6-module.html#compile_ui) to be non-empty.

So the following fails, if `qres` resp `ui_h` are empty: https://github.com/jacktrip/jacktrip/blob/34c50c09f105be5cdfc6ad2d665ad528e7b823ce/meson.build#L234-L236

Since the `qres_files` resp `ui_files` are used later on, my current solution looks like this:
``` 
qres_files = []
if qres.length() > 0
        qres_files = qt.compile_resources(sources: qres)
endif
moc_files = qt.compile_moc(headers: moc_h, extra_args: defines)
ui_files = []
if ui_h.length() > 0
        ui_files = qt.compile_ui(sources: ui_h)
endif
```

if you want, I can do a PR.

ntonnaett:
Thanks for reporting this! Your solution looks like the right thing to do. I would be happy to merge your PR.

psiborg112:
Fixed in #1071

PR #1071, opened 2023-06-14
fix meson.build for pure CLI configurations

umlaeute:
a pure CLI-configuration has no resources nor UI elements, but the qt5-meson module doesn't like this...

Closes: https://github.com/jacktrip/jacktrip/issues/1069


psiborg112:
Was going to say that @ntonnaett is the resident meson expert and he should review this, but then saw that he's already commented on your issue and approved. So, looks good.

psiborg112 (approved):

PR #1086, opened 2023-06-21
Moving connection stats into footer

nwang92:
This moves the connection stats into a new footer component like so:
Screenshot 2023-06-21 at 2 19 47 PM
Screenshot 2023-06-21 at 2 19 56 PM
Screenshot 2023-06-21 at 2 24 28 PM


mikedickey:
I wonder we could shorten "Connection Quality" to just something like "Connection" or "Network" or "Internet"?

Also given that the content of this is variable in length, it may look cleaner if we put it on the left side, swapping with "Connected" status?

mikedickey:
If I'm starting a new studio from the app, I noticed that the footer says "Waiting" while the header says "Starting"

nwang92:
Fixed the "Waiting" -> "Starting" problem also flipped the elements around. Now it looks like this:
Screenshot 2023-06-23 at 10 09 57 AM
Screenshot 2023-06-23 at 10 10 07 AM

mikedickey (approved):

PR #1099, opened 2023-07-14
Switching to external qt builds for all matrix jobs

mikedickey:
Removing all the logic for building qt static from source

New qt builds are provided by https://github.com/jacktriplabs/jacktrip-qt

### DIFF
--- a/.github/workflows/jacktrip.yml
+++ b/.github/workflows/jacktrip.yml
@@ -75,17 +75,6 @@ jobs:
             qt-version: '5.15.10'
             qt-type: 'static'
 
-          - name: Linux-x64-qmake-gcc-shared-no_rtaudio
-            runs-on: ubuntu-22.04
-            system-rtaudio: false
-            bundled-rtaudio: false
-            nogui: false
-            novs: true
-            weakjack: false
-            # jacktrip-path: jacktrip
-            # binary-path: binary
-            build-system: qmake
-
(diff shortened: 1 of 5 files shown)


nwang92 (approved):

ntonnaett:
NetworkAuth isn't used by the "classic" UI?

mikedickey:
> NetworkAuth isn't used by the "classic" UI?

No, this was only be using by the virtual studio code to login to your account via oauth. We revamped the way this works in 1.10 (or 1.9?) and no longer use it there.

I believe classic mode has a mechanism for auth, but it uses raw passwords rather than oauth.

mikedickey:
Disabling vs in the flatpak builds for now, since switching to kde6.2 doesn't seem to be enough on its own, and I know almost nothing about how that works. Going to need to read up on it and address via a separate PR.

For reference, here is the error I get when I try to configure meson:

```Could not find: WebEngineQuick Qt6WebEngineQuick in /usr/lib/x86_64-linux-gnu```

A little googling turned up [this as a possible solution](https://forum.qt.io/topic/138867/create-a-flatpak-package-for-my-qt-app/2), but it seems like it may take some work to figure out how to do that with the GitHub Actions container.

ntonnaett:
The Flatpak fix doesn't seem to be complicated. I can do this.

PR #1101, opened 2023-07-15
Qt6 Flatpak build

ntonnaett:
The Qt6 version of VS needs the WebEngineQuick Qt module that is provided by a base app for Flatpak. Because this base app didn't add the pkgconfig directory for their modules, the Meson build couldn't find those modules. The flatpak manifest adds this directory to the meson configuration as an additional pkgconfig directory.

This PR also consolidates the different flatpak manifests into the build-aux/flatpak/org.jacktrip.JackTrip.json one. This manifest also builds directly from the directory it resides in. No git pulling happening there anymore. There's only one flatpak gha now for release and development that doesn't need Python magic anymore.


mikedickey (approved):
Very nice, thanks for picking this up!

PR #1102, opened 2023-07-16
Adding QML_IMPORT_PATH to flatpak to fix missing qml plugin on Linux

mikedickey:


ntonnaett:
That's great that you found a fix. Is the additional Qt module necessary? If not, I would remove it again.

If you still have problems, the example for the base app also adds `--env=QTWEBENGINEPROCESS_PATH=/app/bin/QtWebEngineProcess` to the flatpak manifest.

ntonnaett (approved):

mikedickey:
> That's great that you found a fix. Is the additional Qt module necessary? If not, I would remove it again.
> 
> If you still have problems, the example for the base app also adds `--env=QTWEBENGINEPROCESS_PATH=/app/bin/QtWebEngineProcess` to the flatpak manifest.

It seems to be working with this env added. I'm pretty sure WebEngineCore is a dependency of WebEngineQuick, so it doesn't really matter either way. But probably better to be explicit than to leave it out.

PR #1104, opened 2023-07-19
NO_VS checks should also include QT_VERSION conditions

nwang92:


psiborg112:
Wouldn't it just be easier to check this once in the build files and set NO_VS to true if we're compiling with a Qt version that's too old? (Especially since we should probably issue a warning that the user is trying to build with an unsupported config?)

psiborg112:
Especially since the NO_VS setting defines which files are being compiled - if it's still false and attempting to compile all of the VS stuff, won't it fail anyway?

mikedickey:
I updated `meson.build` to check for qt6+ when vs is enabled and fail with an error otherwise.

https://github.com/jacktrip/jacktrip/blob/9ea2b5425cb1664a3f0542772ed52790d4462a57/meson.build#L22

Perhaps that is enough?

psiborg112:
I would say so.

PR #1106, opened 2023-07-19
Adding instructions to linux shared binary file README

mikedickey:
about which Qt6 packages are now required on Debian and Fedora based systems

### DIFF
--- a/linux/README.md
+++ b/linux/README.md
@@ -2,6 +2,20 @@
 
 See LICENSE.md for license information.
 
+JackTrip requires that Qt6 is installed on your machine.
+
+For Fedora or RedHat:
+
+```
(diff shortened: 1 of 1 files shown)


mikedickey:
I tested these instructions + latest Flatpak builds on both Ubuntu 22.04 and Fedora 38

ntonnaett:
For what purpose is that file. It doesn't apply for flatpak users and developers have another dependency list of packages in the docs. I would prefer to have everything in the docs.

mikedickey:
> For what purpose is that file. It doesn't apply for flatpak users and developers have another dependency list of packages in the docs. I would prefer to have everything in the docs.

It's included in the (non flatpak) Linux tarball package that people download. Most people will never see or read the docs.

ntonnaett (approved):

PR #1110, opened 2023-07-24
Bump jacktrip version for dev

nwang92:
I don't know if this is what we want it to be, I'm just changing it so it is not equal to `1.10.0` which is the latest released version


mikedickey:
I think we should call it 2.0

mikedickey (approved):

PR #1112, opened 2023-07-27
Refactor Pages and Add Settings For Displaying Recommendations

dhing1024:
Adds some new pages and settings meant to improve the information provided to users on app startup, and to allow them to disable it if desired.

https://github.com/jacktrip/jacktrip/assets/26987971/585f37ce-822a-4b9e-89ca-9afd8b73ef21

Not sure if there's an quick fix for this but some of the page transitions look a little funny because of their positioning on and off the screen.

These two views are different between Windows and non-Windows:

Screenshot 2023-07-27 at 1 45 24 PM
Screenshot 2023-07-27 at 1 45 54 PM

Added a back button to this page, which now shows up right before connecting to a studio.

Screenshot 2023-07-27 at 2 06 58 PM


nwang92 (approved):

nwang92 (approved):

mikedickey (changes requested):
I'm definitely missing the headphone icon. The full page of text seems a bit oppressive.

How about small height screens? Does everything fit OK?

I'm still questioning whether combining two recommendations per page is wise here. My gut is telling me that one per page with a unique illustration for each would help improve the delivery of these very critical messages. Especially with the more explicit "do you want to repeat?" prompt at the end, most people are going to see them once and be done with it..

mikedickey:
Why is it still "configuring audio" at startup? I believe that with the move of this to connecting to studio, it should go away entirely and never even try to open audio interfaces until after you try to connect.

mikedickey:
I feel like we're missing cohesion across these screens. Especially the last one feels disconnected and it may not be obvious that the previous screens are the "recommendations" that it is asking you about.

As opposed to an "intro" page I wonder if having a simple header on all of them would help tie things together? It could just be the JackTrip logo and text such as "Recommendations to help you have the best experience"

mikedickey:
I'm not seeing the "Back" button when I go to connect to a studio

Screenshot 2023-07-28 at 7 49 40 AM

dhing1024:
Took another stab at this. Added a bunch of changes:


https://github.com/jacktrip/jacktrip/assets/26987971/28b66dc6-a288-4868-8a7d-0a4ce6982af6



> I'm definitely missing the headphone icon. The full page of text seems a bit oppressive.
>
> How about small height screens? Does everything fit OK?
>
> I'm still questioning whether combining two recommendations per page is wise here. 

I split them back into 4 pages, making it less text per page. I also picked out some new icons.


> with the move of this to connecting to studio, it should go away entirely and never even try to open audio interfaces until after you try to join/connect.

I think this is adjusted now

> As opposed to an "intro" page I wonder if having a simple header on all of them would help tie things together? It could just be the JackTrip logo and text such as "Recommendations to help you have the best experience" or "Recommendations for success" or even just "Recommendations"

I added a header and it also indicates which page you are on. I called it "Getting Started with JackTrip" since I think that fits a little better.


> I'm not seeing the "Back" button when I go to connect to a studio

Forgot to push the commit!

mikedickey:
I like this update a lot better!

I noticed that it still says "Configuring Audio" after "Logging In".. but I can see in console it's not. Perhaps this is just a minor QML fix?

Could we make the text on the last page "

Remind Me Again Next Time

Would you like to review the getting started recommendations again the next time you start JackTrip?"

Otherwise, the header especially seems to suggest it's going to show them all over again, and that may be all people read

dhing1024:
> I noticed that it still says "Configuring Audio" after "Logging In".. but I can see in console it's not. Perhaps this is just a minor QML fix?

Missed that detail, but should be adjusted now. It'll just show "Logging In..." until the screen moves away.

mikedickey (approved):

Issue #1117, opened 2023-07-29
Fails to build with rtaudio-6.0: error: use of undeclared identifier 'RtAudioError'; did you mean 'RtAudioErrorType'?


yurivict:
```
In file included from jacktrip.p/moc_virtualstudio.cpp:10:
In file included from jacktrip.p/../../src/gui/virtualstudio.h:57:
In file included from jacktrip.p/../../src/gui/vsAudioInterface.h:52:
jacktrip.p/../../src/gui/../RtAudioInterface.h:117:38: error: use of undeclared identifier 'RtAudioError'; did you mean 'RtAudioErrorType'?
    static void RtAudioErrorCallback(RtAudioError::Type type,
                                     ^~~~~~~~~~~~
                                     RtAudioErrorType
/usr/local/include/rtaudio/RtAudio.h:228:6: note: 'RtAudioErrorType' declared here
enum RtAudioErrorType {
     ^
```

Version: 1.10.0
FreeBSD 13.2

mikedickey:
@yurivict yes, that would be the correct new type. Thanks for pointing this out, as I hadn't realized RtAudio 6.0 released. There are more changes required to get it to build & work, which I created [this PR to address](https://github.com/jacktrip/jacktrip/pull/1118).

PR #1118, opened 2023-07-29
Updating rtaudio to 6.0.1 for meson builds

mikedickey:
qmake will still build using rtaudio 5.2.0, and care has been taked with the changes in JackTrip's source code to ensure backwards compatibility

Updated local variable names in RtAudioInterface.cpp to be more consistent, as previously they were using several different schemes in the same methods, which I found to be rather confusing.

Also fixed a bug with failing to build virtual studio support using qmake because the webenginequick module was not included in the project file

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -28,3 +28,9 @@ macos/sign-stuff-withcreds.sh
 
 # windows deploy files
 win/deploy
+
+# meson subproject files
+subprojects/packagecache/
+subprojects/rtaudio-5.*/
+subprojects/rtaudio-6.*/
+subprojects/wingetopt/
--- a/jacktrip.pro
+++ b/jacktrip.pro
@@ -37,6 +37,7 @@ nogui {
     QT += svg
     QT += websockets
     QT += webview
+    QT += webenginequick
     vsftux {
       DEFINES += VS_FTUX
     }
--- a/src/RtAudioInterface.cpp
+++ b/src/RtAudioInterface.cpp
@@ -50,8 +50,7 @@ using std::endl;
 //*******************************************************************************
 void RtAudioDevice::print() const
 {
-    std::cout << "[" << RtAudio::getApiDisplayName(this->api) << " - " << this->apiIndex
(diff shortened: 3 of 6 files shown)


mikedickey:
Fixes #1117

ntonnaett:
I think we should wait to merge this until the soname version is fixed: https://github.com/thestk/rtaudio/issues/405

mikedickey:
> I think we should wait to merge this until the soname version is fixed: [thestk/rtaudio#405](https://github.com/thestk/rtaudio/issues/405)

That doesn't impact us since we link it in directly (no shared rtaudio libs)

ntonnaett:
It does impact us on Linux distributions. If they start using RtAudio 6 before this is fixed they'll run into missing symbols.

mikedickey:
> It does impact us on Linux distributions. If they start using RtAudio 6 before this is fixed they'll run into missing symbols.

Anyone can still build JackTrip using rtaudio 5.2.0 (after these changes). I've preserved backwards-compatibility, and several of the CI jobs continue to use 5.2.0.

Package maintainers for Linux distros are going to be using an existing install of rtaudio anyway. The issue I think applies more to them deciding if or when to uptake rtaudio 6.0 into a distro.  I don't think our uptake of 6.0.0 (while preserving 5.2.0 compatibility) should impact them at all. Both qmake and meson should use whatever is already installed.

I do think we should pick up a fix for that rtaudio issue as soon as it is resolved. I just don't think it blocks anything, and if we are going to uptake 6.0.0 I'd rather do it in the 2.0.0 release (now) than some future minor version. Or wait for 3.0..

mikedickey:
I tested these build artifacts on Windows, Mac OSX, and Linux Fedora, and they all seem to work fine.

mikedickey:
Updated to use rtaudio 6.0.1, which includes fix for https://github.com/thestk/rtaudio/issues/405

nwang92 (approved):

PR #1121, opened 2023-08-02
Updating to qt 6.2.5

mikedickey:
Updates Chromium from 90 to 93

Replaces openssl with schannel on Windows

Also removing scripts/qt -> moved to jacktriplabs/jacktrip-qt repo


mikedickey:
This fixes a bug causing Windows to frequently crash on startup. It was caused by trying to use the linked openssl config with qt when built dynamically. Qt introduced schannel in 5.13 and have since switched over to it in their official builds for windows.

nwang92 (approved):

Issue #1079, opened 2023-06-17
source tarballs do not build


umlaeute:
as of Jacktrip-1.10.0, it is impossible to build from the released sources:

```
FAILED: jacktrip.p/meson-generated_moc_Analyzer.cpp.o 
ccache c++ -Ijacktrip.p -I. -I../../.. -fdiagnostics-color=always -D_FILE_OFFSET_BITS=64 -Wall -Winvalid-pch -Wextra -std=c++17 -g -O2 -ffile-prefix-map=/build/jacktrip-jqMWIe/jacktrip-1.10.0+ds=. -fstack-protector-strong -Wformat -Werror=format-security -Wdate-time -D_FORTIFY_SOURCE=2 -DQT_NETWORK_LIB -isystem/usr/include/x86_64-linux-gnu/qt5/QtNetwork -DQT_CORE_LIB -isystem/usr/include/x86_64-linux-gnu/qt5/QtCore -isystem/usr/include/x86_64-linux-gnu/qt5 -fPIC -pthread -DWAIRTOHUB -DNO_GUI -DQT_OPENSOURCE -MD -MQ jacktrip.p/meson-generated_moc_Analyzer.cpp.o -MF jacktrip.p/meson-generated_moc_Analyzer.cpp.o.d -o jacktrip.p/meson-generated_moc_Analyzer.cpp.o -c jacktrip.p/moc_Analyzer.cpp
In file included from jacktrip.p/moc_Analyzer.cpp:10:
jacktrip.p/../../../../src/Analyzer.h:49:10: fatal error: externals/Simple-FFT/include/simple_fft/fft.h: No such file or directory
   49 | #include "externals/Simple-FFT/include/simple_fft/fft.h"
      |          ^~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
```

that's obviously, because the `externals/Simple-FFT/` directory is empty (as it is a *git submodule*, and the automatically generated source tarballs that GitHub offers do not include any submodules).
it would be great if you could provide a way so we can compile JackTrip from the source tarballs (or some other downloadable artifact) associated with a release.

this is a common problem with *git submodule*s (and one of the reasons i dislike them).
- see https://github.com/obsproject/obs-studio/issues/7407
- see also `git subtree`

[...]

PR #1122, opened 2023-08-02
Moving Simple-FFT into source tree

mikedickey:
so that git submodules are not required to create builds of JackTrip

Also added a `nofeedback` option for meson, so that JackTrip can be built without support for feedback detection, causing it to not use Simple-FFT at all.


mikedickey:
Fixes https://github.com/jacktrip/jacktrip/issues/1079

nwang92 (approved):

PR #1123, opened 2023-08-02
Re-Add Feedback Detection Alert

dhing1024:
Screenshot 2023-08-02 at 12 14 26 PM

### DIFF
--- a/src/gui/Connected.qml
+++ b/src/gui/Connected.qml
@@ -108,12 +108,4 @@ Item {
         id: footer
         anchors.bottom: parent.bottom
     }
-
-    Connections {
-        target: virtualstudio
-
-        function onFeedbackDetected() {
-            feedbackDetectedModal.visible = true;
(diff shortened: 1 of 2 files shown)


nwang92 (approved):

mikedickey:
Love the idea of reusing the audio settings area!

Kind of gets messed up though if it's in a closed state. Maybe we need a check to expand if closed when it triggers?

Screenshot 2023-08-03 at 9 32 20 PM

dhing1024:
Fixed collapsed view to this:
Screenshot 2023-08-04 at 11 49 04 AM

Another thing I noticed - if feedback detection happens while on the "change devices" screen while connected, the popup shows over that area as well. So for the time being I disabled it on that screen as well. (this also means the popup won't be visible when they return to the "connected" screen)

mikedickey (approved):

PR #1130, opened 2023-08-04
Show friendly message when scanning audio devices

mikedickey:
instead of an error about JackTrip being compiled incorrectly

Updated audio warning and error messages spread across 4 components to share a new common DeviceWarning.qml component, so that they look and behave the same way

Screenshot (32)

Screenshot (11)

Screenshot (10)

Screenshot (5)


mikedickey:
I think we still have issues with scanning audio devices on windows, but this temporarily plugs a few holes

nwang92 (approved):

PR #1137, opened 2023-08-10
Add QWebChannel for GUI Control

dhing1024:
Adds necessary code infrastructure needed to facilitate interactions between the web engine view and C++/qml state variables. Includes a new separate class, `VsAppController` to facilitate the state management.

https://github.com/jacktrip/jacktrip/assets/26987971/9db43979-204c-4e4b-86c5-0fb1cac11928

Based on examples that can be found here: https://code.qt.io/cgit/qt/qtwebchannel.git/tree/examples/webchannel?h=6.5. We'll make some adjustments to the device controls UI at the bottom, which will be done in a separate PR.


mikedickey (approved):

meson.build
```diff
@@ -159,11 +160,14 @@ else
 			'src/gui/vsUrlHandler.cpp',
 			'src/gui/vsPermissions.cpp',
 			'src/gui/vsPinger.cpp',
-			'src/gui/vsPing.cpp'
+			'src/gui/vsPing.cpp',
+			'src/gui/websocketclientwrapper.cpp',
```
mikedickey, line 164: The `websocket*` files look to me like an example of how to use the WebChannel library. For example, `websockettransport.h` just defines a single function and slot which we don't appear to use.

Can't we just merge any of the necessary bits into `VsAppController`?

mikedickey (changes requested):

meson.build
dhing1024: > websockettransport.h just defines a single function and slot which we don't appear to use

We do use them indirectly - that is how the socket communicates with the web view. To connect to a qwebchannel, we need a transport object that inherits from `QWebChannelAbstractTransport`. See [here](https://github.com/jacktrip/jacktrip/blob/99cde71d0a33b89ea599902f5ab9d382ca7c3fb9/src/gui/virtualstudio.cpp#L133), [here](https://doc.qt.io/qt-6/qwebchannel.html#connectTo), and [here](https://github.com/jacktrip/jacktrip/blob/99cde71d0a33b89ea599902f5ab9d382ca7c3fb9/src/gui/websockettransport.h#L43).

> Can't we just merge any of the necessary bits into VsAppController?

I did some consolidation by removing `websocketclientwrapper`. I think it might be possible, though it would be pretty unclean in my opinion, to merge it all into `vsAppController`, since the `vsAppController` qobject itself is published to the web channel.

mikedickey: I don't think there is enough reason for `vsAppController` to exist, versus just adding the `deviceControlsCollapsed` boolean as another property of `VirtualStudio`.  I don't believe `websocketclientwrapper` is intended to be used as-is; otherwise, it would be included inside of the WebChannel library. I think it would be better to create our own specialization of `QWebChannelAbstractTransport`, even if it looks a lot like the example. The example seems to do little more than print things to stdout. When using example code, I commonly added a "Based upon code from" note with URL link.

dhing1024: I was thinking that `appctl` would be used for general purpose UI controls, but seeing as you've moved most of the audio stuff out of `VirtualStudio` now, I suppose it's no longer necessary. I took that class out of this PR. As for `WebSocketTransport.h`, I don't understand what special needs we have that warrants a change. This seems perfectly suitable for this use case. I suppose we could rename the class to something else, but other than that I'm not sure what specifically you're looking for.

(Pretty sure this will work as-is, but haven't been able to test it. The changes from #1140 are causing an error for me before I can get to the connected view: 
`RtAudioInterface failed to open stream: RtApi::openStream: output device parameter value is invalid.`)

mikedickey (approved):

PR #1140, opened 2023-08-16
Virtual Studio audio cleanup & refactoring

mikedickey:
Renamed vsAudioInterface to vsAudio and moved everything related to audio from the VirtualStudio class into vsAudio. This eliminated a lot of duplicate code, and strange signal connections across the two.

Various RtAudio cleanup and performance improvements. Updates to RtAudioInterface and vsAudio to reduce audio unnecessary audio device scans and other unnecessary system calls, which can be especially slow (10s of seconds) on Windows machines that have bad audio drivers installed.

Converted most of the raw pointers in VirtualStudio and vsAudio into smart pointers and eliminated unnecessary timers, sleeps and other hacks

Fixed bugs with some of the virtual studio QML paths, especially related to deep links

Added Refresh Devices button to Virtual Studio Change Devices screen, and removing Leave button since we have better paths to leave now. Replacing duplicate code with new DeviceRefreshButton.qml

Removed audioActivated, toggleAudio and restartAudio from virtualstudio/vsAudio since these were redundant and ambiguous.

Renaming JackTrip::sJackStopped to sAudioStopped since this is also needed for RtAudio. Fixing bug where shutdown of RtAudio would deadlock JackTrip's data receiver. This fix allows JackTrip to cleanly stop when audio devices are unplugged while connected.

Updated RtAudioInterface::mStereoToMonoMixer to use a smart pointer

Updated copyright in license file and in the about dialog

Bumping version to 2.0.0-beta2


src/gui/Connected.qml
```diff
@@ -57,29 +58,29 @@ Item {
...
+        let idx = audio.inputComboModel.findIndex(elem => elem.type === "element" && elem.text === audio.inputDevice);
         if (idx < 0) {
-            idx = virtualstudio.inputComboModel.findIndex(elem => elem.type === "element");
+            idx = audio.inputComboModel.findIndex(elem => elem.type === "element");
         }
 
         return idx;
     }
 
     function getCurrentOutputDeviceIndex() {
```
nwang92, line 76: I think we can actually delete `getCurrentOutputDeviceIndex()` and `getCurrentOutputDeviceIndex()` in this file

mikedickey: removed

nwang92:
There's some issues I get when navigating to "Settings" that manifest in different ways:
1. In Settings, select "External Headphones" on macOS and click "Save"
2. Unplug external headphones, then go back into Settings which will result in the following:
Screenshot 2023-08-17 at 9 33 20 AM

<br/>

1. In Settings, select "External Headphones" on macOS and click "Save"
2. Quit the app
3. Unplug external headphones
4. Launch app and go to Settings. Everything seems to work and it detected that the "External Headphones" was missing
5. Click "Cancel"
6. Go back into Settings which will now have an error:
Screenshot 2023-08-17 at 9 32 44 AM

In both cases, the Settings screen gets stuck at "Scanning audio devices...". Manually clicking "Refresh Devices" will make everything come back. I think there's some problem here we don't overwrite the input/output device if the selected one is missing?

src/gui/virtualstudio.cpp
```diff
@@ -1837,41 +832,28 @@ void VirtualStudio::disconnect()
...
+    if (!m_currentStudio.id().isEmpty()) {
+        emit openFeedbackSurveyModal(m_currentStudio.id());
     }
+
+    // cleanup
+    m_currentStudio.setId("");
 }
 
-void VirtualStudio::manageStudio(int studioIndex, bool start)
+void VirtualStudio::manageStudio(const QString& studioId, bool start)
```
nwang92, line 843: I have a feeling this function is pretty much deprecated with the 2.0 work, but in any case it might be helpful to have some guard clause in case `studioId.isEmpty()`?

src/gui/virtualstudio.cpp
```diff
@@ -1900,16 +882,10 @@ void VirtualStudio::manageStudio(int studioIndex, bool start)
     QDesktopServices::openUrl(url);
 }
 
-void VirtualStudio::launchVideo(int studioIndex)
+void VirtualStudio::launchVideo(const QString& studioId)
```
nwang92, line 885: Same comment as above around `isEmpty()` checks

nwang92:
My output volume meter looks fine right when I connected to a studio but after I go into "Devices" and then click "Back", it flickers a lot:

https://github.com/jacktrip/jacktrip/assets/6201822/ab14a1f6-7705-4544-a1d0-b74899a99717

src/gui/Connected.qml
```diff
@@ -93,7 +68,7 @@ Item {
         anchors.bottom: deviceControlsGroup.top
 
         property string accessToken: auth.isAuthenticated && Boolean(auth.accessToken) ? auth.accessToken : ""
-        property string studioId: virtualstudio.currentStudio >= 0 ? serverModel[virtualstudio.currentStudio].id : ""
+        property string studioId: virtualstudio.currentStudio.id
```
nwang92, line 71: After leaving a studio, my camera indicator light is still on so I suspect we may not be setting `virtualstudio.currentStudio.id` to an empty string somewhere

mikedickey:
> There's some issues I get when navigating to "Settings" that manifest in different ways:

Just made a few updates to handle these things better.

src/gui/Connected.qml
mikedickey: I believe I was able to fix this.

src/gui/virtualstudio.cpp
mikedickey: Added checks

src/gui/virtualstudio.cpp
mikedickey: Added checks

nwang92 (approved):

mikedickey:
I have more coming, but this seems to be working pretty well, and better to get it in sooner to minimize potential conflicts.

PR #1141, opened 2023-08-18
Moving VS audio device logic into its own worker thread

mikedickey:


nwang92:
One small thing I'm seeing is that every time I switch something under Settings, everything refreshes:


https://github.com/jacktrip/jacktrip/assets/6201822/c622c0de-60f7-43b3-93f8-b3a7c76471a9



This doesn't happen on the ChangeDevices screen when connected to a studio though

src/gui/vsAudio.cpp
```diff
@@ -365,7 +370,8 @@ void VsAudio::loadSettings()
     m_inMultiplier  = settings.value(QStringLiteral("InMultiplier"), 1).toFloat();
     m_outMultiplier = settings.value(QStringLiteral("OutMultiplier"), 1).toFloat();
     m_monMultiplier = settings.value(QStringLiteral("MonMultiplier"), 0).toFloat();
-    m_inMuted       = settings.value(QStringLiteral("InMuted"), false).toBool();
+    m_inMuted       = false;
```
nwang92, line 373: Are we intentionally keeping this?

mikedickey:
> One small thing I'm seeing is that every time I switch something under Settings, everything refreshes:
...
> This doesn't happen on the ChangeDevices screen when connected to a studio though

Yeah, I'm not a fan of the flicker either, but it is sort of intentional. In particular, it is caused [by this](https://github.com/jacktrip/jacktrip/blob/b570a31be386bbfeb1bfdcf4c9d437e3afce662a/src/gui/AudioSettings.qml#L66)

Your audio is temporarily offline when you change anything. Previously this would just freeze the UI until it was done. Now it displays the "Scanning devices" screen. Probably a modal, or spinny balls, or some other UI thing would look nicer and not cause a flicker.

src/gui/vsAudio.cpp
mikedickey: Yes, I think it's a bug that we were persisting mute state in settings.

In particular it's a problem if you leave a studio while muted, close the app, come back later. When you go to (any of the) AudioSettings you will not be able to make the meters move, because you're muted. So it appears like your audio just isn't working.

nwang92 (approved):

PR #1145, opened 2023-08-19
Making audio error/warning indicators work better

mikedickey:
after changing devices while connected to a studio

Fixed a bug where connecting to a studio from the audio settings setup screen could fail due to audio not being stopped in time. Updated VsAudio methods to add option for certain methods to block to support this and other fixes.

Reworked how device reconnect is handled from the change devices screen while connected to a studio. The JackTrip connection was previously being reset in parallel to audio validation, which could have caused things to not update correctly.

Refreshing devices while connected to a studio would cause device scans and validation in parallel to reconnect, which was even worse and caused crashes on Windows.

Fixing persistence of studio list filters to settings

Updating QML setters to check existing state before updating and emitting signals. Previously, they would emit even when there were no changes.

Added command line shells for msys and cygwin on windows so that JackTrip works better when run from command line in these environments

### DIFF
--- a/src/AudioInterface.cpp
+++ b/src/AudioInterface.cpp
@@ -1004,33 +1004,3 @@ void AudioInterface::setDevicesErrorMsg(errorMessageT msg)
     }
     return;
 }
-
-//*******************************************************************************
-std::string AudioInterface::getDevicesWarningMsg()
-{
-    return mWarningMsg;
-}
-
-//*******************************************************************************
-std::string AudioInterface::getDevicesErrorMsg()
-{
-    return mErrorMsg;
-}
-
-//*******************************************************************************
-std::string AudioInterface::getDevicesWarningHelpUrl()
-{
-    return mWarningHelpUrl;
-}
-
-//*******************************************************************************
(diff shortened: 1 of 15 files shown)


nwang92:
Maybe 50% of the time when I click "Refresh Devices" on macOS where I get disconnected from the server (it kicks me back out to the browse view)

mikedickey:
> Maybe 50% of the time when I click "Refresh Devices" on macOS where I get disconnected from the server (it kicks me back out to the browse view)

I just tried hitting "Refresh Devices" about 100 times on my Mac laptop, and cannot reproduce this. Anything interesting in the logs?

nwang92:
Nevermind, I think I was on a prior commit of this branch. Pulling in the latest changes seems to have fixed it.

mikedickey:
> Nevermind, I think I was on a prior commit of this branch. Pulling in the latest changes seems to have fixed it.

You were right. I was tested a self-hosted studio, and this problem appeared only for managed studios (as far as I can tell). I think it's fixed now.

nwang92 (approved):

PR #1146, opened 2023-08-19
Fix to update UI when VS settings changes are cancelled

mikedickey:


mikedickey:
rebasing this over onto `bugfix/audio-warnings` since fixing it requires changes made in there

PR #1149, opened 2023-08-20
Fixing deadlock conditions involving studio join

mikedickey:
Removing VsQuickView from VsAuth since there seems to be no good reason for it to have visibility into that, and it breaks the pattern of VirtualStudio owning that exclusively.

Removed the authSucceeded and authFailed signals
from VirtualStudio because: (a) these were duplicates from VsAuth, (b) nothing used authFailed, (c) there was only a single qml connection really using
authSucceeded and it was easy to wire that up directly instead, since auth is a context variable, (d) having these in both places and meaning different things was very confusing and made it harder to grok and maintain

Replaced the firstLoad parameter in VirtualStudio::serverList with a m_firstRefresh class member, which is a lot easier to track and understand.

Changing use of m_refreshMutex to be a guard specifically for any access to m_servers or m_refreshInProgress. Fixing a few cases where these were not handled properly.

Fixing various compiler warnings in other source code

### DIFF
--- a/src/Analyzer.cpp
+++ b/src/Analyzer.cpp
@@ -136,7 +136,7 @@ void Analyzer::onTick()
         return;
     }
 
-    const uint32_t buffers = mCircularBufferPtr->size();
+    const uint32_t buffers = uint32_t(mCircularBufferPtr->size());
     const uint32_t samples = buffers * mBufferSize;
 
     // require at least mFftSize values to process, otherwise return
--- a/src/Analyzer.h
+++ b/src/Analyzer.h
@@ -87,9 +87,9 @@ class Analyzer : public ProcessPlugin
     bool testSpectralPeakGrowing();
 
     int mInterval                           = 100;
-    float mPeakThresholdMultipler           = 0.5;
-    float mPeakDeviationThresholdMultiplier = 0.4;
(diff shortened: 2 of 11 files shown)


nwang92 (approved):
This change looks ok to me, but while testing I uncovered a crash when leaving a studio:
```
Stopping JackTrip...
sending exit packet
JackTrip Processes STOPPED!
---------------------------------------------------------
zsh: segmentation fault  ./jacktrip



Thread 0 Crashed:: CrBrowserMain Dispatch queue: com.apple.main-thread
0   jacktrip                      	       0x102930560 VsDevice::stopJackTrip() + 176 (vsDevice.cpp:443)
1   jacktrip                      	       0x102932914 VsDevice::reconcileAgentConfig(QJsonDocument) + 388 (vsDevice.cpp:461)
2   jacktrip                      	       0x102931b5c VsDevice::onTextMessageReceived(QString const&) + 1180 (vsDevice.cpp:605)
3   QtCore                        	       0x103a29aa4 void doActivate<false>(QObject*, int, void**) + 984
4   jacktrip                      	       0x1028970e4 VsWebSocket::textMessageReceived(QString const&) + 52 (moc_vsWebSocket.cpp:163)
```

PR #1151, opened 2023-08-21
Fixing Virtual Studio deeplinks

mikedickey:
Merged vsUrlHandler into vsInit, renamed it to be vsDeeplink for clarity, and reworked much of it to work better


nwang92 (approved):
Tested on macOS and Windows 11, seems to work fine

PR #1154, opened 2023-08-22
Exposing isExiting over webchannel

nwang92:
The exit handler when clicking "x" on the window or force-quitting doesn't always leave property. I'm hoping that by exposing this `m_isExiting` we can do something with it over the web transport channel

### DIFF
--- a/src/gui/virtualstudio.cpp
+++ b/src/gui/virtualstudio.cpp
@@ -411,6 +411,11 @@ bool VirtualStudio::vsFtux()
     return m_vsFtux;
 }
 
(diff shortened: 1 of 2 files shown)


nwang92:
I tried to remove `Q_PROPERTY(bool isExiting READ isExiting NOTIFY isExitingChanged)` but we will need it to handle the case of attempting to close the desktop app while in the middle of connecting. Without this variable to persist the user-exit-event, it can only act upon new close/quit conditions.

mikedickey (approved):

PR #1155, opened 2023-08-22
Fixing startup screen for linux

mikedickey:
Removing the loading screen because linux (possibly certain Qt versions?) were having problems with transitions that happen soon after a window is shown.

VS mode will now use default backend of RtAudio, if supported

Various other fixes for Jack audio and Linux things


nwang92:
On logout I see this and "Reset code" doesn't do anything

Screenshot 2023-08-22 at 2 23 01 PM

nwang92 (approved):

PR #1159, opened 2023-08-24
Moving vsAudio inside vsDevice

nwang92:
Moving the vsAudio pointer inside of vsDevice to manage levels/mute in one object

### DIFF
--- a/docs/changelog.yml
+++ b/docs/changelog.yml
@@ -1,25 +1,20 @@
-- Version: "2.0.0-beta2"
-  Date: 2023-08-22
-  Description:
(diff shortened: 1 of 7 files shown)


mikedickey (approved):
This is so much cleaner!

Note that I resolved the conflict with sending `version` with level updates (I think this should work?)

Also I bumped version and updated changelog when I did that, so we can start cutting release candidate builds.

mikedickey:
FYI I still got "stuck on leave" (with the spinning balls) using this, so that is not solved yet..

nwang92:
I'll look into leave functionality again

Issue #1161, opened 2023-08-24
2.0.0-beta2 doesn't always remember GUI setting after quitting


cchafe:
Seems to be an "every other time" occurrence. 
2.0.0-beta2 linux GUI asset tested on F38.
Run in classic mode, run then quit jacktrip. Relaunch.
Every other time, it will launch in VS mode instead of classic

mikedickey:
It looks like this is a side-effect of running both interfaces in parallel, rather than having them cleanly separated. I'm working on a fix.

cchafe:
Fixed, thanks!

PR #1163, opened 2023-08-24
Fix to preserve the last UI mode that you selected

mikedickey:
Don't call authentication (and other) APIs when you start up in Classic mode. It should now only do this if you've chosen virtual studio mode as your preference (at startup), or when switching to virtual studio mode.

This also adds support for deep links while in classic mode. It will handle the link properly while preserving class mode for the next time you start up JackTrip.


mikedickey:
Fixes #1161

nwang92 (approved):

PR #1166, opened 2023-08-24
Rewriting vswebsocket class to check QWebsocket object properties

nwang92:
Removing `m_connected` and `m_error` because I don't think those are reliable for the lifecycle of a given websocket connection. Instead we can inspect the [state](https://doc.qt.io/qt-6/qwebsocket.html#state) of the given websocket and use that to determine what should happen.

Also putting things inside of a `QScopedPointer` and adding a destructor

### DIFF
--- a/src/gui/vsWebSocket.cpp
+++ b/src/gui/vsWebSocket.cpp
@@ -102,7 +102,9 @@ void VsWebSocket::closeSocket()
 void VsWebSocket::onError(QAbstractSocket::SocketError error)
 {
     // RemoteHostClosedError may be expected due to finite connection durations
-    if (error != QAbstractSocket::RemoteHostClosedError) {
+    // ConnectionRefusedError may be expected if the server-side endpoint is closed
+    if (error != QAbstractSocket::RemoteHostClosedError
+        && error != QAbstractSocket::ConnectionRefusedError) {
         qDebug() << "Websocket error:" << error;
     }
     if (!m_webSocket.isNull()) {
--- a/src/gui/vsWebSocket.cpp
+++ b/src/gui/vsWebSocket.cpp
@@ -49,20 +49,31 @@ VsWebSocket::VsWebSocket(const QUrl& url, QString token, QString apiPrefix,
     , m_apiPrefix(apiPrefix)
     , m_apiSecret(apiSecret)
 {
-    connect(&m_webSocket, &QWebSocket::connected, this, &VsWebSocket::onConnected);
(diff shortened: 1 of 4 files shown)


nwang92:
Added some more handling around `m_studioSocketPtr` which also didn't seem to be properly reconnecting/destroyed.

src/gui/vsWebSocket.h
```diff
@@ -65,16 +67,12 @@ class VsWebSocket : public QObject
     void disconnected();
 
    private slots:
-    void onConnected();
```
mikedickey, line 68: I think `textMessageReceived` and `disconnected` should move in here because they are slots, not signals.

src/gui/vsWebSocket.cpp
```diff
@@ -49,20 +49,31 @@ VsWebSocket::VsWebSocket(const QUrl& url, QString token, QString apiPrefix,
...
+    connect(m_webSocket.get(), &QWebSocket::textMessageReceived, this,
             &VsWebSocket::textMessageReceived);
 }
 
+VsWebSocket::~VsWebSocket()
+{
+    if (isValid()) {
+        closeSocket();
+    }
+    m_webSocket->disconnect();
```
mikedickey, line 70: I don't think it could be null, but probably good to have a `isNull` check around these too

src/gui/vsWebSocket.cpp
```diff
@@ -76,49 +87,50 @@ void VsWebSocket::openSocket()
     req.setRawHeader(QByteArray("APIPrefix"), m_apiPrefix.toUtf8());
     req.setRawHeader(QByteArray("APISecret"), m_apiSecret.toUtf8());
 
-    m_webSocket.open(req);
+    m_webSocket->open(req);
```
mikedickey, line 90: another `isNull` candidate

mikedickey (changes requested):

src/gui/vsWebSocket.h
nwang92: I split this out into explicit signals/slots to make it less confusing. The previous implementation used a signal-to-signal connection.

mikedickey: Sorry, I think I steered you wrong looking at your more recent changes. If the connections are meant to trigger new signals, I think you had it right the first time.

nwang92: I can switch it back

mikedickey (approved):

PR #1167, opened 2023-08-24
Fix for interface not always displaying current audio backend

mikedickey:
Fix for libpng warning: iCCP: known incorrect sRGB profile

Made it a meson configuration error when you try to build with weakjack support but JACK is not available.

If JackTrip is built with support for only a single backend (RtAudio or JACK), don't show the other one as an option under advanced settings.

If only a single audio backend is available, disable the drop-down.

If only 1 channel selection is available for an interface, disable the drop-down and show a different message for input.

If only a single input channel is selected, disable mix mode selector (which should always be "Mono").

Now only singletons are being set as QML context properties


src/gui/vsAudio.cpp
```diff
@@ -99,6 +99,18 @@ VsAudio::VsAudio(QObject* parent)
...
+    if (isBackendAvailable<AudioInterfaceMode::RTAUDIO>()) {
+        if (!jackIsAvailable()) {
+            m_audioBackendComboModel = {"RtAudio"};
+        }
+    } else {
+        if (!jackIsAvailable()) {
+            // shouldn't happen
+            m_audioBackendComboModel = {};
+        }
+        m_audioBackendComboModel = {"JACK"};
```
nwang92, line 111: Should this be in an else branch?

nwang92 (approved):

src/gui/vsAudio.cpp
mikedickey: Yes, although if it hits that "shouldn't happen" case likely other things would break.

PR #1171, opened 2023-08-26
Don't trigger auth succeeded sequence on token refreshes

mikedickey:
This was causing everything to get reset, new device registered, etc. I'm honestly not sure if this works because it's so hard to retrigger the condition, but it seems like this should work..

When connecting to a studio, set enabled flag in local device settings cache. This should help alleviate the problem of it failing to send heartbeats when the initial connection attempt gets refused.

Retry establishing heartbeat webconnection when it fails

Disconnect device if the connection to server is lost

Adding a little more debugging output to help troubleshoot related problems in the future

### DIFF
--- a/src/gui/vsAuth.cpp
+++ b/src/gui/vsAuth.cpp
@@ -116,6 +116,7 @@ void VsAuth::fetchUserInfo(QString accessToken)
 
 void VsAuth::refreshAccessToken(QString refreshToken)
 {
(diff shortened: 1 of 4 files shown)


dhing1024 (approved):
This might work, but I cannot say that with a whole lot of confidence. We may just need to try it out and watch out for it during empirical testing.

PR #1176, opened 2023-08-28
Updating manifests for 2.0.0 release

mikedickey:


dhing1024:
should this go into `dev` or `main`? Don't fully remember which branch is targeted

mikedickey:
> should this go into `dev` or `main`? Don't fully remember which branch is targeted

Older releases use the manifests in `dev`

PR #1178, opened 2023-08-29
Handle invalid device names more gracefully

mikedickey:
Some device names returned by RtAudio, such as RØDE USB microphones, do not use valid utf8 character sequences. This ensures that the device names we get from RtAudio are converted to utf8 prior to comparisons, so they can still be used (even though their names will still look weird).

See also https://github.com/jacktrip/jacktrip/pull/974


mikedickey:
Learning here: always add comments when you are writing code that isn't completely clear

mikedickey:
I think the real bug here is in RtAudio, which seems to be clobbering utf8 when it populates the `std::string`'s for device names

dhing1024 (approved):
The RØDE bug strikes again. That is unfortunate

Issue #1184, opened 2023-08-30
cannot build project without git installed


umlaeute:
i downloaded the soources of the latest and greatest release of jacktrip, installed all the dependencies, but building fails:

```
$ meson setup ../../.. --wrap-mode=nodownload --buildtype=plain --prefix=/usr --sysconfdir=/etc --localstatedir=/var --libdir=lib/x86_64-linux-gnu -Dpython.bytecompile=-1 -Djack=enabled -Dqtversion=6 -Dnogui=true -Drtaudio=disabled
[...]
C++ compiler for the build machine: ccache c++ (gcc 13.2.0 "c++ (Debian 13.2.0-2) 13.2.0")
C++ linker for the build machine: c++ ld.bfd 2.41
Build machine cpu family: x86_64
Build machine cpu: x86_64
Host machine cpu family: x86_64
Host machine cpu: x86_64
Target machine cpu family: x86_64
Target machine cpu: x86_64

../../../meson.build:39:15: ERROR: Program or command 'git' not found or not executable
```

i'm aware that the [documentation](https://github.com/jacktrip/jacktrip/blob/main/docs/Build/Linux.md#ubuntu-and-debianraspbian-qt6) tells me to install `git`, but i don't really see a reason.

meson uses `git` to inject some very optional preprocessor flags https://github.com/jacktrip/jacktrip/blob/75df4267a3f5859c4824dca7d52cac8c10a76366/meson.build#L39-L45 *iff* it detects that we are building jacktrip from a git repository (if it detects that we are not building from a git repository, nothing happens).

since i am not building from a git-repository, there is little use in forcing a `git` installation.

i suggest to handle a missing `git` binary gracefully (the same as if there was a `git` binary that detects that we are not building from a git repository)

umlaeute:
something like the following should do the trick:

```diff
Index: jacktrip-2.0.1+ds/meson.build
===================================================================
--- jacktrip-2.0.1+ds.orig/meson.build
+++ jacktrip-2.0.1+ds/meson.build
@@ -36,12 +36,15 @@ if get_option('buildtype') == 'release'
 	c_defines += ['-DNDEBUG']
 endif
 
-git_tags_cmd = run_command('git', 'describe', '--tags', check: false)
-git_hash_cmd = run_command('git', 'rev-parse', '--short', 'HEAD', check: false)
-if git_tags_cmd.returncode() == 0 and git_hash_cmd.returncode() == 0
+git = find_program('git', required: false)
[...]

Issue #1185, opened 2023-08-30
generated manpage contain useless 


umlaeute:
the manpages generated by the buildsystem with the help of `help2man` have a heading like so:

```
NAME
       JackTrip - manual page for JackTrip VERSION: 2.0.1
```

typically the string after the application name is supposed to hold a terse description of the program.
"manual page for JackTrip VERSION: 2.0.1" is not very useful in this context.

consequently, my build process complains and provides a hint on how to fix this:
```
N:
W: useless-whatis-entry
N: 
N:   The whatis entry for this manual page (the brief description found in the NAME section) is of the form:
N:   
N:    program - manual page for program
N:   
N:   This conveys no information about what the program is for and is repetitive. The short description should contain brief information about what the program is for to aid in searching with apropos and similar programs.
N:   
N:   If this manual page was generated by help2man, use the -n option to provide a more meaningful description.
N: 
N:   Visibility: warning
N:   Show-Always: no
N:   Check: documentation/manual
N:   Renamed from: manpage-has-useless-whatis-entry
N: 
```

umlaeute:
e.g. something like this:

```diff
zmoelnig@xenakis:~/Debian/pkg-multimedia/jacktrip/jacktrip$ git diff meson.build
diff --git a/meson.build b/meson.build
index d4d2c94..9e4f73e 100644
--- a/meson.build
+++ b/meson.build
@@ -290,6 +290,7 @@ if (host_machine.system() == 'linux')
        if help2man.found()
                gzip = find_program('gzip', required: false)
                help2man_opts = [
+                       '--name="high-quality system for audio network performances"',
                        '--no-info',
                        '--section=1']
                manfile = custom_target('jacktrip.1',
```

PR #1186, opened 2023-08-30
Removing need to have git installed for meson builds

mikedickey:
Added a buildinfo flag to set this manually

Adding help2man option to set a synopsis

From @umlaeute


mikedickey:
Fixes #1184 and #1185 

Thanks @umlaeute !

nwang92 (approved):

PR #1189, opened 2023-08-31
Build Linux tarball binary with support for ALSA backend

mikedickey:
Fixes also a bug with compression of Linux manpage

Also printing backend names now instead of RtAudio API ids

### DIFF
--- a/.github/workflows/jacktrip.yml
+++ b/.github/workflows/jacktrip.yml
@@ -440,9 +440,11 @@ jobs:
             CONFIG_STRING="-Drtaudio=enabled $CONFIG_STRING"
           fi
           if [[ "${{ matrix.bundled-rtaudio }}" == true ]]; then 
-            CONFIG_STRING="-Drtaudio:jack=disabled -Drtaudio:alsa=disabled -Drtaudio:default_library=static $CONFIG_STRING"
+            CONFIG_STRING="-Drtaudio:jack=disabled -Drtaudio:default_library=static $CONFIG_STRING"
             if [[ "${{ runner.os }}" == 'Windows' ]]; then 
               CONFIG_STRING="-Drtaudio:wasapi=enabled $CONFIG_STRING"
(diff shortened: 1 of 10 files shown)


nwang92 (approved):

.github/workflows/jacktrip.yml
```diff
@@ -462,9 +450,6 @@ jobs:
           if [[ "${{ matrix.weakjack }}" == true ]]; then 
             CONFIG_STRING="-Dweakjack=true $CONFIG_STRING"
           fi
-          if [[ "${{ runner.os }}" == 'Linux' ]]; then
-            CONFIG_STRING="-Dwerror=true $CONFIG_STRING"
-          fi
```
ntonnaett, line 465: I added this so we don't regress with code quality (at least on Linux). What warning is the reason for the removal?

mikedickey: RtAudio generates warnings when the ALSA backend is enabled. This caused the builds to fail.

mikedickey: Perhaps there is a way to do this without it also applying to subprojects?

ntonnaett: `-Drtaudio:werror=false` should work in our case I think.

mikedickey: Cool thanks, I'll give it a try!

ntonnaett: An alternative would be to add `include_type: 'system'` to the rtaudio dependency, like we did with Qt, but it has additional implications like ordering of build flags.

ntonnaett: Nice, looks like rtaudio:werror worked. :)

mikedickey: Looks like it worked!

nwang92 (approved):

Issue #1179, opened 2023-08-29
Windows 11, JackTrip 2.0, JACK


SynthiaPayne:
UPDATED: Windows 11, JackTrip 2.xx, not working with JACK as the backend. 
RT Audio as the backend works fine with ASIO4ALL or WDM. 
Version 1.10.1 seems to be the last one that works with JACK.

SynthiaPayne:
Win11, JT 2.0 in Classic Mode with backend set to JACK.
JACK set to default (WDM Realtek built-ins) 
No ASIO of any kind

JackTrip 2.0 crashes
JACK messages below

14:58:36.134 Statistics reset.
14:58:39.259 JACK is starting...
14:58:39.259 C:/Program Files/JACK2/jackd.exe -S -X winmme -dportaudio -r48000 -p64
14:58:39.265 JACK was started with PID=7284.
jackdmp 1.9.21
Copyright 2001-2005 Paul Davis and others.
Copyright 2004-2016 Grame.
Copyright 2016-2022 Filipe Coelho.
jackdmp comes with ABSOLUTELY NO WARRANTY
This is free software, and you are welcome to redistribute it
under certain conditions; see the file COPYING for details
Drivers/internals found in : C:\Program Files\JACK2
Drivers/internals found in : C:\Program Files\JACK2
JACK server starting in realtime mode with priority 10
self-connect-mode is "Don't restrict self connect requests"
Cannot connect to named pipe after wait = \\.\pipe\server_jack_default_0 err = 2
Cannot connect to server request channel
jack server is not running or cannot be started
JackShmReadWritePtr::~JackShmReadWritePtr - Init not done for -1, skipping unlock
JackShmReadWritePtr::~JackShmReadWritePtr - Init not done for -1, skipping unlock
jack_deactivate called with a NULL client
jack_deactivate called with a NULL client
jack_client_close called with a NULL client
14:58:41.450 JACK connection change.
14:58:41.451 Statistics reset.
14:58:41.455 Client activated.
14:58:41.455 Patchbay deactivated.
14:58:41.466 JACK connection graph change.
14:58:45.679 JACK connection graph change.
14:58:46.161 JACK connection graph change.
14:59:05.162 JACK connection graph change.
14:59:05.332 JACK connection change.
JackWinSemaphore::TimedWait name = jack_pipe.default_freewheel time_out
JackFreewheelDriver::ProcessSync: SuspendRefNum error
[...]

PR #1191, opened 2023-09-01
Fixing for crash when using the JACK audio backend on Windows

mikedickey:
Jack docs say to use jack_free() not std::free()

Apparently, that matters!


mikedickey:
I believe this fixes #1179

ntonnaett:
Wow! Good catch. Maybe that even fixes some deallocation weirdness I experienced with Pipewire. I have to test.

nwang92 (approved):

SynthiaPayne:
Watching for 2.0.2 to test. Thanks!

PR #1193, opened 2023-09-01
Set the PIPEWIRE_LATENCY environment variable with GUI on Linux

mikedickey:
When using Virtual Studio or QJackTrip GUI on Linux, this sets the environment variable based on GUI settings, so there is no longer any need to set any environment variables or command line parameters for pipewire.

I think this should still work the same when running JackTrip from command line.

Also improving logic for rtaudio backend fallback

### DIFF
--- a/src/gui/qjacktrip.cpp
+++ b/src/gui/qjacktrip.cpp
@@ -930,11 +930,11 @@ void QJackTrip::start()
 
 #ifdef RT_AUDIO
             if (m_ui->backendComboBox->currentIndex() == 1) {
+                unsigned int bufferSize = m_ui->bufferSizeComboBox->currentText().toInt();
+                unsigned int sampleRate = m_ui->sampleRateComboBox->currentText().toInt();
                 m_jackTrip->setAudiointerfaceMode(JackTrip::RTAUDIO);
-                m_jackTrip->setSampleRate(
-                    m_ui->sampleRateComboBox->currentText().toInt());
-                m_jackTrip->setAudioBufferSizeInSamples(
-                    m_ui->bufferSizeComboBox->currentText().toInt());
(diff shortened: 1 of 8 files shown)


nwang92 (approved):

src/gui/qjacktrip.cpp
```diff
@@ -948,6 +948,7 @@ void QJackTrip::start()
                     m_jackTrip->setOutputDevice(
                         m_ui->outputDeviceComboBox->currentText().toStdString());
                 }
+                AudioInterface::setPipewireLatency(bufferSize, sampleRate);
```
ntonnaett, line 951: This would only work if we compile with RtAudio. Let's put this outside of the #ifdef or inside setSampleRate() and setAudioBufferSizeISamples().

mikedickey: I didn't do that because I think the GUI components to set these values inside QJackTrip only work when you are using RtAudio.

ntonnaett: Oh, I thought that's what this change is about. The Flatpak only builds the JACK backend. So for this we need to change the GUI, too.

So you are testing this with the ALSA backend of RtAudio right now?

mikedickey: Yes, I tested both ALSA and JACK in both Classic and VS mode. The limitation here is that you can only CHANGE settings in Classic if you are using RtAudio, but JACK will still work. Both work in VS mode.

cchafe:
RtAudio checks out and I was able to use JACK all the way down to FPP 16
(N.B. pipewire won't go that low unless you set
`pw-metadata -n settings 0 clock.force-quantum 16`
unset with
`pw-metadata -n settings 0 clock.force-quantum 0` )

PR #1210, opened 2023-10-06
Use counter and trigger on consecutive detections to account for false positives

nwang92:
I tried messing with the thresholds here but I got a little lost in the sauce there. This counter mechanism seems to work better empirically though.


mikedickey:
I still don't understand why it was triggering even when I have send volume set to zero (and zero signal), but this does seem to work for the test case.

mikedickey (approved):

Issue #1197, opened 2023-09-05
GUI blank after enabling VirtualStudio


umlaeute:
this started with my Debian packages (of Jacktrip-2.0.1):
- start jacktrip (GUI) - everything looks nice
- click on the <kbd>Virtual Studio</kbd> button
- the GUI gets blank, like so:
![jacktrip](https://github.com/jacktrip/jacktrip/assets/214034/68c83070-6bad-44ab-9076-68ad65320a65)

the really bad thing is that i cannot revert this: closing jacktrip and starting anew gives me the same blank window.

this happened on two different machines running Debian/sid.

i've then downloaded [your Linux/amd64 binaries](https://github.com/jacktrip/jacktrip/releases/download/v2.0.2/JackTrip-v2.0.2-Linux-x64-binary.zip) but again, no GUI is displayed.

umlaeute:
i've also removed the `~/.local/share/jacktrip/` and `~/.config/jacktrip/` folders, but to not avail.

umlaeute:
in the end i've tried building the Jacktrip-2.0.2 Debian packages *without* Virtual Studio support (`-Dnovs=yes`), and lo and behold! the GUI is back.

i've then rebuild the Jacktrip-2.0.1 Debian packages without Virtual Studio support, and the GUI is back as well.


So it appears that the Virtual Studio mode is to blame.

btw, I also get the following on the stderr:


```sh
$ jacktrip
Listening for deep link requests
qrc:/vs/vs.qml:2:1: module "QtQuick.Controls" plugin "qtquickcontrols2plugin" not found 
     import QtQuick.Controls 
     ^
SSL Library:  "OpenSSL 3.0.10 1 Aug 2023"
Log file location: "/home/zmoelnig/.local/share/jacktrip/JackTrip/log.txt"
```

i don't get this when running the novs versions (that work):

```sh
$ jacktrip
RTAudio: scanning devices...
[Jack - 0]: "Tiger Lake-LP Smart Sound Technology Audio Controller Digital Microphone" (2 ins, 0 outs)
[Jack - 1]: "Tiger Lake-LP Smart Sound Technology Audio Controller Headphones Stereo Microphone" (2 ins, 0 outs)
[Jack - 2]: "speech-dispatcher-dummy" (2 ins, 0 outs)

RtApiJack::getDeviceInfo: error determining Jack input/output channels!

[...]

PR #1211, opened 2023-10-07
Added an error dialog displayed when required QML plugins are missing

mikedickey:
Added --classic-gui command line parameter to force classic GUI mode

Added qml6-module-qtquick-window to Linux README.md to account for broken dependencies on Debian

Fixed segmentation fault caused when destructing QNetworkManager during shutdown on OSX. Seems like Qt has bigger bugs underlying that since the only way I could work-around it was to leak it.

### DIFF
--- a/linux/README.md
+++ b/linux/README.md
@@ -13,7 +13,7 @@ dnf install -y qt6-qtbase qt6-qtbase-common qt6-qtbase-gui qt6-qtsvg qt6-qtwebso
 For Debian or Ubuntu:
 
 ```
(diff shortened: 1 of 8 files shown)


mikedickey:
Fixes #1197

nwang92 (approved):

umlaeute:
I haven't actually checked it, but my gut feeling says that this is a bit cryptic for non-devs (who probably have never heard about it, let alone QML).

I have no real suggestion for what actually *should* be written.
Maybe something like:

> JackTrip detected that some modules required for the VirtualStudio mode are missing on your system.
Click "OK" to proceed to classic mode.
>
> Details: JackTrip failed to load the QML view. This is likely caused by missing QML plugins. Please consult ... for possible solutions.

mikedickey:
Sounds good to me. Updated message and behavior of clicking "OK"

Screenshot (443)

PR #1214, opened 2023-10-10
PLC auto tolerance experiments

mikedickey:
PLC tolerance is a receive buffer latency added to account for network jitter. This can be set manually using "-q #" where # is an integer in milliseconds; or, it can be set automatically using "-q auto"

The calculation of auto tolerance requires measuring the time in between receipt of each audio packet. The current approach takes a maximum of these measurements over a slice of time, and then takes the average of all these max measurements since the connection was first established. The longer a connection is established, the less likely it will get updated as conditions change.

Connection quality changes over time, but the current algorithm doesn't adapt well to this. We don't want it to adapt tolerance too fast though, because temporary abnormalities are normal. Changing tolerance too often can also have undesireable effects from the repeated blending together of discontinuities.

This branch uses exponential weighted moving averages (EWMA) to allow for tolerance to adjust over time. How quickly the moving average changes depends on the AutoSmoothingFactor, and finding the best value for that may require more work.

The other significant change is that the moving average is calculated not using maximum measurements, but rather (mean + 3x standard deviations), which mathematically equates to roughly 99.7% of all values. This prevents extremely abnormal measurements from skewing the calculation. A potential problem here is ensuring that the window is large enough to encompass a statistically significant number of measurements.

### DIFF
--- a/src/Regulator.cpp
+++ b/src/Regulator.cpp
@@ -96,10 +96,12 @@ constexpr int NumSlotsMax = 4096;  // mNumSlots looped for recent arrivals
 constexpr double DefaultAutoHeadroom =
     3.0;                           // msec padding for auto adjusting mMsecTolerance
 constexpr double AutoMax = 250.0;  // msec bounds on insane IPI, like ethernet unplugged
-constexpr double AutoInitDur = 6000.0;  // kick in auto after this many msec
+constexpr double AutoInitDur = 1000.0;  // kick in auto after this many msec
 constexpr double AutoInitValFactor =
     0.5;  // scale for initial mMsecTolerance during init phase if unspecified
 constexpr double MaxWaitTime = 30;  // msec
+constexpr double AutoSmoothingFactor =
+    0.01;  // smoothing factor used to calculate long term EWMA for auto mode
 
 // tweak
 constexpr int WindowDivisor = 8;     // for faster auto tracking
@@ -754,9 +756,10 @@ double StdDev::calcAuto(double autoHeadroom, double localFPPdur, double peerFPPd
 {
     //    qDebug() << longTermStdDev << longTermMax << AutoMax << window <<
     //    longTermCnt;
-    if ((longTermStdDev == 0.0) || (longTermMax == 0.0))
+    if ((longTermStdDev == 0.0) || (longTermMean == 0.0))
         return AutoMax;
-    double tmp = longTermStdDev + ((longTermMax > AutoMax) ? AutoMax : longTermMax);
+    // 3 standard deviations = 99.7%
+    double tmp = longTermStdDev * 3 + longTermMean;
     if (tmp > AutoMax)
         tmp = AutoMax;
     if (tmp < localFPPdur)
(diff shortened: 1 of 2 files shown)


mikedickey:
This really could use a lot of testing in different environments, but at least in my limited test cases it seems to shave off several ms of delay when using auto, without any significant degradation in quality.

cchafe:
my testbed is home (comcast) <-> CCRMA (internet2), using iodelay to measure audio RTT
with main branch and dev running on both client and server, it's a rock solid 32ms RTT 
with feature/plc-tolerance-experiment it starts very high,  >70ms, and then comes down very slowly, arriving at 32ms 
after about 2min
(I'll provide more details on the testbed if needed)

cchafe:
further testing, dev branch vs. this one
This test has two roundtrip circuits running as parallel processes (4 jacktrips, 2 iodelays) using the two hosts (client comcast, server internet2)
A 4min plot of iodelay RTT measurements shows pretty similar results (the low 5 ms values are when distortion prevented a good read). Maybe dev branch wins slightly.
devVsTol

mikedickey:
@cchafe please give it another test. The stats being measured when it first starts up were throwing things off. My latest push should fix that.

cchafe:
Second go -- pretty darn close between dev and this branch which seems to be a bit more distorted and ever so slightly more lag. Dev is purple, this one is light blue, and the low values show where iodelay didn't get a good read (due to distortion, dev = 2ms, this = 5ms). 


devVsTol2

cchafe:
dev brach is purple, this one is light blue
bottom values are bad reads in iodelay from distortion (dev = 2, this = 5)
devVsTol3

mikedickey:
@cchafe are you using the same -q autoX values for these comparisons? I made some changes to the defaults for that, mainly because I don't understand why this is necessary. It seems a bit arbitrary and would be better for JackTrip to calculate the best tolerance number for you when this is set to zero.

The other thing that bothers me about dev is that it's using an average of max measurements, which tends to skew overly high for connections that are not perfectly consistent. A rolling average adapts better to the changes in a connection over time. I've made it pretty slow-adapting, but you can tweak this with the smooth factor..

The goal isn't so much to make the numbers for a particular connection lower (which could easily be done by just removing the 2x standard deviations) but rather make them automatically adjust better to a wider variety of connections.

There is also a bug fix in here where the asynchronous nature of things starting up gives pretty wild and unpredictable measurements for the first period. That could have been the root cause of a lot of wacky things I was seeing before..

mikedickey:
Another way perhaps to think of this is replacing the need to specify an extra `mAutoHeadroom` fudge for "auto" -- replacing it with standard deviations measured from the actual connection conditions. So perhaps a better comparison would be this branch with -q auto0 vs dev with -q auto3 (or auto5?)

cchafe:
updated this branch and ran
lightblue dev auto1 (client) <-> auto3 (server)
purplue this branch auto1 (client) <-> auto3 (server)

devVsTol4

then changed settings to
lightblue dev auto3 (client) <-> auto3 (server)
purplue this branch auto0 (client) <-> auto0 (server)

devVsTol5

mikedickey:
@cchafe does the iodelay distortion also correspond to higher glitches & poorer audio quality? That is certainly. It desired and the blue seems to be noticeably heavier.

cchafe:
right, it is distortion flagged by iodelay when it can't hear a good signal coming back 
there's a trade-off between "wait longer for good packets" and "don't wait, invoke burg"
that's what the autoN value sets (N = expected lag, in ms)

mikedickey:
I'm confused why this test is giving more distortion for roughly the same tolerance/latency. Definitely need to narrow this down first, and likely adjust the default `auto` value to ensure it's offering the best balance.

```
purple dev auto1 (client) <-> auto3 (server)
lightblue this branch auto1 (client) <-> auto3 (server)
```

mikedickey:
@cchafe what do you think theoretically about the moving average approach? And about using mean/std deviations versus max? Is there a mathematical reason you went with avg(max) over all time, or did it just seem to work best? Wondering if I should keep trying to tweak this, or am I am just barking up the wrong tree altogether?

PR #1218, opened 2023-10-26
Fix PLC bug when local FPP is smaller than peer FPP

mikedickey:
This also makes mSlots and mIncoming timer the same size (mNumSlots) and simplifies some of the shimming code when local FPP doesn't match peer

This also improves handling for the opposite case: when local FPP is larger than peer FPP. It uses a more robust approach that better accounts for out of order network packets (and sounds better).

Replace `mLostWindow` (which I believe was uninitialized?) with a search through newly arrived packets

Removing hardcoded sample rate of 48000

### DIFF
--- a/src/Regulator.cpp
+++ b/src/Regulator.cpp
@@ -192,8 +192,11 @@ Regulator::Regulator(int rcvChannels, int bit_res, int FPP, int qLen, int bqLen,
     mLastSeqNumOut = -1;
     mPhasor.resize(mNumChannels, 0.0);
     mIncomingTiming.resize(NumSlotsMax);
-    for (int i = 0; i < NumSlotsMax; i++)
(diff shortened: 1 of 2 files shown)


mikedickey:
This fixes the high glitch count but still doesn't sound the same when local FPP < remote FPP AND there is 5% loss simulated on the server

src/Regulator.cpp
```diff
@@ -463,9 +450,6 @@ UNDERRUN : {
 void Regulator::processPacket(bool glitch)
 {
     double tmp = 0.0;
-    if ((glitch) && (mFPPratioDenominator > 1)) {
```
mikedickey, line 466: Not sure what this was for, but it seems to fix the sound issues when there is 5% loss

nwang92 (approved):

PR #1219, opened 2023-10-26
PLC startup and other optimizations

mikedickey:
Drop PLC stats measurements for first period since these are always very skewed by things starting up

Actually wait for DataProtocol to start running before starting audio rather than just sleeping for 1 millisecond and hoping that it works

When calculating PLC tolerance, accomodate poor audio interface clocks that have a wide range callback inconsistency (i.e. realtek), by using the max(pullStats, pushStats)

Allow PLC with auto headroom set to zero

Tick PLC pullStats for underruns and zeros

Add qt6-qtshadertools-devel to required packages for Fedora Linux

### DIFF
--- a/docs/Build/Linux.md
+++ b/docs/Build/Linux.md
@@ -25,7 +25,7 @@ dnf install "pkgconfig(jack)" rtaudio-devel git help2man python3-jinja2
 
 ### Fedora (Qt6)
 ```sh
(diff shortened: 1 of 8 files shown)


nwang92 (approved):

mikedickey:
@cchafe would you be able to give this a run through your iodelay tests? I'm still a bit confused about why the `feature/plc-tolerance-experiment` branch would give more distortion when latency/tolerance is roughly the same. This branch shouldn't change the tolerance calculations, so it will be telling if it also regresses.

PR #1235, opened 2024-01-07
Use two separate RtAudio streams when input != output device

mikedickey:
RtAudio supports the creation of streams that are input-only, output-only, or input + output (duplex). Previously, we always used a single duplex stream even when the input and output devices differ. This is problematic because each device may have its own clock, and trigger callbacks at different points in time.

RtAudio therefore needed to wait for both audio interfaces to trigger callbacks before it could trigger its own, reducing the interval of time that JackTrip has to perform its work. If the delta between the interface callbacks is small enough, relative to the work JackTrip is performing, everything work just fine. However, over time we've added more CPU intensive work in the JackTrip callbacks, such as PLC. If the delta is too short, callbacks get missed causing audio glitches.

By using separate (simplex) RtAudio streams for input and output, we are able to decouple the two audio interfaces from one another, ensuring that JackTrip always has the maximum amount of time available to "do its thing."


mikedickey:
Note that this requires and incorporates a patch for an RtAudio bug as described here https://github.com/thestk/rtaudio/pull/417

mikedickey:
@cchafe note that this appears to solve a problem with bufstrategy 4 (PLC in callback, or no worker thread), where using two different audio interfaces on OSX would cause audio artifacts. With this, I think we can finally do away with the extra queue & worker thread (bufstrategy 3), shaving off an extra buffer or more of latency.

cchafe:
very cool news!

src/gui/virtualstudio.cpp
```diff
@@ -890,6 +896,7 @@ void VirtualStudio::completeConnection()
                 // run PLC without worker (4)
                 buffer_strategy = 4;
             }
+            */
         } else if (buffer_strategy == 5) {
             buffer_strategy = 3;  // run PLC with worker (3)
```
nwang92, line 901: Does this line also need to be changed to `buffer_strategy = 4;`?

mikedickey: I don't think so. Leaving this as-is allows someone to still use the advanced setting to explicitly use a worker queue or not. It only changes the behavior of the "auto" setting so that it's the same as never use a worker.

I considered the alternative of just removing the option altogether so that there was only one PLC setting again, but I'd rather go slower and have more time to test in the wild before removing it entirely.

Also thinking about what happens for upgrade scenarios where the default value has always been 3 (after the +1):
* 3: PLC before auto -> PLC auto -> PLC with "no worker"
* 4: PLC after auto with "use worker" -> "use worker"
* 5: PLC after auto with "no worker" -> "no worker"

Maybe after next release I'd remove all the worker stuff entirely and force migrate everything to 3, but this approach allows us to still explicitly set to one or the other for a little while longer.

nwang92:
There's a few times when opening the app built from this branch where the "Play Test Tone" button doesn't seem to do anything the first time I click it. If I click it again, it plays as expected. I'm not sure if it's because there's still some setup stuff going on but I haven't been able to consistently reproduce this.

Also there are some times when I get this RtAudio error when I change the number of input channels (typically when going from 1 channel, either 1 or 2, to 2 channels):
Screenshot 2024-01-13 at 2 07 47 PM

mikedickey:
> There's a few times when opening the app built from this branch where the "Play Test Tone" button doesn't seem to do anything the first time I click it. If I click it again, it plays as expected. I'm not sure if it's because there's still some setup stuff going on but I haven't been able to consistently reproduce this.
> 
> Also there are some times when I get this RtAudio error when I change the number of input channels (typically when going from 1 channel, either 1 or 2, to 2 channels):

It sounds like your build has the rtaudio bug I had to fix. Wipe out your subprojects/rtaudio-6.0.1 directory, your builddir directory, and re-run meson setup & compile. You can also look at the RtAudio.cpp file to verify that the patch was applied by meson.

nwang92 (approved):

PR #1236, opened 2024-01-13
Bumping version and updating changelog for 2.2.0-beta1

mikedickey:
Also fixing a compiler warning with Settings.h when NO_GUI

### DIFF
--- a/docs/changelog.yml
+++ b/docs/changelog.yml
@@ -1,8 +1,12 @@
-- Version: "2.1.1"
-  Date: 2023-12-xx
+- Version: "2.2.0-beta1"
+  Date: 2024-01-16
   Description:
+  - (updated) Improved support for different input and output devices
+  - (updated) Various latency improvements for packet loss concealment
+  - (updated) VS Mode make it easier to dismiss the user feedback dialog
   - (updated) VS Mode error message for disconnected audio interfaces
   - (fixed) VS Mode refused to connect to studios not 48khz
+  - (fixed) VS Mode showed wrong devices selected when connected
 - Version: "2.1.0"
   Date: 2023-11-06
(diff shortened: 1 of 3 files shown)


nwang92 (approved):

mikedickey:
@ntonnaett any idea why the flatpak build is failing for beta releases?

ntonnaett:
> @ntonnaett any idea why the flatpak build is failing for beta releases?

It seems appstreamcli fails when using xvfb-run. I'm not sure why it tries to run an x11 application. There is no change from our side I could find, that could be responsible for that. Maybe the github actions container changed or the github action module we use.

mikedickey:
> > @ntonnaett any idea why the flatpak build is failing for beta releases?
> 
> It seems appstreamcli fails when using xvfb-run. I'm not sure why it tries to run an x11 application. There is no change from our side I could find, that could be responsible for that. Maybe the github actions container changed or the github action module we use.

What is especially odd is that it's running fine right now on dev. Only if I change the version string in jacktrip_globals.h does it break.

ntonnaett:
#1238 updates the flatpak-builder module and builds fine. If we merge this and it still doesn't work with this one, it's probably not the action itself. Maybe different behavior because of a different trigger.

docs/changelog.yml
```diff
@@ -1,8 +1,11 @@
-- Version: "2.1.1"
-  Date: 2023-12-xx
+- Version: "2.2.0"
+  Date: 2023-01-16
```
ntonnaett, line 2: 2024

mikedickey:
@ntonnaett Thanks, you called it! Turned out to be the ampersand in the changelog.yml.

PR #1240, opened 2024-01-16
Trying to fix flatpak build failure on version change

mikedickey:
Nothing different here except version string


ntonnaett:
Maybe it's the ampersand...

Issue #1208, opened 2023-10-04
URI scheme of 'jacktrip://join/SERVER_ID'


umlaeute:
@dhing1024 said in https://github.com/jacktrip/jacktrip/issues/802#issuecomment-1451100325 
>   It's a URL of the form `jacktrip://join/SERVER_ID`, where `SERVER_ID` is the Virtual Studio server ID in `https://app.jacktrip.org/studios/SERVER_ID`. It's used so that, for example, a Virtual Studio user can click a button in the browser web app to automatically open the installed software app and connect to a studio.

now i wonder: was this URI a typo?
following wikipedia's definition of an [URI](https://en.wikipedia.org/wiki/Uniform_Resource_Identifier), we can be break this URI into three components:
1. schema/protocol `jacktrip`
2. host `join`
3. path `SERVER_ID`

intuition tells me, that the `host` part is weird. typically the `host` are network addresses, or as the wikipedia article puts it:
> A host subcomponent, consisting of either a registered name (including but not limited to a [hostname](https://en.wikipedia.org/wiki/Hostname)) or an [IP address](https://en.wikipedia.org/wiki/IP_address).

i would expec the proper URI to be `jacktrip://app.jacktrip.org/join/SERVER_ID`, or (for using the default server) `jacktrip:///join/SERVER_ID` without any host; i don't think there's an RFC-defined semantics of an URI with an empty hostname, but it's used occasionally (most prominently in the `file:///tmp/yikes.txt`)

(yes there's a loophole in the "not limited to")

mikedickey:
I think you are right that the "//" implies an "authority" and so for example "jacktrip:join/<SERVER_ID>" (no authority) would be more comforming.

If anyone ever wants to extend this feature to support additional capabilities (such as "jacktrip://<HOST_OR_IP>/join" to join a server, that would probably be a good time to update this. We'd have to confirm that it is still works with web browsers, though, as I'm not sure if they have more specific requirements for deep link processing?

PR #1261, opened 2024-02-09
Fix for unreliable deeplink handling on Windows

mikedickey:
Deep links were failing on Windows about 5% of the time. Console showed the following messages when it happened:

Never ready to read
Not ready and no bytes available

It seems that QLocalSocket::flush() does bad things, and docs for it suggest using waitForBytesWritten() instead. That seems to fix the problem.

I also cleaned up some of the error handling, and made it support the format jacktrip:join/<studio_id> which is more comformant with the URI standard.


mikedickey:
Fixes #1208

nwang92 (approved):

Issue #1251, opened 2024-01-28
build dependencies incomplete on ubuntu 22.04 qt6


allroys:
I tried to build the from source on Ubuntu 22.04, and the dependencies listed at https://jacktrip.github.io/jacktrip/Build/Linux/ weren't sufficient to fully build and run it. My QT6 install was still lacking some "qtquickcontrols2" packages/plugins. I went and downloaded the prebuilt binary, and the readme listed more dependencies, and installing them solved the problem. Suggest updating the build instructions with all the dependencies from the readme.

mikedickey:
Is it just this package that is missing @allroys ? `libqt6quickcontrols2-6`

allroys:
I can't be sure, but that's likely the culprit. The error I was getting was
related to quickcontrols2.

PR #1262, opened 2024-02-09
Adding all the Ubuntu Qt6 dependencies for runtime

mikedickey:
to the commands for building on Linux. Some of these things are duplicated or pulled in as nested dependencies, but probably better to be thorough and ensure that you can run what you build.


mikedickey:
Fixes #1251

nwang92 (approved):

Issue #1257, opened 2024-02-02
Make UDP waiting too long (more than 30ms) configurable


kodidu:
Hello!

I am using jacktrip for remote desktop use because jacktrip's multiple platform support is the best!
My remote desktop connection is not the best regarding latency, but easily enough for voice over ip.
So I get a lot UDP waiting too long (more than 30ms) messages.
This 30ms is hard-coded which is very bad practice:
https://github.com/jacktrip/jacktrip/blob/main/src/UdpDataProtocol.cpp#L803

May I ask if it is possible to make this 30ms configurable e.g. via command line argument?

Thanks in advance!

mikedickey:
I don't believe this does anything other than print an error message? Seems like a strange thing to have a configurable parameter for to me. Perhaps we should just change it to using `qDebug()` instead of `cerr`?

kodidu:
yes, it is only a message. On the UI version this message is shown in the bottom statusbar.
Using the commandline version the console is spammed with this message if the latency is bad.

> Perhaps we should just change it to using qDebug() instead of cerr?
maybe?

mikedickey:
> yes, it is only a message. On the UI version this message is shown in the bottom statusbar. Using the commandline version the console is spammed with this message if the latency is bad.

the console spam is definitely annoying. maybe use a #define and throttle the console error logging to at most 1 per second?

PR #1264, opened 2024-02-09
Adding throttle for UDP waited too long errors

mikedickey:
So that it only logs to console once per gap rather than every 30ms

Also replacing local variables and constants for 30ms with a new global gUdpWaitTimeout


mikedickey:
Fixes https://github.com/jacktrip/jacktrip/issues/1257

nwang92 (approved):

PR #1267, opened 2024-02-11
Use correct number of channels for the selected CoreAudio stream

mikedickey:
Patching RtAudio bug where OSX users of some audio devices with high channel counts (including RME) could experience garbled audio

Also hiding "Devices" label when Jack audio backend is selected

### DIFF
--- a/src/gui/DeviceControlsGroup.qml
+++ b/src/gui/DeviceControlsGroup.qml
@@ -123,6 +123,7 @@ Rectangle {
                 }
 
                 Item {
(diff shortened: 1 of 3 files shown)


mikedickey:
See https://github.com/thestk/rtaudio/pull/422

mikedickey:
I'm going to merge this since it fixes a pretty ugly audio bug, and was accepted and merged upstream in rtaudio

nwang92 (approved):

PR #1268, opened 2024-02-12
Updating domain name for virtual studio endpoints

mikedickey:
app.jacktrip.org => app.jacktrip.com
test.jacktrip.org => test.jacktrip.com


mikedickey:
Depends on other backend changes

nwang92 (approved):

PR #1269, opened 2024-02-18
Adding files for building a docker image

mikedickey:
Uses jacktrip/jack base image built here https://github.com/jacktriplabs/jack-container

This replaces the old repo at https://github.com/jacktrip/jacktrip-container

Fixing compiler warning for guiEnabled in Settings.h

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -47,5 +47,6 @@ COPY linux/container/jacktrip.service /etc/systemd/system/
 COPY --from=builder /lib64/libQt5Core.so.5 /lib64/libQt5Network.so.5 /lib64/
 COPY --from=builder /root/builddir/jacktrip /usr/local/bin/
 
-# jacktrip hub server listens on 4464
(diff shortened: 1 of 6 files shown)


Dockerfile
```diff
@@ -0,0 +1,47 @@
...
+
+# install jacktrip service
+COPY linux/container/jacktrip.service /etc/systemd/system/
+
+# copy the artifacts we built into the final container image
+COPY --from=builder /lib64/libQt5Core.so.5 /lib64/libQt5Network.so.5 /lib64/
+COPY --from=builder /root/builddir/jacktrip /usr/local/bin/
+
+# jacktrip hub server listens on 4464
+EXPOSE 4464/tcp
```
nwang92, line 47: I could be wrong but I thought the hub server also needs to expose a set of UDP ports too?

nwang92 (approved):

Dockerfile
mikedickey: I believe EXPOSE is more for documentation, but I added it

PR #1287, opened 2024-04-21
Updating PLC to run using peer instead of local FPP

mikedickey:
Adjustments for FPP differences are made at pull time rather than when packets are being pushed. This enables us to perform more targetted predictions (using more "truth" and having smaller "glitches") when packets are missing and peer's FPP is larger than our own.

Removing PLC Regulator worker thread. After optimizations made previously, there is no longer any benefits to using this versus running things inside the callback. At this point it was just dead code, not worth updating for these changes.

Updated calculation of pushStats (packets arriving from peer) to be calculated at pull time, and to use previously known timestamps to calculate deltas. The previous approach did not handle out of order packets, which could significantly skew the stats lower than they should be.

Optimized memory usage for mSlots to reduce fragmentation

### DIFF
--- a/docs/changelog.yml
+++ b/docs/changelog.yml
@@ -1,3 +1,11 @@
+- Version: "2.3.0-beta1"
+  Date: 2024-04-xx
+  Description:
+  - (fixed) PLC audio degradation when buffer sizes don't match
+  - (fixed) PLC statistics for auto latency adjustments
+  - (updated) VS Mode help links to point to support.jacktrip.com
+  - (updated) Removed worker thread for PLC buffer strategies
+  - (updated) Improved PLC performance and quality
 - Version: "2.2.5"
   Date: 2024-03-28
   Description:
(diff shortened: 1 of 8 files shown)


mikedickey:
Try testing this with steady loss set on the server's interface. Use a client with this branch (server version doesn't matter), and test:

* client and server buffer sizes match (baseline)
* client buffer size (dev or recent release) is greater than server
* client buffer size (with this branch) is greater than server

You should hear a significant drop in quality from 1->2, but 1 & 3 should be identical (higher quality, and probably lower cpu) since it's using more "truth" and performing smaller PLC predictions for glitches.

cchafe:
I tested using wifi as my impairment

1)    client and server buffer sizes match (baseline)
 2)   client buffer size (dev or recent release) is greater than server
 3)   client buffer size (with this branch) is greater than server
 4)   client buffer size (with this branch) is less than server

You should hear a significant drop in quality from 1->2 
-- yes, and outgoing client->server stream is less consistent (possibly load related)

but 1 & 3 should be identical (higher quality, and probably lower cpu) since it's using more "truth" and performing smaller PLC predictions for glitches
--yes, working, and improves outgoing stream

condition 4) has a problem that results in a loud buzz intermittently that is not present in dev branch

mikedickey:
> condition 4) has a problem that results in a loud buzz intermittently that is not present in dev branch

Thanks, I think this is fixed by my latest push.

cchafe:
That did the trick!

cchafe (approved):

PR #1288, opened 2024-04-29
plc meld with peer fpp

cchafe:
Improved PLC algorithm and handling as a direct replacement for Regulator. Earlier version was tested as a subclass of Regulator and used --bufstrategy 5 to be able to compare algorithms.

Regulator still adjusts headroom, but that may become vestigial if this works as intended.

Current handling prints average Burg execution time taken and needs to be below  buffersize duration in ms (for example, this run is 128 FPP and is below 2.66 ms)
`avg 0.748308 glitches 8 `
`avg 0.761231 glitches 2 `
`avg 0 glitches 0 `
`avg 0 glitches 0 `
...it helps for now but should become a verbosity option.

**meson setup --reconfigure --buildtype=release**
has a large effect on keeping the average low


cchafe:
looks like clang-format under Fedora 40 may be treating the project differently than what I was getting the day before under Fedora 38... something's up anyway

src/Regulator.cpp
```diff
@@ -134,14 +282,25 @@ Regulator::Regulator(int rcvChannels, int bit_res, int FPP, int qLen, int bqLen,
...
-    //                  << " larger than max channels = " << MaxChans << "\n";
-    //        exit(1);
-    //    }
+    if (gVerboseFlag)
+        cout << "mPacketsInThePast = " << mPacketsInThePast << " at " << mFPP << "\n";
+
+    mPcnt = 0;
+    mTime = new Time();
+    mTime->start();
+    mUpToNow   = mPacketsInThePast * mFPP;        // duration
```
mikedickey, line 291: seems like all of these should be initialized in `setFPPratio` and use `mPeerFPP` instead?

src/Regulator.h
```diff
@@ -176,15 +229,33 @@ class Regulator : public RingBuffer
     virtual bool getStats(IOStat* stat, bool reset);
 
    private:
-    void shimFPP(const int8_t* buf, int seq_num);
+    // !peerFPP   void shimFPP(const int8_t* buf, int seq_num);
```
mikedickey, line 232: maybe remove this?

mikedickey (changes requested):
Wow lots of formatting changes! I'm just going to assume that the "*dsp" code is unchanged. Also just doing a general review since I don't claim to understand all the new burg code! ;)

For some reason, this is sounding far more glitchy to me than previous versions.

mikedickey:
Not sure if any of these are significant?
```
../src/Regulator.cpp(117): warning C4267: '=': conversion from 'size_t' to 'int', possible loss of data
../src/Regulator.cpp(169): warning C4305: '=': truncation from 'double' to 'float'
../src/Regulator.cpp(320): warning C4305: '=': truncation from 'double' to 'float'
```

cchafe:
I didn't see these kinds of warnings
../src/Regulator.cpp(169): warning C4305: '=': truncation from 'double' to 'float'
Where should I check?

src/Regulator.h
```diff
@@ -128,12 +181,14 @@ class Regulator : public RingBuffer
...
    public:
     /// construct a new regulator
-    Regulator(int rcvChannels, int bit_res, int FPP, int qLen, int bqLen,
+    Regulator(int rcvChannels, int bit_res, int localFPP, int qLen, int bqLen,
               int sample_rate);
 
     // virtual destructor
     virtual ~Regulator();
 
+    Time* mTime;
```
mikedickey, line 190: This should probably move down to private variables section

mikedickey: Also it should be initialized to NULL either in header or constructor, to avoid delete's of initialized memory (null is ok)

src/Regulator.h
```diff
@@ -176,48 +231,57 @@ class Regulator : public RingBuffer
...
     int mNumChannels;
     int mAudioBitRes;
-    int mFPP;
+    int mLocalFPP;
     int mPeerFPP;
     int mSampleRate;
+    int mPcnt;
+    std::vector<float> mTmpFloatBuf;
+    std::vector<Channel*> mChanData;
+    BurgAlgorithm* ba;
```
mikedickey, line 258: any raw pointers should be initialized to NULL either in header or constructor, to avoid delete's of initialized memory (null is ok)

src/Regulator.cpp
```diff
@@ -134,16 +285,8 @@ Regulator::Regulator(int rcvChannels, int bit_res, int FPP, int qLen, int bqLen,
...
-    // hub client rcvChannels is set from client's settings parameters
-    // hub server rcvChannels is set from connecting client, not from hub parameters
-    //    if (mNumChannels > MaxChans) {
-    //        std::cerr << "*** Regulator.cpp: receive channels = " << mNumChannels
-    //                  << " larger than max channels = " << MaxChans << "\n";
-    //        exit(1);
-    //    }
-    if (mFPP > MaxFPP) {
-        std::cerr << "*** Regulator.cpp: local FPP = " << mFPP
+    if (mLocalFPP > MaxFPP) {
```
mikedickey, line 288: seems like this check should maybe be for `mPeerFPP`, after it's set in `setFPPratio`?

src/Regulator.cpp
```diff
@@ -275,50 +382,36 @@ void Regulator::setFPPratio(int len)
...
     mFadeUp.resize(mPeerFPP, 0.0);
     mFadeDown.resize(mPeerFPP, 0.0);
     for (int i = 0; i < mPeerFPP; i++) {
         mFadeUp[i]   = (double)i / (double)mPeerFPP;
         mFadeDown[i] = 1.0 - mFadeUp[i];
     }
-    mLastWasGlitch = false;
-    mNumSlots      = NumSlotsMax;
+
+    mNumSlots = NumSlotsMax;
```
mikedickey, line 414: seems like we can get rid of the variable `mNumSlots` now that it's effectively a constant equal to `NumSlotsMax`

mikedickey (changes requested):

src/Regulator.cpp
cchafe: that was a placeholder but let's do

src/Regulator.cpp
cchafe: old check, I'll remove it -- unless we fear devices that are greater than 1024

src/Regulator.h
cchafe: sprinkled = nullptr througout the .h

cchafe:
this should wrap it up, but one of the macOS checks failed... not sure if that's me (adding cfloat?) or something coincidental and unrelated

mikedickey:
> this should wrap it up, but one of the macOS checks failed... not sure if that's me (adding cfloat?) or something coincidental and unrelated

This seems to work for me. I kicked the job -- looks like it was just homebrew being flaky.

I have a few suggestions but I'll open a new branch for that.

mikedickey (approved):

PR #1291, opened 2024-04-30
Bug/formatting

cchafe:
scope out what's going on with clang-format versions


cchafe:
works as it should, test completed

PR #1293, opened 2024-05-03
Few more tweaks for plc meld

mikedickey:
Moved floatBufToXfrBuffer() into processPacket(). I think ZERO_OUTPUT is the only block in pullPacket() that doesn't already call this, and seems like just zeroing the buffer is correct for that case.

Fixing some compiler warnings about unused variables

Fixing qt moc warning about Regulator.h not being a Qt class

Changelog updates for 2.3.0-beta1 release


cchafe (approved):
Good stuff.

cchafe:
thanks -- tested it across a wide range -- FPP 32 - 512

PR #1294, opened 2024-05-04
Updates for PLC variable headroom calculations

mikedickey:
The previous approach for calculating auto headroom was overly aggressive at increasing it, in particular counting instances where the packets never arrived and therefore couldn't be saved by a larger headroom value.

This approach only increases headroom when there were actual instances of packets skipped that would have not been skipped if the headroom was 1ms larger.

Allow headroom to grow up to value of long term max versus arbitrary 5ms max.

Updated push lastMin and lastMax and dsp elapsed stats to track values over an entire time interval rather than just the most recent window observed. This provides a more accurate picture for testing and comparing different approaches.

Don't advance mLastSeqNumOut unless we actual use it; otherwise, it could throw other things off.

Fixed bug in Stats where the first measurement of a window was always ignored.

meson use -DQT_NO_DEBUG for non-debug builds (includes release buildtype)

Move once per second stats into Regulator

Fix some of the cout formatting in Regulator

Changelog tweaks

### DIFF
--- a/docs/changelog.yml
+++ b/docs/changelog.yml
@@ -5,10 +5,10 @@
   - (updated) Improved PLC performance and quality
   - (updated) Improved auto latency adjustments for PLC
   - (updated) Removed worker thread for PLC buffer strategies
-  - (updated) VS Mode help links go to support.jacktrip.com
-  - (updated) VS Mode manage button goes to new studio dashboard
+  - (updated) VS Mode help links now go to support.jacktrip.com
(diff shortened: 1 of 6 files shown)


mikedickey:
Note this also preserves arrival time of packets in `mIncomingTiming[]` so that it's easy to adjust and compare things later

cchafe:
merging!

cchafe (approved):

PR #1296, opened 2024-05-07
Conditionally use a PLC worker thread when predictions take too long

mikedickey:
Using peerFPP for PLC predictions has many advantages, but one disadvantage can come into play if peerFPP is much larger than localFPP. The time it takes to run a prediction for PLC is proporational to the size of FPP, and if it ends up taking longer than the local audio callback interval, it will cause xruns.

The same problem exists for example, if peer is sending many channels via a single stream. The time it takes for a PLC prediction is also proportional to the number of channels, so this problem can occur even when localFPP matches or is greater than peerFPP.

This PR resurrects the PLC/Regulator worker thread and queue to help handle these situations better. However unlike before it is not always enabled or disabled via a configuration setting, but rather starts out disabled and is only utilized if a prediction takes longer than a percentage of the local audio callback time (currently 70%).


mikedickey:
This also moves initialization of variables from cpp to header file to be more consistent, cleaner and easier to manage, and removes dead code from the JackTrip, JackTripWorker, and UdpHubListener class which is no longer used.

mikedickey:
This is ready but just blocked on merging https://github.com/jacktrip/jacktrip/pull/1294

cchafe (approved):

PR #1298, opened 2024-05-08
Feature/plc auto worker merge fix

mikedickey:
Fix for merging things in the wrong order


mikedickey:
Same thing as before.. as careful as I tried to be I merged things in the wrong order

PR #1300, opened 2024-05-09
Start P2P server on system without IPv6 support

mawe42:
When I start jacktrip in P2P server mode on a Linux system without IPv6 support, jacktrip exits without any apparent error:
```
# jacktrip -s -V
Verbose mode
Settings:startJackTrip before new JackTrip
Settings:startJackTrip before mJackTrip->startProcess
step 1
  JackTrip:startProcess before checkIfPortIsBinded(mReceiverBindPort)
Stopping JackTrip...
JackTrip Processes STOPPED!
---------------------------------------------------------
step 6
jmain before app->exec()
```

The reason is that `JackTrip::checkIfPortIsBinded` tries to bind the UDP port for IPv4 and IPv6. It fails for IPv6, because the Linux system has no `ipv6` module loaded.

This patch adds a check for `UnsupportedOperation` errors into that method, so that failures to bind to an IPv6 port is ignored.

P.S.: That I don't see the proper "Could not bind ... UDP socket..." error message is another problem that I will open a separate issue for.


mawe42:
Thanks for the review and merge!

PR #1307, opened 2024-05-19
Fix crash in Regulator if destructed before initialization

mikedickey:
This can happen if a connection fails / no initial packet from peer


cchafe (approved):
thanks!

PR #1318, opened 2024-08-17
Decoupling QJackTrip/classic and Virtual Studio GUIs

mikedickey:
This PR accomplishes a few related goals:

- eliminates all dependencies of the QJacktrip/classic GUI code on the Virtual Studio GUI code, and vice-versa
- moves all the logic related to graphical interfaces from main.cpp into a new UserInterface class, while making it cleaner and easier to follow. There's still a lot of #ifdefs in UserInterface.cpp, but I tried to minimize them and also avoid any of the deep nesting (see old main.cpp) that made the code so hard to follow.
- moves all GUI setup logic specific to QJackTrip into there (still in src/gui directory), and moves all logic specific to Virtual Studio into there
- moved all Virtual studio specific code into a new src/vs subdirectory (QJackTrip/classic mode code remains in src/gui)
- fixes some incorrect headers for code that was developed by JackTrip Labs, Inc.
- adds a new "noclassic" meson build parameter for synchronicity with "novs" ("novs" and "nogui" should work the same as before)
- updates license for all Virtual Studio code to MIT (the same used by original project and core library), to eliminate copyleft issues related to GPL (note that MIT is still "GPL compatible", but it's also a lot more friendly and compatible with other licenses)

By re-releasing all the virtual studio code under an MIT license, my hope is to encourage more people to get involved by reusing and helping to enhance it, embedding it in their own projects, and helping incorporate any additional new or existing functionality (whether you choose to log in or not), without all the issues that are associated with copyleft and GPL. These changes also make it easier for others who may want to create and add their own GUIs for JackTrip, which are unrelated to the existing ones that we provide.

It's a lot, but I believe long overdue work to clean things up and improve code hygiene. Note that most of the "files changed" are just header fixes or file moves related to cleaning all this stuff up. I made a special effort to minimize changes to existing code. I believe it all still works as before, but it could certainly use more eyes and testing.

@psiborg112 I tested everything I could think of, but this would benefit especially from your eyes to make sure I didn't break anything with QJackTrip. Note that I moved the "NoNap" class into core library and updated its license to MIT, which is a change that I think would require your approval.


nwang92:
I hit a crash when attempting to start + join an existing studio within the desktop app:
```
Thread 0 Crashed:: CrBrowserMain Dispatch queue: com.apple.main-thread
0   libsystem_kernel.dylib        	       0x19cfb55f0 __pthread_kill + 8
1   libsystem_pthread.dylib       	       0x19cfedc20 pthread_kill + 288
2   libsystem_c.dylib             	       0x19cefaa30 abort + 180
3   jacktrip                      	       0x1010c97b8 Analyzer::onTick() + 1256
4   QtCore                        	       0x1024086f0 void doActivate<false>(QObject*, int, void**) + 984
5   QtCore                        	       0x1024163b8 QTimer::timeout(QTimer::QPrivateSignal) + 56
6   QtCore                        	       0x1024046c4 QObject::event(QEvent*) + 100
7   QtWidgets                     	       0x103e53340 QApplicationPrivate::notify_helper(QObject*, QEvent*) + 236
8   QtWidgets                     	       0x103e54194 QApplication::notify(QObject*, QEvent*) + 528
9   QtCore                        	       0x1023d8924 QCoreApplication::notifyInternal2(QObject*, QEvent*) + 180
```
the crash log doesn't seem specific to the changes here but just noting

meson.build
```diff
@@ -223,8 +233,6 @@ else
 		]
 		moc_h += [
 			'src/dblsqd/feed.h',
-			'src/dblsqd/release.h',
-			'src/dblsqd/semver.h',
```
nwang92, line 227: Don't we need to include these header files?

mikedickey: `moc_h` is only for header files that include classes which derive from `QObject`. It tells the Qt tooling to inject a bunch of other things into the source code before compilation. Neither of these classes derive from `QObject` so it's unnecessary to moc them. In fact, meson currently gives warnings because they are included and there is nothing for it to do.

mikedickey: You can tell if moc'ing is necessary by looking for a `Q_OBJECT` macro inside the source; this is what the tool looks for. i.e. for `src/dblsqd/feed.h`

```
class Feed : public QObject
{
    Q_OBJECT
```

mikedickey:
> I hit a crash when attempting to start + join an existing studio within the desktop app:

I'm pretty certain this is unrelated, and haven't been able to reproduce it. However, I added a bug in our tracking list so that we don't forget about it.

nwang92 (approved):

ntonnaett:
From my short testing with nogui and novs, this runs fine. Code looks fine too. Will we have two separate GUI applications now or in the near future?

psiborg112:
Sorry, super late to this. (It's been a crazy few months.) Doesn't look like anything should have broken (just spent some time looking through all the diffs), other than a few minor cmake/me related config things which I've just made a pull request for.

PR #1321, opened 2024-08-28
Adding a confirmation dialog when connecting with audio warnings

mikedickey:
When connecting to a studio session via the device setup page, this checks to see if you have active audio warnings. If you have no warnings, it continues to connect as before. If you have audio warnings, it will open a dialog asking you to confirm.

Basically, we found that a lot of people were missing the warning icon and not realizing that their audio config had problems. This makes it clearer that something is wrong, helping educate people about requirements while still allowing them to proceed if they wish.

This also changes the red text buttons to black text to be more consistent with our newer UI designs.


mikedickey:
Light mode

Screenshot 2024-08-28 at 2 37 15 PM

Dark mode

Screenshot 2024-08-28 at 2 36 31 PM

Max UI scale

Screenshot 2024-08-28 at 2 36 31 PM

nwang92 (approved):

PR #1324, opened 2024-08-30
Getting rid of the 5-second studio refresh timer

mikedickey:
It makes for a poor user experience when the list keeps jumping around, and we already have a button to manually perform a refresh (in which case it isn't so annoying).

Before it not only would refresh when not desired, it failed to refresh when you wanted it to (or you had to wait for it, which made things confusing with status changes). Now it should ensure a refresh happens immediately, whenever you navigate to the "browse" window.

### DIFF
--- a/src/vs/Browse.qml
+++ b/src/vs/Browse.qml
@@ -10,8 +10,6 @@ Item {
         color: backgroundColour
     }
 
-    property bool refreshing: false
-
     property int buttonHeight: 25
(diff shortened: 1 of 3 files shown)


nwang92 (approved):

mikedickey:
Just a comment that I already found and fixed some other bugs that the auto-refresh timer was peanut-buttering over. There may be more lurking, so we should keep a look out.

PR #1327, opened 2024-09-05
Use new create studio flow

dhing1024:
Uses new create studio interface


nwang92 (approved):

mikedickey (approved):

mikedickey:
Just a note that this currently only works when in "Test Mode" because it requires jacktrip.com changes that haven't rolled out quite yet. We'll need to re-test after those changes are published to prod.

PR #1328, opened 2024-09-05
VS Mode improved error handling for login screen

mikedickey:
Assuming that DNS errors are caused by Internet connection offline and added special handling for that case


mikedickey:
Screenshot 2024-09-05 at 11 53 45 AM

nwang92 (approved):

PR #1335, opened 2024-09-13
Slightly different take on advanced settings for latency

mikedickey:
I'm not all that happy with the 3-way "Queue Type" combo + slider, and wondering if it may be better to simplify it to a checkbox + slider. Please give this a try and compare to the latest in dev, and let me know which one you prefer.

Screenshot 2024-09-12 at 10 07 12 PM

### DIFF
--- a/src/vs/Settings.qml
+++ b/src/vs/Settings.qml
@@ -60,16 +60,10 @@ Item {
     }
 
     function getQueueBufferString () {
-        if (audio.queueBuffer == -500) {
+        if (audio.queueBuffer == 0) {
             return "auto";
         }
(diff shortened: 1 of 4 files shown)


nwang92:
I think I like this one better. I'd probably flip the "Fixed Latency" and "Network Latency" lines though because the checkbox feels more like the "coarse" adjustment and the slider feels more like fine tuning the selection.

mikedickey:
I decided in the end to just drop the checkbox along with support for fixed latency. I think this is really just an experimental setting. Auto is always going to be a better choice for real world usage, so it's better to just keep things simpler and not have to explain it.

Screenshot 2024-09-13 at 12 38 11 PM

Screenshot 2024-09-13 at 12 37 59 PM

nwang92 (approved):

PR #1341, opened 2024-10-08
Feature/specify tcp server bind address

nwang92:


nwang92:
> Does this change support splitting a hub listener into two hosts? TCP host for managing ephemeral UDP ports and then the usual UDP host. Or is it for something different?

I'm attempting to support running the jacktrip server on a specific network interface so that on systems with multiple network interfaces we could potentially run multiple jacktrip servers. For example in the system below, I could theoretically run two different jacktrip servers on `10.2.17.225:4464` and `10.2.30.177:4464`:
Screenshot 2024-10-08 at 10 20 25 AM

I'm not sure how feasible that is yet but a prerequisite would be some option to only bind to a specific address which seems to be missing at the moment. I think this new bindaddress option also needs to be used when binding UDP ports too...

cchafe:
Very nice!

nwang92:
Closing because this isn't needed anymore

PR #1342, opened 2024-10-15
Fixes for cmake builds

psiborg112:
Fixes to cmake builds after recent changes. Also includes a fix to allow the classic interface to build with ancient versions of Qt.


psiborg112:
On a related note about building with ancient libraries, I notice that there seem to be code pathways for versions of RtAudio prior to 6, but the code failed to build on a machine with RtAudio 5. The problem was with RtAudio::getApiDisplayName in RtAudioDevice::print(). (Since I'm using JACK anyway, I just built it without RtAudio support, but there might be some people who still want this?)

mikedickey (approved):

PR #1345, opened 2024-10-18
Work-around for broken base64 on macos 13

mikedickey:
Also add sha tag to container for develop branch builds

base64: invalid argument --output
Usage:	base64 [-hDd] [-b num] [-i in_file] [-o out_file]
  -h, --help     display this message
  -Dd, --decode   decodes input
  -b, --break    break encoded string into num character lines
  -i, --input    input file (default: "-" for stdin)
  -o, --output   output file (default: "-" for stdout)

### DIFF
--- a/.github/workflows/jacktrip.yml
+++ b/.github/workflows/jacktrip.yml
@@ -494,8 +494,8 @@ jobs:
           KEYCHAIN_PATH=$RUNNER_TEMP/app-signing.keychain-db
 
           # import certificate and provisioning profile from secrets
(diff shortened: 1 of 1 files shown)


mikedickey:
The failing task only run in the JackTrip Labs account (when credentials are present). 

[Failure run here](https://github.com/jacktriplabs/jacktrip/actions/runs/11407646046/job/31744123133)

[Fix working here](https://github.com/jacktriplabs/jacktrip/actions/runs/11408160024)

nwang92 (approved):

PR #1348, opened 2024-11-15
Use Cookies For Authorization

dhing1024:
Passes in the access cookie into the `auth_code` cookie rather than as a bearer token authorization header. Applies to requests made from the C++ code and from webengine. TODOs:

- [x] Update places where the access token is passed in as an Authorization header to instead pass it in as a cookie
- [x] Would be nice to ensure that the changes in https://github.com/jacktrip/jacktrip/pull/1354 also apply to the web engine cookies.


dhing1024:
I think this is ready for a merge. I updated the PR to also include the QtWebEngine cases as a follow up to https://github.com/jacktrip/jacktrip/pull/1354

mikedickey (approved):

PR #1350, opened 2024-11-21
Adding OSC config server

nwang92:
This PR embeds https://github.com/kaoskorobase/oscpp to establish the ability to send OSC packets to the jacktrip server process


mikedickey:
I think we should move `oscpp` into the `externals` subdirectory, and have a separate #define for that (independent of virtual studio), as well as a separate meson build time option to enable it, similar to how we use the Simple-FFT project. We usually build jacktrip server with `nogui` (which also disables all the virtual studio mode stuff).

mikedickey (changes requested):

nwang92:
> I think we should move `oscpp` into the `externals` subdirectory, and have a separate #define for that (independent of virtual studio), as well as a separate meson build time option to enable it, similar to how we use the Simple-FFT project. We usually build jacktrip server with `nogui` (which also disables all the virtual studio mode stuff).

Can do

.gitmodules
```diff
@@ -7,3 +7,6 @@
 [submodule "externals/rtaudio"]
 	path = externals/rtaudio
 	url = https://github.com/thestk/rtaudio.git
+[submodule "externals/oscpp"]
```
mikedickey, line 10: I consider the use of git submodules to be deprecated, in favor of using meson's subprojects or (in the case of Simple-FFT which is a also header only) just checking things directly into the repo's "externals" directory. People working down-stream complained when we first added Simple-FFT as a git submodule. So I think this should also be added not as a subproject but just by checking the header files in directly.

mikedickey: I think it would also be ok if this was wrapped as a meson subproject, but then it should probably either reference the source repo or be a fork in the jacktrip project, if necessary.

nwang92: I added the source code of oscpp into `externals/oscpp` directly + removed the submodule declaration

mikedickey (approved):

PR #1351, opened 2024-11-27
A few updates for osc server support

mikedickey:
Updated logic for Regulator::setQueueBufferLength to handle both fixed and auto headroom.

Moved queue buffer update logging up to once per change versus once per client per change.

Added mAutoHeadroomStartTime to handle headroom calculation being started after initialization.

No need to moc oscpp headers. It just gives warnings since there is nothing to moc.


nwang92 (approved):
LGTM, just noting that we're currently sending values of [0, 128] to this endpoint, so we'll probably want to make some changes to use [-128, 0] when using buffer strategy 3/4

PR #1355, opened 2024-12-09
Prefer glitch count to skipped for increasing headroom

mikedickey:
This seems to be a better indicator which allows it grow and adjust faster for bad connections, without having an adverse impact on good connections.

Adding latency stat for Regulator (use "-I 1" to see it). This is helpful because tolerance is a maximum, and not necessarily indicative of true latency.

Adding receive latency to vs device heartbeat


cchafe:
took it for a spin, seems well-behaved -- on this network it's similar to the current 'main' branch

nwang92 (approved):

PR #1359, opened 2024-12-18
Adding support to build for Linux using a Docker container

mikedickey:
Using docker container to create all Linux release builds. Added new Linux arm64 static and dynamic builds, and a Linux arm32 static build for old Raspberry Pi bridge devices.

Switching qAsConst instead of std::as_const since older compilers that we are still using do not support it

Disabling PulseAudio in Linux static builds since it brings in a bunch of extra dependencies and is very SLOW

### DIFF
--- a/.github/workflows/jacktrip.yml
+++ b/.github/workflows/jacktrip.yml
@@ -86,25 +86,10 @@ jobs:
             # binary-path: binary
             build-system: meson
             meson-library-type: shared
+            static-analysis: true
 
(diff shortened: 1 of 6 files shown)


mikedickey:
[New builds here](https://github.com/jacktrip/jacktrip/actions/runs/12402698424). I tested them out on various devices and all seem to work great. arm32 should work on our old bridge devices.

nwang92 (approved):

PR #1368, opened 2025-01-05
JackTrip Audio Bridge VST3 Plugin

mikedickey:
This PR adds support for the "JackTrip Audio Bridge", a new VST3 compatible plugin that makes it much easier to integrate audio from other applications (like digital audio workstations) with JackTrip. A primary design goal has been to keep it simple: all audio sent to the plugin is mixed into audio from your "input device", and all audio coming back from a remote peer or JackTrip server is routed to the plugin's output.

The plugin is designed to be standalone from the JackTrip desktop application, as opposed to being a wrapper around it. This allows us to preserve the focus of the desktop app on providing the best UX for online collaboration, without trying to jam it into an architecture was never meant for that. It also means that you can have any number of plugin instances, running across any number of digital audio applications, with JackTrip being the glue that brings it all online.

The exchange of audio between the plugin and JackTrip's desktop app is accomplished using inter-process communication (IPC), a technique well tested and proven by the namesake [JACK Audio Connection Kit](https://jackaudio.org/). More specifically, we use local sockets or named pipes provided by [QLocalSocket](https://doc.qt.io/qt-6/qlocalsocket.html). This is built as extension to the protocol initially added to support deep links, and through this is extended to include support for live audio pipes. All the real work is encapsulated into a new `AudioSocket` class and I would envision this being reusable for other types of plugins, including things like LV2 and LADSPA as well as specialized commercial ones like those in Reason or Universal Audio. It's all open source, and ideally every audio application would just bake in support for JackTrip as a native feature.

The plugin offers a basic GUI with three volume level controls:
* Send: adjusts level for audio being sent from plugin to JackTrip's input
* Receive: adjusts level for audio being received by plugin from JackTrip's output
* Passthrough: adjusts level for audio passing from plugin's input to plugin's output

In addition, there is an LED in the upper-right corner that indicates whether or not it's connected to the JackTrip desktop app. It automatically handles (re)connections whenever the app is available.

jacktrip_audio_bridge

Building the plugin requires [Steinberg's VST3 SDK](https://github.com/steinbergmedia/vst3sdk), which is available under a dual-license (GPL3 and proprietary). JackTrip Labs has a proprietary license agreement with Steinberg that allows for binary redistribution without copyleft requirements. It will only be built if you add the parameters `vst-sdkdir` and `vst-libdir` to `meson setup` command, i.e. `-Dvst-sdkdir=/opt/vst3sdk -Dvst-libdir=/opt/vst3sdk/lib`

TODOs
- [x] Update CICD to build and package plugin
- [x] Support peer with different buffer size
- [x] ~Support peer with different sample rate?~ (punting this to a different PR)


src/AudioInterface.cpp
```diff
@@ -715,34 +718,51 @@ void AudioInterface::appendProcessPluginToMonitor(ProcessPlugin* plugin)
     mProcessPluginsToMonitor.append(plugin);
 }
 
+void AudioInterface::appendAudioSocket(QSharedPointer<AudioSocket>& s)
```
nwang92, line 721: NIT but should there be a `isNull()` check for the argument `s` here?

src/AudioSocket.cpp
```diff
@@ -0,0 +1,612 @@
...
+    // TODO: handle sample rate conversions
+
+    // copy bytes from remote to outputs
+    char* nextPtr = mRecvBuffer.data();
+    for (int i = 0; i < AudioSocketNumChannels; i++) {
+        int outputChannel = i < getNumOutputs() ? i : 0;  // mix to mono
+        float* framePtr   = reinterpret_cast<float*>(nextPtr);
+        for (int j = 0; j < nframes; j++) {
+            // TODO: is addition sufficient for mixing audio?
+            outputs[outputChannel][j] += *(framePtr++);
```
nwang92, line 236: I don't know what the correct answer is but I assume just pure addition would be susceptible to clipping?

mikedickey: I believe [this commit](https://github.com/jacktrip/jacktrip/pull/1368/commits/42589ac6a7b680daa75c59c8eb9457072f08efb8) addresses it well enough for now..

cchafe:
Can you point me to the build assets? If there's a linux VST, I'll take it for a spin.

nwang92:
> Can you point me to the build assets? If there's a linux VST, I'll take it for a spin.

Seems like one of the artifacts `JackTrip-eaaec2a-Linux-*` produced in https://github.com/jacktrip/jacktrip/actions/runs/12751489391?pr=1368 should have it

cchafe:
Works great with Ardour under linux!

src/AudioSocket.cpp
```diff
@@ -0,0 +1,665 @@
...
+    }
+
+    cout << "Established audio socket connection" << endl;
+    emit signalConnectionEstablished();
+}
+
+//*******************************************************************************
+void AudioSocketWorker::close()
+{
+    if (!isConnected()) {
```
nwang92, line 392: This seems like a very rare case but `mSocketPtr->state()` could be `QLocalSocket::ConnectingState` as it's initiating the connection. If `AudioSocketWorker::close()` is called during this transitory phase, the socket wouldn't get closed and might leave a dangling connection behind?

mikedickey: I don't think it's possible because we block on connecting and only the worker thread can perform the socket operations, but I added an extra check to be safe here https://github.com/jacktrip/jacktrip/pull/1368/commits/091c5ff02b14f630fc2c888c99736547af7fe228

PR #1370, opened 2025-01-15
Sample rate conversion for audio sockets

mikedickey:
Audio socket support for sample rate conversion

Interleave samples for audio sockets. This makes the code a lot simpler and eliminate the need to even think about things like local and remote buffer sizes. Not having to wait for buffer size boundaries can also help minimize any latency introduced.

Improvements to vst logging so that it includes cout and cerr


mikedickey:
@ntonnaett is there some magic required to allow flatpak to download files over the Internet? It seems to be failing [here](https://github.com/jacktrip/jacktrip/actions/runs/12795879564/job/35674236195#step:5:103). Or perhaps there is a way to use a local package of libsamplerate? I disabled sample rate conversion in Flatpak for now.

nwang92 (approved):

meson.build
```diff
@@ -343,6 +343,28 @@ if rtaudio_dep.found() == false and jack_dep.found() == false
 	configure.''')
 endif
 
+libsamplerate_dep = []
+found_libsamplerate = false
+if get_option('libsamplerate').allowed()
+	opt_var = cmake.subproject_options()
+	if get_option('buildtype') == 'release'
```
ntonnaett, line 350: This is not a good choice. Distributions use debugoptimized for releases because they generate packages for debug symbols. The wording is just stupid because they copied the naming from cmake.

ntonnaett:
> @ntonnaett is there some magic required to allow flatpak to download files over the Internet? It seems to be failing [here](https://github.com/jacktrip/jacktrip/actions/runs/12795879564/job/35674236195#step:5:103). Or perhaps there is a way to use a local package of libsamplerate? I disabled sample rate conversion in Flatpak for now.

I think what you want is to install the dependency first and then check for the dependency. If the dependency is not available, do the logic that is already there.

ntonnaett:
It's not stable yet, but in the future the best way might be to use this for the wrap: https://mesonbuild.com/Wrap-dependency-system-manual.html#cmake-wraps

PR #1375, opened 2025-01-17
meson: support libsamplerate as system dependency

ntonnaett:
This should handle the build with libsamplerate as a system dependency.

WIP: There's still a logic error, because I needed to set required: false.


mikedickey:
@ntonnaett works ok for me after [applying this patch](https://github.com/jacktrip/jacktrip/pull/1389)

PR #1376, opened 2025-01-17
Removing the extra Gain dial from audio bridge interface

mikedickey:
Jan and I went through a bunch of iterations on this, and came to the realization that there is really no need to applying extra gain to the plugin's output audio, and the extra UX complexity this third dial adds wasn't worth it.

This updated design also adds some extra labels to help make it clearer where the "Send" is going, that the "Mix" applies to audio output from the plugin, and what you are mixing between.

Screenshot 2025-01-17 at 1 59 31 PM


mikedickey:
Note also that the new approach and default settings should be non-destructive, in that the signal going out of the plugin will not be altered from the signal going into it. I think this is especially important given that people will usually be using this as a plugin that they drop into their insert chain, which likely has other effects. Most use cases are for purely sending audio, and I don't think we want to alter the track's signal unless they explicitly change the output mix to do so. Conversely, this allows someone to record using an unaltered signal from JackTrip by just turning the mix dial all the way to the right.

nwang92 (approved):

PR #1384, opened 2025-01-30
Fix regression with reconnecting JackTrip audio during live session

mikedickey:
this reverts the logic in VirtualStudio::connectToStudio except that it adds an extra check for the server host being empty. The first time you connect, this should always be empty, forcing it to wait for the websocket response. But it will not be empty for subsequent reconnects to the same studio, enabling it to not have to wait for a new websocket message.


nwang92 (approved):

mikedickey:
flatpak job seems flaky.. like it doesn't like the git hash or something?

PR #1385, opened 2025-01-30
Fix for display of failed message screen during startup on Linux

mikedickey:
Fix studio padding (or lack thereof) when browsing on Linux

I think these are more related to the use of newer Qt releases versus Linux, but since we pin to older releases on other platforms I'm only mentioning that in the changelog

I honestly don't understand why rearranging things in vs.qml fixes the display of the failed screen, but it seems to work..


nwang92 (approved):

mikedickey:
It turns out that the flatpak-builder github action we're using is using an older API (v3) to upload artifacts to github, and github stopped supporting it today.

PR #1389, opened 2025-02-06
Fix for -Dlibsamplerate=disabled

mikedickey:
I think it works ok with system and subproject after this change

### DIFF
--- a/meson.build
+++ b/meson.build
@@ -343,25 +343,27 @@ if rtaudio_dep.found() == false and jack_dep.found() == false
 	configure.''')
 endif
 
-libsamplerate_dep = dependency('samplerate', required: false)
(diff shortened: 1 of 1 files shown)


ntonnaett:
This still uses the wrap when setting wrap_mode to nofallback.

ntonnaett:
If the libsamplerate is found on the system, it uses that one. So I think that's an improvement.

ntonnaett:
The Flatpak CI built fine but the upload didn't work. Don't know what's happening there.

ntonnaett (approved):

mikedickey:
I think I fixed the flatpak issue here https://github.com/jacktrip/jacktrip/pull/1385/commits/150ca355969c335653a7c84416f32227ab68fa5e

PR #1394, opened 2025-02-13
Ignore timestamps recording in gzip metadata

Antiz96:
Use the `-n / --noname` option to ignore non-deterministric information, such as timestamps, in gzip metadata.

This is required for [reproducible builds](https://reproducible-builds.org/).


mikedickey (approved):

mikedickey:
Ubuntu seems to be having trouble with their libc packages on arm64. Not going to hold this up for it; hopefully the fix it soon.

Antiz96:
Alright, thanks! :pray: :tada:

PR #1403, opened 2025-04-12
Embed jacktrip.com website into desktop app

mikedickey:
This updates virtual studio mode to replace the studios list with with a logged-in view of the home page at jacktrip.com.

### DIFF
--- a/src/vs/Browse.qml
+++ b/src/vs/Browse.qml
@@ -1,5 +1,6 @@
 import QtQuick
 import QtQuick.Controls
+import QtWebEngine
 
 Item {
     width: parent.width; height: parent.height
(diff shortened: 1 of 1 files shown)


mikedickey:
This is mainly just a POC at this point. Some things I've noticed so far:
* auth seems to work ok (except "Test Mode" only works for 1 session and requires log-out/in again after restarting app)
* playback of recordings seems to work ok

We'd need to add conditionals in a number of places:
* "Download App" obviously doesn't make sense
* Join buttons should skip past popup and jump right in
* Logout from profile menu should log out app
* Re-evaluate buttons at bottom, like "Refresh List" should probably be a browser reload button, add a back button, home button?
* ...

Another thing to consider is links from vs-web: ideally these don't leave the app but switch to browser view. Perhaps there needs to be some sort of tabs where a session can be a tab? Or maybe a "Return to Session" button in footer?

Lots of little things but seems to work fairly well for the most part..

Issue #1402, opened 2025-03-31
error: invalid use of incomplete type ‘class QDebug’


Enchufa2:
Since Qt 6.9.0 RC1 landed on Fedora rawhide last week, builds fail with:

```
../src/AudioSocket.cpp: In member function ‘void AudioSocketWorker::scheduleReconnect()’:
../src/AudioSocket.cpp:588:15: error: invalid use of incomplete type ‘class QDebug’
  588 |         qDebug() << "Attempting to reconnect audio socket";
      |               ^
In file included from /usr/include/qt6/QtCore/qglobal.h:45,
                 from /usr/include/qt6/QtCore/qiodevice.h:7,
                 from /usr/include/qt6/QtNetwork/qlocalsocket.h:8,
                 from /usr/include/qt6/QtNetwork/QLocalSocket:1,
                 from ../src/AudioSocket.h:41,
                 from ../src/AudioSocket.cpp:38:
/usr/include/qt6/QtCore/qtypeinfo.h:15:7: note: forward declaration of ‘class QDebug’
   15 | class QDebug;
      |       ^~~~~~
```

see [the complete log](https://kojipkgs.fedoraproject.org/work/tasks/5132/130845132/build.log).

mikedickey:
@Enchufa2 I don't have 6.9 readily available, but it seems straight-forward enough.. would you be able to verify if [this PR](https://github.com/jacktrip/jacktrip/pull/1406) fixes it?

Enchufa2:
It does.

PR #1406, opened 2025-04-17
Replacing qDebug() with cout for Audio Socket reconnect

mikedickey:
Bumping ubuntu-20.04 to ubuntu-22.04 for qmake static build, since 20.04 is no longer supported by GitHub

Disabling vs mode for Linux-x64-meson-gcc-shared-nortaudio build since ubuntu-22.04 has qt 5.13 (not sure how this worked before?)

### DIFF
--- a/.github/workflows/jacktrip.yml
+++ b/.github/workflows/jacktrip.yml
@@ -59,7 +59,7 @@ jobs:
       matrix:
         include:
           - name: Linux-x64-qmake-gcc-static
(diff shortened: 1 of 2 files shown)


mikedickey:
I think this fixes https://github.com/jacktrip/jacktrip/issues/1402

BSDKaffee:
FWIW, simply adding `#include <QDebug>` to src/AudioSocket.cpp fixes the issue for me.

PR #1419, opened 2025-04-23
Fix build with Qt 6.9

BSDKaffee:
Partially revert dd2d908. Add missing QDebug include to fix forward declaration issue with Qt 6.9 mentioned in #1402.

### DIFF
--- a/src/AudioSocket.cpp
+++ b/src/AudioSocket.cpp
@@ -37,6 +37,7 @@
 
 #include "AudioSocket.h"
 
+#include <QDebug>
 #include <QEventLoop>
 #include <iostream>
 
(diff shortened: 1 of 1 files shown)


mikedickey:
What is your concern with using `cout` versus `qDebug`? Does this not work with qt 6.9 as is?

Granted, I think we have a bigger issue with logging in general: when to use stdout versus stderr, when something is "debug" versus should always be logged, etc. Developing and documenting some standards on all this, and going through all the code to confirm, would be a worthy exercise for someone to tackle.

But for this particular line, I think an audio socket reconnect is notable and possibly indicative of problems, worthy of more than just "debug."

BSDKaffee:
> What is your concern with using `cout` versus `qDebug`? Does this not work with qt 6.9 as is?
> 
> Granted, I think we have a bigger issue with logging in general: when to use stdout versus stderr, when something is "debug" versus should always be logged, etc. Developing and documenting some standards on all this, and going through all the code to confirm, would be a worthy exercise for someone to tackle.
> 
> But for this particular line, I think an audio socket reconnect is notable and possibly indicative of problems, worthy of more than just "debug."

It's really of no concern to me. I was just trying to get the original code to build with Qt 6.9 with minimal intrusion. Feel free to close this if using qDebug is overall problematic.

mikedickey:
Got it. I think using cout is fine, and likely preferred. That is why I opted to do that in https://github.com/jacktrip/jacktrip/pull/1406.

PR #1426, opened 2025-06-10
Use atomic<int> for glitch counter shared across threads

mikedickey:
This only impacts stats when using -I

I've seen evidence of the reported glitches not making sense because I think the stats thread is using a cached value on arm64


mikedickey:
I think there is a different problem at play here...

PR #1430, opened 2025-06-21
When skipping packets, advance last and use next for training

mikedickey:
Previously, we would use the next good packet for training, but never use it as "real" packet. Now if at least one packet is being skipped, we will use the next good packet for training but only advance the last pointer to the slot before it. This helps to ensure that the next pull will return the next good packet.

My testing so far indicates that this significantly improves clarity when there is packet loss or delay spikes.

### DIFF
--- a/src/Regulator.cpp
+++ b/src/Regulator.cpp
@@ -612,7 +612,6 @@ bool Regulator::pullPacket()
 {
     const double now       = (double)mIncomingTimer.nsecsElapsed() / 1000000.0;
     const int lastSeqNumIn = mLastSeqNumIn.load(std::memory_order_acquire);
-    int skipped            = 0;
 
     if ((lastSeqNumIn == -1) || (!mInitialized) || (now < mMsecTolerance)) {
         // return silence during startup:
@@ -641,21 +640,44 @@ bool Regulator::pullPacket()
                            > mMsecTolerance)
                     continue;
                 updatePushStats(next);
-                // count how many we have skipped
-                skipped = next - mLastSeqNumOut - 1;
-                if (skipped < 0)
-                    skipped += NumSlots;
             }
             // check if packet's age matches tolerance, or is the best candidate we have
-            if (mIncomingTiming[next] + mMsecTolerance >= now || i == 0) {
+            const bool meetsTolerance = mIncomingTiming[next] + mMsecTolerance >= now;
+            if (meetsTolerance || i == 0) {
                 // next is the best candidate
+                // we will use it for output, or training if skipped
                 memcpy(mXfrBuffer, mSlots[next], mPeerBytes);
-                mLastSeqNumOut = next;
-                double latency = (now - mIncomingTiming[mLastSeqNumOut]);
-                if (latency > mStatsMaxLatency) {
-                    mStatsMaxLatency = latency;
+                // count how many we have skipped
+                int skipped = next - mLastSeqNumOut - 1;
+                if (skipped < 0)
+                    skipped += NumSlots;
+                if (skipped) {
+                    // if we skipped any packets, process it as a glitch
+                    if (meetsTolerance) {
+                        // increment last out to the previous skipped packet
+                        mLastSeqNumOut = next - 1;
+                        if (mLastSeqNumOut < 0)
+                            mLastSeqNumOut = NumSlots;
+                    } else {
+                        // increment last out to the next valid one
+                        // so that it only gets used for training
+                        mLastSeqNumOut = next;
+                    }
+                    pullStat->plcOverruns += skipped;
+                    processPacket(true);
+                    return true;
+                } else {
+                    // none skipped: next is a "real" packet (not a glitch)
+                    // update max latency
+                    double latency = (now - mIncomingTiming[next]);
+                    if (latency > mStatsMaxLatency) {
+                        mStatsMaxLatency = latency;
+                    }
+                    // advance last out and process OK packet
+                    mLastSeqNumOut = next;
+                    processPacket(false);
+                    return false;
                 }
-                goto PACKETOK;
             }
             // track how many good packets we skipped due to tolerance < 1ms
             if (mIncomingTiming[next] + mMsecTolerance + 1 >= now) {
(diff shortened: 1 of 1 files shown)


mikedickey:
@cchafe i would love to get your eyes and ears on this one

cchafe (commented):
I'm getting something that's got a new sound  --
here's main 
image
here's the new version -- seems to set up a sequence 
I need a better way to compare but they sound different 
image

cchafe:
I'll bring back the internal test tone
sineToXfrBuffer();
to create regular packet loss.
For example,

    if (mLastSeqNumOut%15 < 10) glitch = true;

    double tmp = 0.0;
    if (glitch)
        tmp = (double)mIncomingTimer.nsecsElapsed();
    zeroTmpFloatBuf();  // ahead of either call to burg

    sineToXfrBuffer();

    xfrBufferToFloatBuf();
    burg(glitch);
    floatBufToXfrBuffer();

What's the most convenient way to add that and in the same branch turn on / off the new skipping feature ?

mikedickey:
@cchafe is the "new sound" a better sound? The audio should be identical if there are no packets being skipped. Unless I'm missing something there, the code path is exactly the same. When there is loss (skips), it should sound different (much better / closer to the original sound). There is no flag to turn this on or off; you'd have to use different branches for the tests.

Note that dev branch also includes [this bug fix ](https://github.com/jacktrip/jacktrip/pull/1431/files) which is not reflected in this PR. That will also change (it should improve) the sound when we receive out-of-order packets, and I think fixes a really ugly bug that causes it to get stuck looping over the entire set of NumSlots.

It may be better to compare main vs dev vs this branch?

src/Regulator.cpp
```diff
@@ -648,41 +649,71 @@ bool Regulator::pullPacket()
...
+                        // next packet doesn't meet tolerance
+                        mLastSeqNumOut = next;
+                        memcpy(mXfrBuffer, mSlots[mLastSeqNumOut], mPeerBytes);
+                    }
+                    pullStat->plcOverruns += numSkipped;
+                    pullStat->plcUnderruns++;
+                    processPacket(true);
+                    return true;
+                } else {
+                    // none skipped: next is a "real" packet (not a glitch)
```
mikedickey, line 692: Note that if no slots are being skipped, this block is called. It's the same logic we currently have with the "PACKETOK" and "OUTPUT" gotos. I'm just eliminating those because they are not used elsewhere, and I think gotos in general are hard to read and should generally be avoided.

mikedickey: note that it's best to use a fixed queue size when testing (i.e. `-q5`) that includes enough headroom to get "extra fresh" packets into the mix.

cchafe:
Works well.
I imposed 50% underrun artificially and directly recorded the effect on internal sine tones, tapping the signal after Burg.
Same result as with dev and main. I have a patch file for setting up the test.

mikedickey:
> Skipped = overrun, sorry about that. Now I impose overruns and see a different behavior between this branch and dev where dev is better behaved. I'll have to dig further... some reality checking.

Which `-q` setting are  you using? If tolerance is too low (less than say 2 packet intervals), it could be because you don't have any "extra fresh" packets. There is a very ugly bug in main/dev that causes it glitch out continuously, sometimes for several seconds, due to it picking a packet that just barely meets tolerance. It could have several good packets available, but because we always take the oldest, the next one times out before the next callback. The latest commits in this branch fix it by skipping ahead -- when there already is a skip -- far enough so that the next packet isn't going to time out before the next callback. It could appear to cause a degradation when tolerance is artificially low. Using something like `-q10` should avoid it.

mikedickey:
> Earlier: I imposed 50% underrun artificially and directly recorded the effect on internal sine tones, tapping the signal after Burg. Same result as with dev and main. I have a patch file for setting up the test.

I've been using "Bridge Over Troubled Waters" as my test track. Playing it via VST audio bridge in Audacity to a loopback server on the same network, and setting x% loss on that server's interface for outbound traffic.

Testing 50% underruns I see something like this when using "-V" with dev branch:

```
PLC avg 0.172144 glitches 215 skipped 0 tolerance 5.04369 +4
PLC avg 0.172065 glitches 210 skipped 0 tolerance 5.04369 +4
PLC avg 0.172377 glitches 225 skipped 0 tolerance 5.04369 +4
PLC avg 0.171037 glitches 223 skipped 0 tolerance 5.02827 +4
PLC avg 0.165926 glitches 205 skipped 0 tolerance 5.02827 +4
PLC avg 0.171266 glitches 210 skipped 0 tolerance 4.96372 +4
PLC avg 0.172447 glitches 231 skipped 0 tolerance 4.96372 +4
PLC avg 0.169424 glitches 224 skipped 0 tolerance 5.11022 +4
PLC avg 0.166438 glitches 232 skipped 0 tolerance 5.11022 +4
PLC avg 0.171829 glitches 225 skipped 0 tolerance 5.11022 +4
```

Versus this when using either of my PR branches:

```
PLC avg 0.177678 glitches 147 skipped 0 tolerance 5.01986 +4
PLC avg 0.178154 glitches 152 skipped 0 tolerance 5.00591 +4
PLC avg 0.178011 glitches 149 skipped 0 tolerance 5.00591 +4
PLC avg 0.177995 glitches 138 skipped 0 tolerance 4.99616 +4
PLC avg 0.178786 glitches 149 skipped 0 tolerance 4.99616 +4
PLC avg 0.177581 glitches 147 skipped 0 tolerance 4.99616 +4
PLC avg 0.176796 glitches 154 skipped 0 tolerance 4.98891 +4
PLC avg 0.176948 glitches 147 skipped 0 tolerance 4.98891 +4
PLC avg 0.177359 glitches 142 skipped 0 tolerance 4.99129 +4
PLC avg 0.172459 glitches 143 skipped 0 tolerance 4.99129 +4
```

Although there is a lot of garble masking it, I can make out the song pretty well with the newer branches. Using dev it's completely indiscernible. 

Note that 48000 with 128 buffer size = 375 packets per second. The big change in this branch has to do with what we send to the burg algorithm when we have a choice of packets available. It causes far fewer glitches by using more valid packets for training.

PR #1433, opened 2025-06-22
Fixing lockups with some VST hosts on Windows

mikedickey:


mikedickey:
Not sure "fixing" is the right word, but this at least seems to work around the most egregious issues

PR #1437, opened 2025-06-26
Don't glitch on the first good packet if previous was a glitch

mikedickey:
If the previous call returned a predicted result (glitch), and we have a good packet available, don't process it as a glitch.

Updated changelog for 2.7.0-beta2 release


cchafe (approved):
huge win!
tested vs. current dev branch and in the overrun case it outperforms amazingly

PR #1462, opened 2025-12-30
Asked GPT-5 to write docs on the network protocol

mikedickey:


cchafe:
great to have this doc!
_Where it says:_
a short-lived TCP connection is used to exchange UDP port information (and optionally do TLS + credentials).
let's add something about "**ephemeral port**"
_Where it says:_
When redundancy is enabled, multiple full packets are concatenated into a single UDP datagram (see “UDP redundancy”).
let's add something about "**forward error correction**"
_Where it says:_
(JamLink compatibility).
let's add "**an obsolete device**"
_Needs more specificity:_
is populated from local *output* channel count. **Which output, audio or network?**
_Needs correct .md math formatting:_
\[
\text{headerBytes} + (N \times C \times \text{bytesPerSample})
\]
_For More:_
add a reference to the following article which translates -V debug info pertinent to the above
AUTHOR=Chafe Chris 
TITLE=I am Streaming in a Room
JOURNAL=Frontiers in Digital Humanities
VOLUME=Volume 5 - 2018
YEAR=2018
URL=https://www.frontiersin.org/journals/digital-humanities/articles/10.3389/fdigh.2018.00027
DOI=10.3389/fdigh.2018.00027

PR #1484, opened 2026-04-05
Various recommended patches

mikedickey:


cchafe (approved):
compiles and runs -- fedora 43 pipewire

PR #1491, opened 2026-04-18
Adding some more constraints on the growth of auto headroom

mikedickey:
Give a little bit more adjustment time during startup. Especially for Windows (which has poor kernel scheduling) I have observed it can take a little longer for things to stabalize. Even on a very fast and powerful computer.

Half the adjustment when growing headroom by more than 1ms. This just adds a little more safety over it growing too fast.

Zero both mLastMaxLatency and mStatsMaxLatency during warmup period.

Fix log typo when setting realtime priority on windows.


cchafe (approved):

cchafe:
I notice there are failing checks, so I won't merge

